PR #421, opened 2015-11-28
Fix #418: implement paste-by-cell when pasting multiple effects

### LINKED ISSUE #418: Add ability to paste by cell
Xlights should support the ability to paste by cell rather than by time like Light-O-Rama does. When enabled, it causes pasted effects to end at the next timing gridline, rather than having the same duration as the copied effect.

Just to clarify: this issue only applies when copying multiple effects at once.

josephcsible:


josephcsible:
I now notice that this doesn't work right if there's gaps, or if multiple rows are copied and pasted at once. I'll add another commit to this to fix that.

josephcsible:
This works right with those now. I think it's good to merge.

AzGilrock:
I just happened to notice this.  Developers normally communicate with our team first before going off and doing work.  Copy/Paste involves a lot more things because the same logic is used to place keyboard shortcut commands and presets in addition to normal copy/paste operations.  I gotta troubleshoot all this stuff so I don't want anything merged in that I don't fully understand.  We've been possibly needing a change to the way our data is stored in the clipboard also.  Plus inspecting your changes it appears to assume one always wants to paste by cell if a timing track is active.  I don't want to turn off my timing track every time I want to do a paste.  This wasn't how I envisioned this feature being added and it's on my ToDo list.

PR #567, opened 2016-05-14
Allow dev to set number of of build threads

cp16net:
This will default to 4 threads as was set for the wxwidgets build but
also for the main xlights build as well. Allows you to override the 
number by setting the ENV "XLIGHTS_COMPILE_THREADS" if you
like.


cp16net:
@cjd this will make things build a little faster like wxWidgets.

cjd:
This can already happen if you pass the -j option to make (ie: 'make -j4')
It will automatically be passed to the rest of the compile (except for wxwidgets as that is forced to -j4)

cp16net:
Oh. i didnt realize that would work as well. good tip.

PR #572, opened 2016-05-15
default the linux desktop app link to use 1 thread

cp16net:


cjd:
Craig - is taskset still needed?
I know it was helpful a while ago - but I've seen no issues caused by threading since the move to wxwidgets 3.1

cp16net:
@cjd after i got all the dependencies and it built, i had some unexpected crashes until i ran with a single thread. *maybe* it was pre-3.1, not sure. I've been running with it for a few days testing my p10 panels. I'll try it out again without this and see what happens and reopen if its still an issue for stability.

PR #763, opened 2016-12-04
Added functionality to delete multiple layers

jrhubott:


keithsw1111:
I can understand why I would want to be able to delete several layers at once ... but deleting all layers for a model seems like a rare thing to want to do. Could we instead make it delete n bottom layers and prompt the user for n. Users could then move effects up to the top layers and then remove the bottom layers they no longer need. Maybe also check if any of the layers about to be deleted have any effects on them and warn them if they do and ask them to confirm deletion.

AzGilrock:
I would like it to be N number of layers from the layer you right-clicked.

jrhubott:
Updated the merge to include the requested functionality.  The delete now deletes multiple layers below the selected layer and asks for confirmation if any of the layers contain effect data.

keithsw1111:
I dont see why you would ever need to add back in a layer. It should be impossible to delete the top layer as by default this should only delete layers below the selected layer. It doesnt seem that you are validating the number of layers the user is asking to delete exist below the current layer.

ie if I have 10 layers and I am on layer 5 and I try to delete 7 layers it should only delete layers 6-10 ... not layers 4-10 which I believe this code will try to do.

AzGilrock:
No reason why you shouldn't be able to delete the top layer(s) as long as there is at least one layer left after your done.

AzGilrock:
This doesn't seem intuitive to me.  When I select a layer and say delete 3 layers I expect the layer I selected plus 2 below it to be deleted.

jrhubott:
That was how I had it working originally. 

Should I rename to Delete Layers and implement the change to behave as follows:

1. Delete layers down, including the selected layer. 
2. Ask for confirmation if layer contains effects.

AzGilrock:
I would say the name should be close to whatever the Insert menu item is.  I can't see it right now but if it's "Insert Multiple Layers" then this one should be "Delete Multiple Layers".  I'd expect it to delete the one I'm selecting plus N-1 below it and if that deletes all layers then add a new blank layer to the model afterwards.  That allows you to delete multiple layers that contains effects including the top layer and then you replace it with a blank layer.

AzGilrock:
I actually needed this 2 days ago.  I accidentally imported a SuperStar file onto the wrong model because I forgot to select my proxy model.  So I ended up with like 200 layers I needed to delete.  That means I wanted all layers including the top layer deleted and I wouldn't want to have to delete the effects manually first to be able to use this delete.  And what would be nice is if I don't know how many layers I have I could just type in a really big number like 500 and it would just delete whatever is there and then leave it with one blank layer.  Its pretty easy to add a layer to a model.  After you delete them all it's just a single call to add a new layer.  Sorry for having you continue to modifying this but I'm pretty good at knowing what users are going to want....and in this case I'm the user...lol.

jrhubott:
I also had the same problem and this is why I added this. I will submit an update later today.

PR #771, opened 2016-12-08
Add a button to reverse the numbering on custom models

jrhubott:
Reverses the numbering on custom models.  I used this to fix several models that were wired from the opposite end then the model I designed because I wired it backward.


keithsw1111:
I cant test but are we sure this generates the desired result if the user has left a gap in their node numbering. ie if they have nodes 1,2,3,5 then this seems to turn it into 5,4,3,1. I can see an argument that says this is likely what they would want but i am not sure.

jrhubott:
It will do exactly what you said.  I implemented this to solve a problem on several of my light models.  Just sharing this in case anyone else needs similar functionality.

PR #832, opened 2017-02-28
Fixed off by one and timing error in single strand chases

MrRocketman:
Chases were inconsistent in how they ended. Some would end with the last pixel still visible. Some would end with the last pixel not visible.
     
     There were also inconsistencies in the timing of frames. The first and last pixel locations (frames) were only half a normal frame length in duration.
     
     Most of these issue were resolved by using integer math. As a result of using integer math, the code was simplified.
     
     I also added in a commented out switch case for a true ‘dual bounce’ option. Since the current ‘dual bounce’ option is actually just a dual chase. It doesn’t bounce ha.
     
I also added some comments to make it easier to follow what’s happening.

### DIFF
--- a/xLights/effects/SingleStrandEffect.cpp
+++ b/xLights/effects/SingleStrandEffect.cpp
@@ -267,68 +267,82 @@ void SingleStrandEffect::RenderSingleStrandChase(RenderBuffer &buffer,
         case 3: // "Auto reverse r-l"
             AutoReverse=true;
             break;
-        case 4: // "Bounce"
+        case 4: // "Bounce" // Note this should actually be called "Dual Chase" here and in the UI, not dual bounce. Because this doesn't set [AutoReverse] to true, so it's only chasing
             Dual_Chases=true;
             break;
+            /*case 6: // "Actual Dual Bounce" // This is all you'd have to do to have a dual bounce
+             Dual_Chases=true;
+             AutoReverse=true;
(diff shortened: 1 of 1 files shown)


keithsw1111:
Did you test this on odd and even sized props which is likely where anomalies as a result of an off by 1 may appear?

MrRocketman:
Yeah, I tested it pretty thoroughly with odd and even sized matrices. Wouldn't hurt to double check though. Issues mostly showed up previously because the chase was calculated as a (double) 

Example:
10 wide matrix, 36 chase size setting.
It was doing the math off of 3.6 pixels wide (and rounding up to 4), but then would only draw it as 3 pixels wide. So that would produce different results than a 32 chase size setting (rounded down to 3) that was still only drawing 3 pixels wide. So overall it should be much more consistent now since the math is integer based.

And the timing was strange too. Because it wasn't evenly dividing the total time by the number of movements. So the first and last pixel locations would only be on screen for half the time as the rest of them. Now all the locations on are screen for the same time.

Definitely not saying it's bug free but it seems to be much more consistent.

dkulp:
Did some basic testing.   Looks ok to me.  Feel free to merge it to master.

MrRocketman:
Merged. https://github.com/smeighan/xLights/commit/822cd79cbf63b1eb59967a1f438568dcad1606c1

PR #912, opened 2017-09-14
Added nullptr check to prevent software from crashing if model parent object is null.

computergeek1507:
I cannot explain why the parent object is null but this will prevent the software from crashing.


AzGilrock:
Your touching temporary troubleshooting code.  We as a rule don't like to slow the program down with nullptr checks on every pointer in the program in places where it should be impossible to have a nullptr.  If it's null here I think we want it to crash instead of pushing the problem on down the line.

PR #914, opened 2017-09-15
Custom Model Renumber

computergeek1507:
I added a feature to increment or decrement the node numbers of a custom model. It basically just "shifts" the node numbers by the inputted value. I also added the Maximize Box to the windows.

### DIFF
--- a/xLights/CustomModelDialog.cpp
+++ b/xLights/CustomModelDialog.cpp
@@ -931,7 +931,6 @@ void CustomModelDialog::OnButton_RenumberClick(wxCommandEvent& event)
         }
     }
 
-    //max++;
     //Rewrite the grid values
     for(auto c = 0; c < GridCustom->GetNumberCols(); c++)
     {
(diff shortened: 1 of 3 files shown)


AzGilrock:
If you plan to start being a regular xLights contributor you should email Sean and he will introduce you to all of us through email and probably give you different privileges.  None of the rest of us work using pull requests and we have private ways of communicating between the developers of what we are working on.  Its best if we aren't surprised by some new code showing up because you don't want to do a bunch of work to find out there's a reason it's not a good idea.

Regarding this change just based on the description I have to question why you would need to offset all the numbers in a custom model?  It makes be think you might not know how they work because all custom models should start at 1.

computergeek1507:
It has a loop around element, if you have a model with 50 nodes, you can renumber by 10 it will shift 1 to 11 and 41 to 1. 

I very sorry for being so intrusive that was not my intention. This is my first year doing Pixel Christmas lights and I like the concepts behind how xLights works. As a relatively new user I am just trying to add simple code that may make it easier for other new users. I was listening to the Zoom meeting two weeks ago and someone had to renumber their whole model because they wired the element wrong. (I don't remember all the specifics) If this person could have been able to shift the starting location of the nodes it would of prevented this. I'm a Fedora user so I have to compile from source to use xLights, so I thought I would see if I could contribute a simple fix.

AzGilrock:
No problem.  I just saw 3 new pull requests in 2 days so if you were going to be doing regular updates we would bring you into the developer communications.  If you only expect to do small changes here or there then this way is fine.

PR #917, opened 2017-09-18
Custom Model Renumber

computergeek1507:
I added a feature to increment or decrement the node numbers of a custom model. It basically just "shifts" the node numbers by the inputted value. If you have a model with 50 nodes, you can renumber by 10 it will shift 1 to 11 and 41 to 1. I also added the Maximize Box to the windows.


computergeek1507:
Less commits to not fill up the main commit log. Took Kieth's idea and removed the spinbox

computergeek1507:
Fixed. Thanks for the idea.

PR #956, opened 2017-10-16
Added xLights/xSchedule Shortcuts to the Windows Start Menu

computergeek1507:


keithsw1111:
A couple of thoughts ... If i install both 32 & 64 bit will they clash?

You seem to refer to xLights4 ... shouldnt it just be xLights? the 4 we dropped back in 2016.

computergeek1507:
I actually changed the desktop shortcut from "xLights4" to "xLights" and added a start menu entry of just "xLights" for the 32bit version. For 64 bit, the desktop shortcut was already "xLights64" so I created a "xLights64" start menu entry. If you install both versions it will creates two start menu shortcuts ("xLights" and "xLights64") in the same xLights start menu folder that points to the two different install locations. For the "xSchedule" start menu entry it will only point to the last version you installed. I didn't want to change it to "xSchedule64" unless you think it is a good idea.

keithsw1111:
I just think you should be consistent. if you have xLights64 then xSchedule64 would make sense.

keithsw1111:
Ok. I also notice you dont set the working directory ... is there a reason for that? What working directory does it default to?

computergeek1507:
image

It defaults to the EXE install location.

Issue #973, opened 2017-11-07
Bars uses Irish math for color placement


stevegiron:
Here is the bars effect with 5 colors set to Palette Rep of 1 and cycles set to 0.

![image](https://user-images.githubusercontent.com/10568498/32519060-58858ce6-c3c0-11e7-8025-57081c8b908e.png)

When the color count is evenly divisible by the model height in pixels, the colors don't distribute evenly.
This would be understandable with non-evenly divisible height model. This is a 50px tall tree for example. It does this as well horizontally. 
![image](https://user-images.githubusercontent.com/10568498/32519227-e3ce919e-c3c0-11e7-90f3-a975e1d92335.png)

I'd like to be able to use bars to transition in and out of other effects but the bars distribution needs to be even.

stevegiron:
Any chance of merging this now that the season is over?

PR #974, opened 2017-11-08
Switched to ceiling function.

computergeek1507:
Fixes rounding errors on even divisible numbers. #973


keithsw1111:
While I am sure your fix is fine I think we should sit on this until after Christmas. Changing the way such a popular effect renders so close to Christmas seems inadvisable.

stevegiron:
In general I agree with Keith's cautionary approach. This is not a corporate endeavour where we're fiddling with the financials before a government audit. This is just blinking some lights.  I HIGHLY doubt anyone but me would even detect this change as most folks use some movement with bars and once you get them moving you can't detect this anomaly unless you use the alternate movements in which this oddity can be observed.

PR #989, opened 2017-11-21
Prevent crash if a sub-model row header is selected

### LINKED ISSUE #985: Most of 2017.3x will crash when doing this
>>>Accidental discovery<<<

In the sequencer, toggle strands on a model with submodels
Select a submodel then toggle strands again
Crash

computergeek1507:
Prevent crash if a sub-model row header is selected when clicking Toggle Stands. Fixes #985

### DIFF
--- a/xLights/sequencer/RowHeading.cpp
+++ b/xLights/sequencer/RowHeading.cpp
@@ -617,7 +617,7 @@ void RowHeading::OnLayerPopup(wxCommandEvent& event)
     else if (id == ID_ROW_MNU_TOGGLE_STRANDS) {
         ModelElement *me = dynamic_cast<ModelElement *>(element);
         if (me == nullptr) {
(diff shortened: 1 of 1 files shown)


keithsw1111:
Are you sure this is right? Does it break the toggling behaviour on strands? I cant test it for another 12 hours.

computergeek1507:
I didn't change the current functionality but I'm unsure it is correct to begin with.

Current Version:

toggle crash


With My Change Version:

toggle fix

computergeek1507:
StrandElement class is a subclass of SubModelElement so the cast should be good. The method to get the parent model is part of submodel class. I originally just added another dynamic cast to check for a null object when a submodel row was selected but since the StrandElement and SubModelElement row headers do the same action (just collapse the addition rows) I didn't want to add an unnecessary dynamic cast.

toggletree

AzGilrock:
No problem just asking.  I wasn't at home to test it.

PR #1007, opened 2017-12-03
Allow Negative Value in the Tilt Orientation Field

### LINKED ISSUE #1005: Allow Negative Value in the Tilt Orientation Field
Moving heads with a Tilt Degree of Rotation greater than 180 degrees need to have a negative value in the Tilt Orientation field in order to accurately show in the preview that the mid-point of the tilt value range is straight up.

computergeek1507:
Fixes #1005

### DIFF
--- a/xLights/models/DmxModel.cpp
+++ b/xLights/models/DmxModel.cpp
@@ -183,8 +183,8 @@ void DmxModel::AddTypeProperties(wxPropertyGridInterface *grid) {
     p->SetAttribute("Max", 360);
     p->SetEditor("SpinCtrl");
 
-    p = grid->Append(new wxUIntProperty("Tilt Deg of Rot", "DmxTiltDegOfRot", tilt_deg_of_rot));
-    p->SetAttribute("Min", 0);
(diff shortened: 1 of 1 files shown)


AzGilrock:
No one needs a negative value.  If Matt needed -36 degrees orientation then he should type in 224.  It's a 0 to 360 field so every orientation is possible with positive numbers.  I first want to see proof that using the equivalent positive number doesn't work.

AzGilrock:
I was waiting to hear back on Matt about the issue to make sure there not something I'm missing.

keithsw1111:
Any word from Matt yet?

AzGilrock:
No I guess he was happy to edit the XML with a negative value which works so he never tried using 360 minus that value like I suggested.  I tried it and it seemed to work but I wasn't moving a real head so wanted to make sure there isn't some hidden issue.  The whole idea of the orientation value was to be able to type in any angle from 0 to 360 to define the installed location.

AzGilrock:
So using a calculator you can confirm that the sin or cos of X equals that of (360 - X) so negative values are not needed.

PR #1008, opened 2017-12-03
Fixing Spinners Channel Count

### LINKED ISSUE #1003: Spinner Model Bug
This is a bug that was discussed in one of the Zoom meetings (Nov 30 in the evening US time).  Not sure if the guy who found it knew how to submit it, or if Keith asked to have it submitted (he was on the call), but figured I would add it just in case.

In the spinner model the number of channels is not being calculated properly.  For example, with # Strings set to 3, Arms/String as 2, and Lights/Arm as 25, it should be 150 nodes = 450 channels.  However the start and end channels show it as 300 channels or 100 nodes.  If clicking on the node layout, only 75 node (from 75 to 150) are displayed.  The wiring diagram appears correct however.

When putting an effect on this model at the strand level:
Strand 1 works as expected.
Any effect on Strand 2 OR 3 will always light up both strands.
Any effect on Strand 4 OR 5 will always light up both strands.
Strand 6 works as expected.

computergeek1507:
fixes #1003

### DIFF
--- a/xLights/models/SpinnerModel.cpp
+++ b/xLights/models/SpinnerModel.cpp
@@ -369,5 +369,5 @@ int SpinnerModel::CalcCannelsPerString() {
     if (SingleNode) {
         return GetNodeChannelCount(StringType);
     }
-    return GetNodeChannelCount(StringType) * parm2;
+    return GetNodeChannelCount(StringType) * parm2 * parm3;
 }


keithsw1111:
Does this change still work for say single channel or dumb node spinners where the number of bulbs per arm dont change the number of channels? I just am not sure so wanted to check you had tested those configurations.

computergeek1507:
The channel count math wasn't taking into account the number of strings parameter. I dint change the Channel Count math for the node types. (It's still using the standard method in Model.cpp) If you set 3 Strings of Single Channel Lights it uses 3 channels and 9 channels for dumb RGB nodes.

**Single Color:**
image

**3 Channel RGB:**
image

Issue #991, opened 2017-11-22
Batch render in 2017.37 failure


lithgowlights:
My show directory is Z:/lithgow_lights/ and sequences are in Z:/lithgow_lights/sequences/ directory. Batch render sees them, and allows me to tick the files but when it renders them, it places them in Z:/lithgow_lights/ directory.

Also the rendering does not complete properly. I have one test sequence used when installing the lights that is in the Z:/lithgow_lights/ directory and all the show sequences have the same fseq file size as that test animation, much smaller than the 200-600MB they should be. 

Tried both 1 and multiple files and they all fail to render correctly

cp16net:
i can confirm this happens on mac as well.

PR #1009, opened 2017-12-03
Fix for Batch Render dialog if the Sequence File is in a subfolder directory.

computergeek1507:
wxFileName would throw an error if the file name from batch render had a sub folder appended to the front. I changed it to create the full file path instead and added some logging if the file path doesn't exist.


computergeek1507:
fixes #1009

computergeek1507:
fixes #991

PR #1014, opened 2017-12-07
Check for 32 bit version of xLights on a 64 bit Computer.

computergeek1507:
Check for 32 bit version of xLights on a 64 bit Computer. It uses wxConfigBase to persist a bool property to prevent the dialog from displaying at every launch, just on first launch of the app.


keithsw1111:
Given the whole 32 vs 64 bit version thing is really just a windows thing I would use #ifdef and only call this code if it is running on windows.

Issue #1021, opened 2017-12-09
Occasional crashes when modifying effects in AC mode


nkaminski:
Hello,
With xLights 2017.39 and 2017.40, I am experiencing occasional crashes when modifying effects in AC mode. All effects in question are single color on/off/ramp effects applied to line models.

I can reproduce this consistently via the following steps: 
Enter AC mode, replace (individually) several ramp effects with intensity effects in rapid succession and this crash will occur. 

I am running xLights on Arch Linux, fully updated as of 12/8/17.

Looking at the debug reports. is appears that the crash is consistently occurring within xLightsFrame::SetEffectControls.

Debug Reports:
[xLights_dbgrpt-31273-20171208T004825.zip](https://github.com/smeighan/xLights/files/1544425/xLights_dbgrpt-31273-20171208T004825.zip)
[xLights_dbgrpt-1152-20171208T235047.zip](https://github.com/smeighan/xLights/files/1544426/xLights_dbgrpt-1152-20171208T235047.zip)
[xLights_dbgrpt-24885-20171209T001609.zip](https://github.com/smeighan/xLights/files/1544427/xLights_dbgrpt-24885-20171209T001609.zip)

nkaminski:
I can reproduce this consistently via the following steps: 

Enter AC mode, replace (individually) several ramp effects with intensity effects in rapid succession and this crash will occur.

keithsw1111:
Yeah. I think I will have this fixed in .41. I just can't be sure.

nkaminski:
@keithsw1111 Any insights as to where/how it happens? It seems like a use-after-free case, possibly where member functions of an effect are called on an effect that has been deleted.  I have a fairly strong C++ background and have been digging into this myself as well.

nkaminski:
Log from GCC7 + libasan memory safety analyser:
==24016==ERROR: AddressSanitizer: heap-use-after-free on address 0x613000075b60 at pc 0x5591a185d581 bp 0x7fff5e281010 sp 0x7fff5e281000
READ of size 8 at 0x613000075b60 thread T0
    #0 0x5591a185d580 in Effect::GetEffectName[abi:cxx11]() const sequencer/Effect.cpp:278
    #1 0x5591a1ca37c1 in xLightsFrame::SelectedEffectChanged(SelectedEffectChangedEvent&) sequencer/tabSequencer.cpp:802
[...]

PR #1022, opened 2017-12-10
Resolve crash due to out of order events in AC mode

nkaminski:
This PR is a potential solution for the crashes I am experiencing in issue #1021. Since the issue appears to be a result of the wxWidgets handling the SelectedEffectChanged and mouse button up events in a nondeterministic order, which risks a crash in AC mode due to a use-after-free scenario, my proposed fix is to force both custom events fired as a result of the mouse down handler to occur synchronously if AC mode is enabled.

### DIFF
--- a/xLights/sequencer/EffectsGrid.cpp
+++ b/xLights/sequencer/EffectsGrid.cpp
@@ -5674,7 +5674,15 @@ void EffectsGrid::RaiseSelectedEffectChanged(Effect* effect, bool isNew, bool up
     }
     // Place effect pointer in client data
     SelectedEffectChangedEvent eventEffectChanged(effect, isNew, updateUI);
-    wxPostEvent(GetParent(), eventEffectChanged);
+
+    //Mouse events and custom events do not always occur in the correct order, which causes crashes in AC mode.
+    if(IsACActive()){
+        //Therefore perform the actions synchronously
+        GetParent()->GetEventHandler()->ProcessEvent(eventEffectChanged);
+    }
+    else{
+        wxPostEvent(GetParent(), eventEffectChanged);
+    }
 }
 
 void EffectsGrid::RaisePlayModelEffect(Element* element, Effect* effect,bool renderEffect)
@@ -5685,7 +5693,15 @@ void EffectsGrid::RaisePlayModelEffect(Element* element, Effect* effect,bool ren
     playArgs->effect = effect;
     playArgs->renderEffect = renderEffect;
     eventPlayModelEffect.SetClientData(playArgs);
-    wxPostEvent(GetParent(), eventPlayModelEffect);
+
+    //Mouse events and custom events do not always occur in the correct order, which causes crashes in AC mode.
+    if(IsACActive()){
+        //Therefore perform the actions synchronously
+        GetParent()->GetEventHandler()->ProcessEvent(eventPlayModelEffect);
+    }
+    else{
+        wxPostEvent(GetParent(), eventPlayModelEffect);
+    }
 }
 
 void EffectsGrid::RaiseEffectDropped(int x, int y)


keithsw1111:
Ok ... so the problem is that we are passing a pointer to an effect in a message and by the time the effect is being processed it is not necessarily guaranteed that the pointer will still be valid.

If this were as simple as guaranteeing message order then all crashes due to an invalid effect pointer would occur the first time the effect pointer was used ... but this does not seem to be the case based on my analysis yesterday of crashes relating to this.

The one I implemented does not solve the problem you are seeing ... in fact I specifically noted that I would not catch race conditions where an effect was being deleted ... only where it was being unselected.

There is a downside of ProcessEvent ... and that is that the message is not processed on the UI thread. This limits the actions that can be taken in those functions as this causes issues on some operating systems.

Consequently we are going to need to look harder at this one.

nkaminski:
OK, it's clear that we came to the same conclusion that this is indeed a use-after-free where a given effect pointer is invalidated before all outstanding events referencing it have been processed.

Another potential (and likely better after seeing your response) solution I thought of was to have a member variable in the effect class that counts the number of pending actions against such effect, as well as a boolean "valid" flag, initially set to true.

When an attempt is made to raise an event referencing an effect, only actually post the event if the referenced effect is marked valid.

If an effect is deleted with pending actions, mark the effect as invalid, but do not delete/free effect class instance.

When handling any custom event referencing an effect, decrement the pending action count, then check if the effect is marked valid. 

If yes, continue as normal. 

If no, delete/free the effect only if the pending action count is zero, and return immediately regardless.

Thoughts?


On December 10, 2017 1:51:45 PM CST, Keith Westley <notifications@github.com> wrote:
>Ok ... so the problem is that we are passing a pointer to an effect in
>a message and by the time the effect is being processed it is not
>necessarily guaranteed that the pointer will still be valid.
>
>If this were as simple as guaranteeing message order then all crashes
>due to an invalid effect pointer would occur the first time the effect
>pointer was used ... but this does not seem to be the case based on my
>analysis yesterday of crashes relating to this.
>
>The one I implemented does not solve the problem you are seeing ... in
>fact I specifically noted that I would not catch race conditions where
>an effect was being deleted ... only where it was being unselected.
>
>There is a downside of ProcessEvent ... and that is that the message is
>not processed on the UI thread. This limits the actions that can be
>taken in those functions as this causes issues on some operating
>systems.
>
>Consequently we are going to need to look harder at this one.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/smeighan/xLights/pull/1022#issuecomment-350576581

-- 
Sent from my Android device with K-9 Mail.

keithsw1111:
The problem as you describe it can leave behind effects which then need to be garbage collected. You also have to worry about what does it mean to be taking an action on a deleted effect ... could that cause other issues.

There are other options ... one is to check the effect pointer is valid before using it ... problem with this is it may still become invalid while it is in use.

Another option is to implement a lock around or even in each effect ... causing threads to block until the effect is released.

I really need to spend some time looking at all the places it is called to decide the best approach.

keithsw1111:
I think I have a solution ... buts its a scary change so I am not putting it on the master branch right now. Basically I am going to lock the effect layer on receipt of the message and then validate the pointer before i use it. I say it is a scary change because it is really easy to make a mistake in this. If you would like to test it say so and i will push the branch to github.

nkaminski:
Please do push such branch. I would certainly like to test it :)

Nash Kaminski

On December 11, 2017 1:45:52 AM CST, Keith Westley <notifications@github.com> wrote:
>I think I have a solution ... buts its a scary change so I am not
>putting it on the master branch right now. Basically I am going to lock
>the effect layer on receipt of the message and then validate the
>pointer before i use it. I say it is a scary change because it is
>really easy to make a mistake in this. If you would like to test it say
>so and i will push the branch to github.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/smeighan/xLights/pull/1022#issuecomment-350645762

-- 
Sent from my Android device with K-9 Mail.

nkaminski:
Okay, sounds good. I plan to sequence a couple more songs and will use this branch (as opposed to my workaround) and let you know how it goes. Will save very frequently of course and let you know if I run into any deadlock cases or crashes.

keithsw1111:
So after some discussions with Dan ... our guru when it comes to rendering we have decided on a different course of action to solve these problems. My branch will be abandoned.

nkaminski:
Okay good to know. Let me know if you have any other new changes you would like me to test or even implement. 

As a side note, keep up the great work! Xlights is by far my light sequencer of choice.

On December 11, 2017 5:38:51 PM CST, Keith Westley <notifications@github.com> wrote:
>So after some discussions with Dan ... our guru when it comes to
>rendering we have decided on a different course of action to solve
>these problems. My branch will be abandoned.
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/smeighan/xLights/pull/1022#issuecomment-350894930

-- 
Sent from my Android device with K-9 Mail.

keithsw1111:
Ping me on facebook ... if you want to code xLights there is a massive backlog of changes.

keithsw1111:
It would be useful if you could test Dan's latest checkin on the main branch to see if this has resolved this problem as we believe it should have.

nkaminski:
Still seeing this crash as of release 2017.42:

xLights version 2017.42 64bit

xLights(_Z11handleCrashPv+0xc35) [0x55a824154675]
xLights(wxFatalSignalHandler+0x19) [0x55a8246d76c9]
/usr/lib/libpthread.so.0(+0x11da0) [0x7fab770c6da0]
/usr/lib/libc.so.6(+0x96496) [0x7fab72682496]
/usr/lib/libc.so.6(+0xaea32) [0x7fab7269aa32]
xLights(_Z7wxMB2WCPwPKcm+0x6e) [0x55a8246ac3de]
xLights(_ZNK8wxMBConv7ToWCharEPwmPKcm+0xc5) [0x55a82466e585]
xLights(_ZNK8wxMBConv6cMB2WCEPKcmPm+0x40) [0x55a82466eee0]
xLights(_ZN8wxString10ConvertStrEPKcmRK8wxMBConv+0x4d) [0x55a82467d70d]
xLights(_ZN8wxStringC2EPKc+0x42) [0x55a823e97232]
xLights(_ZN12xLightsFrame17SetEffectControlsERK11SettingsMap+0xa3) [0x55a8240a0003]
xLights(_ZN12xLightsFrame17SetEffectControlsERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEES7_RK11SettingsMapSA_b+0x153) [0x55a8240a0443]
xLights(_ZN12xLightsFrame21SelectedEffectChangedER26SelectedEffectChangedEvent+0xb9) [0x55a8240a07d9]
xLights(_ZN12wxEvtHandler23ProcessEventIfMatchesIdERK21wxEventTableEntryBasePS_R7wxEvent+0x6b) [0x55a8246fa23b]
xLights(_ZN12wxEvtHandler23SearchDynamicEventTableER7wxEvent+0x1bf) [0x55a8246fa7bf]
xLights(_ZN12wxEvtHandler11TryHereOnlyER7wxEvent+0x1f) [0x55a8246faa7f]
xLights(_ZN12wxEvtHandler10DoTryChainER7wxEvent+0x43) [0x55a8246fa4e3]
xLights(_ZN12wxEvtHandler12ProcessEventER7wxEvent+0x3d) [0x55a8246fab5d]
xLights(_ZN12wxWindowBase8TryAfterER7wxEvent+0xab) [0x55a8248837db]
xLights(_ZN12wxWindowBase8TryAfterER7wxEvent+0xab) [0x55a8248837db]
xLights(_ZN12wxWindowBase8TryAfterER7wxEvent+0xab) [0x55a8248837db]
xLights(_ZN12wxEvtHandler20ProcessPendingEventsEv+0x152) [0x55a8246fc712]
xLights(_ZN16wxAppConsoleBase20ProcessPendingEventsEv+0x67) [0x55a8245d38a7]
xLights(_ZN5wxApp6DoIdleEv+0xad) [0x55a82472b8cd]
xLights(+0x10be9f0) [0x55a82472b9f0]
/usr/lib/libglib-2.0.so.0(g_main_context_dispatch+0x15e) [0x7fab74a270be]
/usr/lib/libglib-2.0.so.0(+0x6cf69) [0x7fab74a28f69]
/usr/lib/libglib-2.0.so.0(g_main_loop_run+0xd2) [0x7fab74a29f42]
/usr/lib/libgtk-x11-2.0.so.0(gtk_main+0xb7) [0x7fab76ba33a7]
xLights(_ZN14wxGUIEventLoop5DoRunEv+0x25) [0x55a82473ef95]
xLights(_ZN15wxEventLoopBase3RunEv+0xa4) [0x55a824603194]
xLights(_ZN16wxAppConsoleBase8MainLoopEv+0x5c) [0x55a8245d115c]
xLights(_Z7wxEntryRiPPw+0x41) [0x55a824648d71]
xLights(main+0xa6) [0x55a823e50d16]
/usr/lib/libc.so.6(__libc_start_main+0xea) [0x7fab7260cf4a]
xLights(_start+0x2a) [0x55a823e87c5a]

Crashed thread id: 0x7AABB9C0 or 2058074560

keithsw1111:
I have included this ... slightly different but net impact is the same.

PR #1037, opened 2017-12-20
randomize is working better now

cp16net:
* added the ability to randomize the currently selected effect settings and locks apply
* create random effects acutally randomizes the settings in the effect now


AzGilrock:
Why did there need to be changes to the UpdateEffect function?  We allow editing while a sequence is playing so I don't think we want PlayType to get changed because you could be in the middle of model playback.

cp16net:
@AzGilrock technically there should be no change in the update effect function but the git diff shows a change in it because i based the event button off it and they are close to each other. i think the only change to the update effect is some whitespace was deleted on save of the file.

AzGilrock:
Ok yep that confused me cause it looked like added code.  I see it already there so no problem.  I need to wait till I'm at home to look more into this change.

cp16net:
No worries since this is the first "real" feature i've added i'd love to make sure i'm not missing something, i tested the paths that i know of which are using the new button i created and using the create random effects from the grid option. I don't know of another entry point to the code. Learned a ton about how the system works in the process. :)

xLights/EffectsPanel.cpp
```diff
@@ -174,6 +189,9 @@ wxString EffectsPanel::GetRandomEffectStringFromWindow(wxWindow *w, const wxStri
 			s += AttrName + wxString::Format(wxT("%i"), v);
 		} else if (ChildName.StartsWith("ID_CHOICE")) {
             wxChoice* ctrl=(wxChoice*)ChildWin;
+            if (ctrl->GetCount() <= 0) {
```
cp16net: not sure exactly why but sometimes i noticed this would crash when creating lots of random effects from a timing track selection and the count was 0 which caused a divide by zero error below.

xLights/EffectsPanel.cpp
```diff
@@ -210,7 +234,8 @@ wxString EffectsPanel::GetRandomEffectString(int effidx)
 
     // get effect controls
     wxWindow *window = EffectChoicebook->GetPage(effidx);
-    s += GetRandomEffectStringFromWindow(window, prefix);
+    wxWindow *wPanel = GetWindowPanel(window);
```
cp16net: Here is why the there were never any random settings generated, this was passing the wrong window with all the settings.

For reference i found this commit looks like it broke the randomization. https://github.com/smeighan/xLights/commit/b2e87fbd9f0151dc5610bb6ffe70f018504701f8

xLights/sequencer/tabSequencer.cpp
```diff
@@ -1502,24 +1502,75 @@ void xLightsFrame::UpdateEffect(wxCommandEvent& event)
...
+        int endms = -1;
+
+        for(int j=0;j< el->GetEffectCount();j++) {
+            if(el->GetEffect(j)->GetSelected() != EFFECT_NOT_SELECTED)  {
+                std::string effectName = el->GetEffect(j)->GetEffectName();
+                int effectIndex = el->GetEffect(j)->GetEffectIndex();
+                std::string settings, palette;
+
+                settings = EffectsPanel1->GetRandomEffectString(effectIndex);
+                palette = colorPanel->GetRandomColorString();
```
cp16net: these 2 lines and the last line of this method are the only diff between the randomize and update function above.

AzGilrock:
My big complaint is when you select a big block of cells it takes forever to populate random effects.  I tried to speed it up but ran into a roadblock with the design.  Everything was tied into the darn panel object.

cp16net:
yeah i saw a comment about speed in there that was commented out. i wondered if each effect could return its available settings instead of going through the window objects to get them. Not sure if the time is spent rendering or generating. but there should be no need to render each generated effect when create lots of them.

PR #1055, opened 2018-01-02
Model layout print image

KMojek:
This adds capability to print out the canvas in the Model Layout tab
Initially, I set it up to simply print a canvas-dimensions copy; later I made some updates so that the ModelLayout canvas can be rendered at arbitrary dimensions, so I render the image at something closer to printer-friendly dimensions, which arguably looks a little nicer.

Demo video available here: https://vimeo.com/249436627

Now that I look at the results some more, it might be better to just send the canvas-dimensions version to the printer; it seems like it might be useful elsewhere to be able to render the ModelPreview at arbitrary dimensions though?

### DIFF
--- a/xLights/DrawGLUtils31.cpp
+++ b/xLights/DrawGLUtils31.cpp
@@ -67,6 +67,19 @@ PFNGLDISABLEVERTEXATTRIBARRAYPROC glDisableVertexAttribArray;
 PFNGLGETATTRIBLOCATIONPROC glGetAttribLocation;
 PFNGLPOINTPARAMETERFPROC glPointParameterf;
 
+PFNGLGENFRAMEBUFFERSPROC glGenFramebuffers;
+PFNGLBINDFRAMEBUFFERPROC glBindFramebuffer;
+PFNGLDELETEFRAMEBUFFERSPROC glDeleteFramebuffers;
+PFNGLISFRAMEBUFFERPROC glIsFramebuffer;
+PFNGLFRAMEBUFFERPARAMETERIPROC glFramebufferParameteri;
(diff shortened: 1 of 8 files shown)


dkulp:
Definitely not working on the Mac... :(

Couple of issues:  the context menu never pops up.   There are already a rightClick and rightDown events registered on the modelPreview and they seem to take precedence on the Mac.  If I DON'T register those, I can get the popup, but we obviously need those.  I would move the popup menu into the OnPreviewRightDown that is already there and used to create popup menus.  

Second, the printout (and the save as image thing is doing this as well) is not printing much.  I THINK it's not taking the content scaling factor into account as it looks like it's just printing the bottom quarter of the window on my machine.

KMojek:
Thanks @dkulp for the feedback. I have a Mac running Xcode 8.3.3 but haven't tried to build xLights on there yet. It sounds like the changes that will be required are pretty straightforward; I'll try making those and pushing up another commit.

dkulp:
The content scaling stuff seems to be easy...

`                origSize = wxSize(mWindowWidth, mWindowHeight);
-               mWindowWidth = size.GetWidth();
-               mWindowHeight = size.GetHeight();
-               float mult = float(size.GetWidth()) / origSize.GetWidth();
+               mWindowWidth = ((float)size.GetWidth() / GetContentScaleFactor());
+               mWindowHeight = ((float)size.GetHeight() / GetContentScaleFactor());
+               float mult = float(mWindowWidth) / origSize.GetWidth();
`

Also note that all Mac's are OpenGL3 so the check for that is irrelevant (and should always just return true).

dkulp:
Looks like Keith already fixed the popup menu issue....   I'll get the rest of this merged over.

PR #1059, opened 2018-01-04
Added Right Click Ping controller Option To Setup Tab

computergeek1507:


keithsw1111:
Rather than checking for multicast ... doesnt output have a method that returns whether or not the output supports ping?

computergeek1507:
Currently The SerialOutput Class CanPing() method returns true. I didn't know if this was a mistake or if "Ping" means xLights supports communication(outuput) with the device in general. I didn't find anywhere in xLights that it was used but I didn't know about xSchedule.

image

keithsw1111:
Ping on serial validated that the port is open or can be opened. Ok so it isn’t a real ping but it does check the output exists and works. So yes that is correct.

computergeek1507:
I updated it to "Ping" all pingable controllers

PR #1062, opened 2018-01-07
Export house preview video

KMojek:
Adds a file-menu item to export the house-preview as a video.

This is more-or-less working but needs some more work; saving that for now as this PR is already larger than I'd like. Some things left to clean-up:
* Audio - there seems to be a bit of static in the exported audio
* Audio - volume in the exported audio doesn't quite match audio when previewed in xLights
* Audio - currently exporting mono-only
* Audio - still one place where it is assuming a 20-fps sequence; needs to work for 40-fps
* Video - dimensions match the house-preview panel; should have options to match layout dimensions and/or standard video dimensions (1280p, 720p)
* Etc - Needs a progress dialog during export
* Etc - Some code-cleanup needed; some redundant stuff being passed around, etc.


KMojek:
It'd also be nice to move VideoExporter into a separate source/header. Not sure at this point if there's an easy way to add new files in a way that won't break things on Mac, Linux,  or MinGW on Windows so punted on that.

AzGilrock:
We just add files and the Mac and Linux guys update their project file to fix their builds.

keithsw1111:
Before we merge it would be nice to fix some of these up ... specifically:

Audio - there seems to be a bit of static in the exported audio
Audio - volume in the exported audio doesn't quite match audio when previewed in xLights - I assume the audio is just 100% ... if so I think that is ok.
Audio - currently exporting mono-only - It would be nice if this could be addressed but probably not essential
Audio - still one place where it is assuming a 20-fps sequence; needs to work for 40-fps
Etc - Needs a progress dialog during export

keithsw1111:
It would definitely be good to break out your new class into its own video writing class file.

Otherwise code looks good.

keithsw1111:
Making good progress ... only

Audio - volume in the exported audio doesn't quite match audio when previewed in xLights - I assume the audio is just 100% ... if so I think that is ok.
Audio - currently exporting mono-only - It would be nice if this could be addressed but probably not essential
Break out into separate source files

Are really the only things left and only the last one is really important.

KMojek:
Okay, I've added commits which address all the issues except for audio still being mono-only. I'll need some help from the Mac and Linux guys since I broke their builds as mentioned above. I think the volume difference was just an artifact of the static-in-audio problem; it seems to match up with audio in xLights at 100% now.

keithsw1111:
So do you want to sort the mono vs stereo or merge now?

keithsw1111:
BTW ... the linux build builds off the codeblocks file ... if you got that right Linux should still build ok.

KMojek:
I think I'd like to merge now if that's okay... I want to make some updates for mono vs stereo and the export dimensions but I may not  get to those for a few days.

PR #1067, opened 2018-01-11
Add more error-checking and logging to VideoExport

KMojek:


KMojek:
I'm a little confused by what I'm seeing in Keith's video of the video export hanging; it looks like the model preview and house preview panes are continuing to update while the export is in progress. From what I can tell though, those updates are always done on the main thread, so if those are running, my guess is that the export silently bailed out but we're somehow blissfully unaware so continue to show the progress at 0% forever (or until it's canceled, if it's allowing for that). So this commit is throwing a bunch of stuff against the wall and hoping something sticks. It adds:
* debug logging to xlGLCanvas::CaptureHelper - if we are early-returning because we think something is screwy with OpenGL3, that will be logged
* VideoExporter - add more error logging and create the progress dialog as a child of the xLights main-frame; I doubt that the dialog's parent is the problem here but seems like a good practice
* Capture the sequencer's play-status before starting the export, set it to stopped before starting the export, and restore it afterward. If we are somehow updating preview windows while a video export is in progress, that should solve that problem.

KMojek:
If the logging doesn't reveal anything useful, I think the behavior in Visual Studio was to crash rather than hang. If it's easy to get a capture of the callstack at the time of the crash, that might be helpful.

keithsw1111:
Are you opening the progress dialog on the main thread ... it seems to crash in VS when you open it and it does not seem to be on the UI thread.

keithsw1111:
From my logs ... I am writing an animation video using the release version:

2018-01-11 21:58:28,607 3520 log_base [DEBUG] Writing house-preview video to C:\Users\keith\Desktop\sdd.mp4.
2018-01-11 21:59:00,971 3520 log_base [DEBUG] Headers written ... writing the video 600 frames.
2018-01-11 21:59:00,971 3520 log_base [DEBUG]     Writing frame 0.
2018-01-11 21:59:00,978 3520 log_base [DEBUG] Finished writing house-preview video.

Doing it with a audio file same result.

keithsw1111:
Ok ... I kind of have it working. It still bombs in VS but it now at least runs in gcc. I wonder if it is the lambda stuff ... although I dont understand why it works for you and not for me.

JeffDC1979:
Keith, seems Kevin is getting the same VS errors I am seeing in release. The dll's you forwarded me did not seem to help at all.

keithsw1111:
The thread issues are to do with progress dialogue. basically it crashes in the thread startp functions. As long as the release build in gcc works i am good. The rest can be sorted later.

keithsw1111:
Of course they didnt jeff ... i told you i only do debug in vs. To do release you need to rebuild the dependencies. Althought in the case above the fix is pretty obvious.

keithsw1111:
More commentary ... from Dan Kulp:

With some basic experiments of trying to re-configure ffmpeg, I can only get h264 encoding if I enable the VideoToolbox stuff.  However, that wasn’t introduced until OSX 10.8.  Thus, we’d have to drop support for 10.7.   Its something I’ve been thinking about doing, but not something I was planning on right now.

I wonder if it’s possible that if h264 isn’t available if we could use mpeg4 or mpeg2video or something so it would at least work.

There are a few other problems as well…. Even with the a video-encoder, it fails on a Retina Mac as the video cannot be captured as it’s not handling the content-scale-factor correctly.

dkulp:
I just pushed a change to use the codecs from the detected/guessed format which (once the scale factor issue was fixed) results in an MP4.   That said, the MP4 isn't playable in Quicktime for some reason.   I can use Handbrake to re-encode it and that works so I'm not really sure what the issue is.  :(

KMojek:
Okay, maybe it'd be best to just punt on this one for the time being (disable/hide it since the code is already merged or whatever is easiest). I made a couple of smaller changes with minimal problems, but it seems like I should maybe get to a point where I'm either able to build and test on Windows with MinGW or on Mac before tackling a larger feature like this one. Sorry for the hassle and thanks for the feedback!

keithsw1111:
Given the function is not core ... people dont have to use it ... I would leave it in place and polish it over several releases.

Kevin ... having teething issues with new ambitious features is not uncommon and as long as it doesnt break the core we can work through it.

PR #1079, opened 2018-01-22
randomize only effect settings with locks

cp16net:
* add a tooltip with message saying what exactly is randomized
* method to check for which settings can be randomized (only with locks)
* update all the effects lock id's


cp16net:
Well now by default there are no `ID_BITMAP_TEXTCTRL_*` locks because it only uses the default `SLIDER`,`CHOICE`, or `CHECKBOX` for locks. We can talk about this a bit more later but it seems to be working now after a minor change i missed earlier.

PR #1091, opened 2018-02-08
Fix to use stand name if not empty, i.e polylines

computergeek1507:


computergeek1507:
In the Select Effect Dialog I have been working on it displays a list with models, submodels, and strands names. I created an override for GetFullName on the strand element class and it currently displays "Stand 1", "Stand 2", etc as the name in the display so the user can select and edit effects on the stand level. On models that allow Stand Naming i.e PolyLines, I thought it would be better to use the Stand Name set by the user not the generic name "Stand 1" when possible.

image

keithsw1111:
Gil ... looking for your call on this one. Should we merge.

AzGilrock:
Sure I don't think it affected PolyLines it was for the Effect Selection dialog.

PR #1094, opened 2018-02-17
Added try catch block to prevent crash for invalid regex in new Effect Search Dialog

computergeek1507:


keithsw1111:
Did you verify this actually works ... in both Visual Studio and CodeBlocks builds?

computergeek1507:
I can't get CodeBlocks to build yet so I only verified it worked with VS2017.

Issue #964, opened 2017-10-25
Single strand effect


ebuechner:
While selecting rainbow in the single strand effect the 3D fade ability is lost. I know it's been like this for a while and it's not really a big deal but it doesn't seem logical that changing the color palette would change other functions.

computergeek1507:
Should work now, let me know if you run into any issues.

PR #1095, opened 2018-02-18
Added 3D fade to Single Strand Effect for Rainbow Pallet.

computergeek1507:
#964 I also fixed a rounding error for color pallet selection. The current code does not evenly color the chase based on it's size and the number of colors selected. Two colors would look like this before:

image

Now it will color the chase evenly:

image


keithsw1111:
Dan ... looking for your call on this one ... should we merge?

PR #1100, opened 2018-02-26
Add sequence-video preview panel

KMojek:
I may have some more changes coming and am open to suggestions/advice but here is a quick video of how it's working in it's current state: https://vimeo.com/257424286

### DIFF
--- a/xLights/SequenceVideoPanel.cpp
+++ b/xLights/SequenceVideoPanel.cpp
@@ -1,4 +1,5 @@
 #include "SequenceVideoPanel.h"
+#include "VideoReader.h"
 
 //(*InternalHeaders(SequenceVideoPanel)
 #include <wx/intl.h>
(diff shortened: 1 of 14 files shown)


keithsw1111:
Looks good Kevin from the video. The video reader class does not cache the video at all ... the main reason being the cache ends up insanely large if the video is long. xSchedule does have a cachedvideoreader class if you want to experiment with it. Basically it reads and caches the video on a separate thread. This became necessary when trying to load and play multiple large videos without impacting frame rates.

KMojek:
Thanks for the merge; the additional changes should be okay to add later since this is going to be pretty hidden (pane is not shown by default and won't do anything unless you've got a video as your media file). The updates I think that should be made are:
* should it be synced with Model Preview? Seems like yes if you're displaying the same video onto a matrix but otherwise may not be useful.
* find a better initial docking location for the pane... or maybe it should just be floating by default if that's possible.

Having the video running didn't seem to slow things down much (just based on some quick Task Manager monitoring), so I'm not too worried about that currently. It could maybe be smarter though about caching the current frame, since we'll be updating at either 20 or 40 fps and video will likely be encoded at a different frame rate.

keithsw1111:
I believe it does hold onto the last frame so that should be ok.

keithsw1111:
I need you to look at my TODO in my checkin tonight. If i have a sequence with a video and then i create an animation or just close the sequence the video needs to be removed and the window blanked out. I have removed the video (or at least disabled it) but I am not sure the best way to clear the panel.

PR #1120, opened 2018-04-19
Added Folder Setting dialog for Media, FSEQ, and Backup Directories.

computergeek1507:


computergeek1507:
With the ability to change the backup folder, I thought the alternate backup was unnecessary. I can add it to the dialog.

keithsw1111:
I can see that point of view ... but in this case I think I would leave it in for now ... but move the setting. It is as you say quite possible people only use it because they dont want to back up their FSEQ files.

computergeek1507:
I kept the functionally that if you click "Do Alt Backup" and the folder setting is empty, it will prompt you to select a folder. Because of this I thought we should still save the setting on exit.

PR #1129, opened 2018-05-10
Fix for Issues of Loading the effects file before the network file

computergeek1507:
Dan found this in my last commit.


dkulp:
This was fixed in 36f16338f1ddbc19034213184849ed89d93fc433

PR #1135, opened 2018-05-15
Adding a warp effect

KMojek:
The warp effect is intended to work with the new "canvas" layer blending option. Rather than adding anything new, it will modify whatever was done in the layers below this effect.

With this initial commit, there are just two warp types and just "in" and "out" options, so the effect behaves like a transition.


KMojek:
The warp effects are currently pixel shader code adapted to run on the CPU. I’m hoping to get further along with the shaders and have them running on the GPU before too long; that should speed up their rendering substantially. The 3D layout stuff sounds like it’d be interesting to work on.

dkulp:
Getting OpenGL to work in an effect on the background thread is not going to be a trivial task.   We don't have any locks on the main thread around any of the OpenGL things.   Thus, any effect that renders using OpenGL is going to have to be flipped to the main thread which does have a significant performance penalty.

KMojek:
Hmm, okay I was sort-of assuming that it’d be okay on a background thread since it wouldn’t be rendering to a window and wouldn’t need to share any state with the other OpenGL windows. I’ll look at getting the current implementation running faster before messing with that, then. Thanks!

dkulp:
It can be done, it's just non-trivial.   Each thread would need it's own OpenGL context.  Since we create 10's of threads, that can be a lot.   However, you will want to share the shader program objects and such, thus, the contexts need a "parent" context for the shared things.    But you cannot "modify" that context on the background threads.   

Additionally, the OpenGL engine may or may not need special setup to enable multi-threaded use cases.   Also, OSX, Linux, and Windows all have different requirements for how to create the OpenGL contexts that can be used on the background threads.    Thus, it starts getting complicated fairly fast.

KMojek:
I think this should be safe to merge. I just rebased and it builds okay with VS and CodeBlocks. As it is currently, the effect only makes sense with canvas blend and mainly useful for matrices and dense pixel trees but shouldn't cause harm otherwise.

KMojek:
I'll probably add one or two more options before 2018.15. I don't know if the ripple one is all that useful so I'll likely change that from being the default.

KMojek:
I think this may be in a good state to merge now. Today I updated the default warp and added a new one. I probably won’t have time to work on it much more for a week or so. Here is an updated demo video: https://vimeo.com/271594174

keithsw1111:
Kevin ... should I be surprised that the warps don't seem to have settings? You seem to have several of them but no way to adjust what each one does?

Is that a side effect of how you implemented it? Or maybe there are not really adjustments that would make sense? Or maybe just something that we can add later?

keithsw1111:
I also note you have a whole bunch of changes to the project file that you should not be committing.

KMojek:
Oops, you're right, there's definitely some XLights.vcxproj stuff that should not be in there. I'll fix that.

Yeah, there is a lot that is hard-coded currently but it'd be nice to have settings for different warps. My thinking was to allow for a few initially and see which are useful. I think it should be easy to add those adjustments later.

keithsw1111:
Adding them later should be fine as long as you have the defaults which work the way the effects work now. If you cant then we should hold off merging.

So far no significant issues raised with 14 so I will likely do the merge over the weekend.

dkulp:
The only one that is working for me is the dissolve... The other three are completely black.

dkulp:
There are a bunch of issues with the OpenGL code.   That's disabled by default so not critical.  For one, gl_FragColor is not valid for OpenGL 3.3+.   Thus, none of the shaders compile.

KMojek:
Well this seems like a bit of a problem for xLights... apparently OpenGL is going to be deprecated on Mac: https://developer.apple.com/macos/whats-new/

KMojek:
Not pushed up just yet, but here's a "water drop" warp that may be useful over image effects. We'd definitely want to have settings for this one before adding it: https://vimeo.com/273624121

PR #1139, opened 2018-05-20
Added File Read and HTTP GET to Text Playlist Item in xSchedule

computergeek1507:
Demo:
https://youtu.be/domLOspLk-Q


keithsw1111:
The file use case I kind of get. The http one does not really make sense to me. Odds are the response is some sort of json or http type response. Surely we would need to do more to extract something useful.

computergeek1507:
I was thinking it would be better to make an HTTP event type and also add a way to pass data from a event to a playlist items property.

keithsw1111:
But an event can already call the text api

PR #1151, opened 2018-06-10
Remove OpenGL stuff from warp effect

KMojek:
The intention here was that since this may be an expensive effect to render, a shader implementation might be faster even though that would require that the rendering be done on the main thread. I'm not sure that it was actually faster though, and it was only working correctly on Windows, so it seems like it's more trouble than it's worth at this point.


KMojek:
Just fixing the build with this change but I'll have some more fiexs within the next few days.

KMojek:
Does the travis-ci build only build the master branch? The failure looks the same as before by changes with this PR.

PR #1157, opened 2018-06-14
Warp effect value curves

KMojek:
Allow for value curves on the X and Y settings so that the warps can be sort-of animated.

### DIFF
--- a/xLights/DrawGLUtils.cpp
+++ b/xLights/DrawGLUtils.cpp
@@ -7,6 +7,11 @@
  #include "OpenGL/gl.h"
 #else
  #include <GL/gl.h>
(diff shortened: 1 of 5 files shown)


KMojek:
Hmm, not sure what's going on with my system; I removed that change, did a full rebuild and am getting the same errors in DrawGLUtils.cpp... missing GL_COMBINE_RGB, GL_SOURCE0_RGB, etc.

KMojek:
It looks like the code that where I was getting the errors was just recently added: https://github.com/KMojek/xLights/commit/10c93b28f913ddc02b97361a7a3d5f8aec0dce35

PR #1159, opened 2018-06-16
Adding "single water drop" warp

KMojek:
One small addition to warp effect; short demo here - https://vimeo.com/275363942


KMojek:
I think it'd be possible for a single effect to do multiple random drops but deferring on that for now.

PR #1169, opened 2018-06-22
Adding 'drop' and 'wavy' warp types

KMojek:
Adding another warp type... not sure if this one will be super-useful but it was easy to implement.

drop fade


keithsw1111:
The drop is kind of cool but i dont understand why you make fade part of the effect itself. Surely it is better to let the user do the fade using brightness or a mask ... that way if the user just wants the drop they can have it.

KMojek:
Yeah, good point... I'll fix that.

KMojek:
Added one more warp type... not currently planning to add any more for now. Here's a short demo: https://vimeo.com/276565730

KMojek:
Waves on waves :smiley: https://vimeo.com/276567145

Issue #1179, opened 2018-07-01
Version 17+ On any prop only 1 pixel lights


jspiker77:
When using version 17+ when I drop an effect on any prop only 1 pixel lights up. This is true on the preview and the output to lights. I have a few other people test this with the same result. Attached is my packaged sequence for reference. 
[red_white_blue.zip](https://github.com/smeighan/xLights/files/2152347/red_white_blue.zip)

keithsw1111:
I dont see a problem. The single strand works fine. I dropped other effects with no problems.

darylc:
I can also replicate this

keithsw1111:
Even with Daryl’s file I still can’t replicate the problem in .19.

keithsw1111:
Maybe try a -w and reset all options.

jspiker77:
That seemed to resolve the issue. Thanks Keith!

keithsw1111:
I dont understand why it solved the issue and speaking with Daryl last night it didnt help his display.

PR #1183, opened 2018-07-03
fix for roto-zoom reset bug.

computergeek1507:
default value should be 10 not 1, because of divisor.


computergeek1507:
Fixes #1179

PR #1185, opened 2018-07-04
Added User Dictionary Editor

computergeek1507:
https://www.youtube.com/watch?v=2NvuSWp3th8

### DIFF
--- a/xLights/LyricUserDictDialog.cpp
+++ b/xLights/LyricUserDictDialog.cpp
@@ -5,12 +5,21 @@
 #include <wx/string.h>
 //*)
 
+#include "PhonemeDictionary.h"
+#include "xLightsApp.h"
+#include <wx/txtstrm.h>
+#include <wx/wfstream.h>
+#include <wx/filename.h>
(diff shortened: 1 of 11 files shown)


keithsw1111:
One thing I have always found useful when adding words to the custom dictionary is the ability to look up similar words and then copy those phonemes and tweak them. Would that be possible to include?

keithsw1111:
I didnt see the Youtube link ... :)

keithsw1111:
Do the words need to be upper case. I notice you had to backspace to change it. If they need to be can you just programatically force it.

What if I want to add a word ... which is similar to another word but different. So I add from the first word but then I want to look up the second so i can work out the phoneme change to make ... I guess I add a row and then copy what I need and then delete it

computergeek1507:
I cannot figure out who to force the text control to only accept upper case letters with out making a custom subclass, so I just force everything uppercase when you add them. I also added a read only text control to display the "Copy From" word's phenemes. The Text can be selected and copy for use in the grid view.
image

PR #1215, opened 2018-08-20
Fix bug with 'Export House Preview' on high DPI

KMojek:
Addresses https://github.com/smeighan/xLights/issues/1134

It looks like CaptureHelper and VideoExporter don't need to take GetContentScaleFactor() into account; that seems to be a fix for Windows; just want to verify it makes sense for Mac.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,7 @@ Issue Tracker is found here: www.github.com/smeighan/xLights/issues
 
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
+   -- bug (kevin)  Fix house preview video export on Windows with high DPI settings
    -- enh (keith)  Change the colour panel update button to purely apply the colour palette. All other colour settings are ignored.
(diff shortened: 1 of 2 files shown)


KMojek:
Fun fact: it looks like it *is* possible to repro this bug without a 4k monitor or Surface Pro. But... the usual way of setting DPI on Win 10 actually sets it per-monitor, so the "system DPI" is still reported as 100%. In order to repro it, you have to instead use the "Custom scaling" DPI option which seems to be system-wide and what xLights uses as the DPI setting. It also seems to be repro-able on Win 7 (with less trouble since per-monitor DPI wasn't a thing back then).

dkulp:
The contentScaleFactor things must remain for the Mac.   Thus you likely need to keep all the params and wherever they are passed in, use 1.0f on Windows and use the real ContentScaleFactor on the Mac.   Linux doesn't have HiDPI support at all so it would 1.0 for that as well.

KMojek:
Okay, fixed up; thanks @dkulp

PR #1233, opened 2018-09-02
Force a house-preview update after video export

KMojek:
This *seems* to mostly fix the issue where if you do a video export and then immediately close the sequence, the effects grid becomes unusable, even after opening or starting a new sequence. I can still get it to get in a state sometimes though where the range selection over the waveform either doesn't draw at all or just repeatedly draws a line in the same place. So this probably isn't a complete fix but seems like an improvement over the effects grid being completely unusable, so I think it's worth merging.

Note: so far, I've only been able to repro this on Win 10 but not on the Win 7 machine that I normally build xLights on.

### DIFF
--- a/xLights/xLightsMain.cpp
+++ b/xLights/xLightsMain.cpp
@@ -3195,6 +3195,10 @@ void xLightsFrame::OnMenuItem_File_Export_VideoSelected(wxCommandEvent& event)
         m_mgr->GetPane("HousePreview").Hide();
         m_mgr->Update();
     }
+    else
+    {
+       housePreview->Update();
+    }
 
     if (exportStatus)
     {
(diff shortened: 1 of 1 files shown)


keithsw1111:
I dont understand how this fixes the issue ... does the update function restore something that the video export breaks.

KMojek:
Yeah... I'm not completely sure why it seems to work; it seems like the capture-helper should be cleaning up correctly. One thing with the capture-helper is that it uses a single buffer for rendering; I *think* all the other OpenGL rendering is to double-buffered windows, so what is happening is that OpenGL is somehow thinking after the video export that one or more of those windows is not double-buffered anymore, so that's effectively disabling those windows (no buffer swap). I think a better fix may be to just forgo the extra temporary buffer for video export and just use the house preview's back buffer but that's a bigger change. I will give that a try, but this is all I had time for today.

PR #1234, opened 2018-09-03
Simplify video export

KMojek:
House Preview pane is forced to be visible and we're temporarily disabling the buffer swap, so using a renderbuffer for capture isn't really necessary.

This undoes my somewhat-hacky initial fix.

### DIFF
--- a/xLights/xLightsMain.cpp
+++ b/xLights/xLightsMain.cpp
@@ -3184,9 +3184,7 @@ void xLightsFrame::OnMenuItem_File_Export_VideoSelected(wxCommandEvent& event)
     }
     );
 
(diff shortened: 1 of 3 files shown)


KMojek:
Testing - repeated this three times:
* Start up xLights
* Load sequence
* Undock House Preview, move to second monitor, and resize to something appropriate for video export
* Export video
* Close sequence - before the two commits, this would leave artifacts behind in the sequencing grid
* Verify exported video is as-expected using Windows Media Player and VLC player
* Re-load sequence
* Verify interaction with sequence grid and waveform view works as expected - before the second commit, waveform view would sometimes not work correctly.

PR #1236, opened 2018-09-03
-Added Node Select Dialog to the SubModel, Faces, and State Dialogs

computergeek1507:
-Renamed Class to NodeSelectGrid from SubModelDrawGrid

https://youtu.be/e_clrljDzZE


keithsw1111:
Looks great ... but ...

1. Where nodes are in order increasing/decreasing can we instead use the range notation to reduce the numbers.
2. If I use your new row selection of nodes I am assuming it doesnt use the comma spacing ... which is good
3. Now this is hard ... but ... could the dialog also remember the order the nodes were selected ... as often the order of the nodes matters. It may be another mode ... or maybe not.

PR #1259, opened 2018-09-28
Making drag and drop seem more right.

computergeek1507:
Dropped item will move to new dropped location and other items will shift down one.


keithsw1111:
I am cool with changing what was there. It never felt quite right. Just before you merge make sure you test things like
- dragging when there is only one item in the list
- dragging beyond the first item
- dragging beyond the last item
- dragging the first item to the end
- dragging the last item to the start
- dragging a middle item to the start
- dragging a middle item to the end

dragging multiple items

Also did you do both advanced and simple mode?

PR #1317, opened 2018-11-06
fix bug 1314

tmeyermn:
The end value in spiral effect wasn't getting scaled.  slideridd wasn't getting set to true so OnTextCtrl_Paremeter2Text wasn't getting the multiplier.  Also, the slider wasn't getting scaled either in UpdateLinkedSlider.


keithsw1111:
There are several problems here which mean I am not going to accept this pull request.

1. You have included a lot of junk visual studio files that should not be checked in.
2. Making the change here causes issues with other value curves. This needs to be comprehensively tested with all value curves that have a divisor as changes have been made here in the past and then had to be reversed.

PR #1322, opened 2018-11-09
Fix bug where some text values were not getting multiplied by the divisor.

tmeyermn:
Found cases where text values and slider values were not matching.  slideridd was not getting set sometimes.  Also some cases where the multiplier factor wasn't getting set both on the text side and on the slider side.  Tested all the  types on different effects.


tmeyermn:
Closing this as it looks like Keith made some major changes that fix the issues, thanks @keithsw1111  :)

PR #1399, opened 2019-01-05
Add ability to pan in the X/Y plane when in 3D view.

tygunn:
Existing behavior of panning in the X/Z plane is retained when no modifier key is down.  This change adds the ability to hold down the CTRL key while using the middle mouse button (scroll wheel button) to pan the 3D layout or house preview views.  This provides the user with the ability to pan in the X/Y plane.

Adding X/Y plane panning handles the use case where the user wishes to move closer to a face of a mesh model they have imported (e.g. their house) and layout lights on that face.  The user would use X/Z plane panning to move close to a face of the model.  At this point, it is desirable to be able to pan across the face of the model in order to place lighting elements.  For intricate models (e.g. houses with lots of architectural details), being able to pan along the X/Y plane is helpful to move around the face of their model without having to rotate the camera and re-adjust panning in the X/Z plane.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -1002,7 +1002,7 @@ void LayoutPanel::refreshModelList() {
                     TreeListViewModels->SetItemText(item, Col_EndChan, endStr);
                 }
                 cv = TreeListViewModels->GetItemText(item, Col_ControllerConnection);
-                
+
                 std::string cc = model->GetControllerConnectionRangeString();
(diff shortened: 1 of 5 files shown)


tygunn:
Note: This is a change I had discussed with Gil in [Issue 1397](https://github.com/smeighan/xLights/issues/1397). 

I have seen other users on the xLights support forum ask how to pan along Y (e.g. [post](https://www.facebook.com/groups/xLights/permalink/2062120540490357/)).

PR #1410, opened 2019-01-11
Initial release Easylights controller communication

jhinklejr:
Release of EasyLights Controller config data communication


keithsw1111:
A few changes before we can merge.

Can you remove the library.
Can you explain your thinking behind the reset controller function?
Can we get rid of the //************************** surrounding your code ... it is just clutter
Can we remove the change to the platform SDK version in vcxproj
Can we remove your .user file
How sure are you the #pragmas are going to work on LInux and OSX?

jhinklejr:
See Below

jhinklejr:
Keith:

 

I went to remove the library – I confused – the xlights.lib – I thought you meant that one – is not in my fork directory.

 

What library are your referring to?

 

Thanks.

 

Joe

AzGilrock:
Probably this one:
image

keithsw1111:
Cant you reset and upload in the one action ... if necessary merge the inputs and outputs into a single upload like we had to do with the pixlite because of limitations in that controller. Having a separate reset is odd.

keithsw1111:
As the developer you need to do the legwork as much as possible to check cross platform support. Dan/Chris will deal with some oddities but they expect the code to be written for cross platform.

keithsw1111:
I use VS 2017 as well ... yet my SDK seems more current than yours ... we dont want to be going backwards ... or just leave that line out when you check in.

PR #1428, opened 2019-01-18
Allow the user to specify a manual X\Y position for the fireworks effect

davidmpeace:
Allows the user to optionally specify a manual X\Y position for the fireworks effect, rather than the default "random" position.
If checked, you may also use the value curve option as well that can move the fireworks position over time.

### DIFF
--- a/xLights/effects/FireworksEffect.cpp
+++ b/xLights/effects/FireworksEffect.cpp
@@ -57,6 +57,7 @@ public:
     float vel;
     float angle;
     bool _bActive;
+    bool _bManualLocationSet;
     int _cycles;
     int _colorindex;
     int startPeriod;
@@ -98,6 +99,18 @@ public:
         _colorindex = colorindex;
         startPeriod = start;
         _bActive = false;
+        _bManualLocationSet = false;
+    }
+    
+    void SetManualLocation(int x, int y)
+    {
+        if( !_bManualLocationSet ) {
+            _bManualLocationSet = true;
+            _x       = x;
+            orig_x = x;
+            _y       = y;
+            orig_y = y;
+        }
     }
 protected:
 private:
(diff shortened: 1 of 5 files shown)


keithsw1111:
Where is the wxs file for the UI changes?

davidmpeace:
I just pushed the wxs file changes -  Gil, you are correct, I'm making these changes within xcode manually on a mac.  What IDE are you guys using?

AzGilrock:
On Windows it's CodeBlocks.  I'm not sure if Dan uses it on MAC to do the GUI changes.  He could be using a Windows virtual machine as well.  But as long as you update the wxs file I guess it doesn't matter if its manually updated it should work.  Thanks.

keithsw1111:
Please dont manually edit wxs ... that wont end well.

cp16net:
Yeah i noticed on mac the version of codeblocks was old and screwed all kinds of things up that i didnt intend to when i was editing the GUI before. The mac version is REALLY old because there is no one in the codeblocks community to build and maintain it. I suggest windows for those updates unfortunately.

dkulp:
Unfortunately, I haven't found a codeblocks/Mac build that actually works.   :(
You can try the nightlies at: https://drive.google.com/drive/folders/0B2rFQ8rNHzEeN0xtU3R6emdhUWs

However, for me, the Dec-31 builds crash when loading the xLights project.   The Jan 12 build loads the project, but you cannot edit any of the panels as it's just a big black screen.  I think the latter is due to it only being a wx3.0.4 build which does not work well on Mojave so if you are using an older version of OSX, it might work better.

In general, I have a windows VM specifically for editing the xLights panels.  :(      We talked briefly about moving to QT, but that's a HUGE undertaking.   A company is currently paying some folks to get the wxQT port up to speed.   That may eventually be a bridge solution where we can use the wx wrappers, but then start using QT based things along with it.   Not really sure yet, and the work they are doing is still ongoing.

davidmpeace:
Thanks guys - I will get a windows VM spun up, and resubmit when I have CodeBlocks operational.

PR #1449, opened 2019-02-02
Enable timeline icon for galaxy effect / bug fix

KMojek:
Tom BetGeorge requested this earlier today on FB and a few others liked it. The post has since been deleted and Tom reached out to Keith directly instead. From what I can tell, the galaxy effect has never had an icon since Gil added it back in 2015. It seems like a harmless change.

I also included a small bug fix for #1450 in this branch.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,4 +1,4 @@
-xLights is a show sequencer and player/scheduler designed to control
+﻿xLights is a show sequencer and player/scheduler designed to control
 USB/DMX/sACN(e1.31)/ArtNET(e.1.17)/DDP controllers.
 xLights also integrates with the Falcon Player.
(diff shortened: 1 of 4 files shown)


KMojek:
Ah okay, SuperStar was before my time. 😃 If there are reasons for keeping some without icons, fine with me.

AzGilrock:
Well I'm not sure I'm opposed to it I'll need to see what it looks like.  Currently it does end up displaying just like a Morph so it's hard to tell which is which so it might be a good thing.  I'd just make sure it's displaying the tiny icon like Shockwave does.

AzGilrock:
Yeah that looks good.  Fan probably needs the same thing.  Morph is probably ok without it cause everyone expects it to look that way.

KMojek:
Okay, updated fan with the same change.

PR #1474, opened 2019-02-15
Update preview when toggling 3D in Layout tab

federicobond:
Otherwise the preview will not update until the user left clicks on it.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -5750,6 +5750,8 @@ void LayoutPanel::OnCheckBox_3DClick(wxCommandEvent& event)
     }
     obj_button->Enable(is_3d);
 
+    UpdatePreview();
+
     wxConfigBase* config = wxConfigBase::Get();
     config->Write("LayoutMode3D", is_3d);
     Refresh();


keithsw1111:
But it does update?

federicobond:
Not sure what you mean. The preview switches to and back from 3D mode the moment you click on the checkbox.

I'll check if changes done on one view update on the other and report back.

federicobond:
I can confirm the preview updates with any changes done in the other mode.

federicobond:
@keithsw1111 why was this PR closed without merging?

keithsw1111:
Because it doesn’t solve any problem as you admitted in your comment above.

Sent from my iPhone

On 20 Feb 2019, at 4:35 am, Federico Bond <notifications@github.com<mailto:notifications@github.com>> wrote:


@keithsw1111<https://github.com/keithsw1111> why was this PR closed without merging?

federicobond:
I'm sorry if I wasn't clear enough. The behavior I describe is only AFTER having applied this patch.

keithsw1111:
Except when I try it I don’t see an issue. I click the 3D and it immediately updates. Maybe attach a video of the bug you are fixing.

Sent from my iPhone

On 20 Feb 2019, at 10:48 am, Federico Bond <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm sorry if I wasn't clear enough. The behavior I describe is only AFTER having applied this patch.

AzGilrock:
I just tried it with version 2019.12 and I don't see a problem either.  I created an arch in 2D and then checked the box and the arch was showing in 3D.  What OS are you on?

federicobond:
I can consistently reproduce this on macOS.

If no module is currently selected, it will refresh when toggling from 2D to 3D, but will not refresh when toggling back to 2D. In that case you need to click on the preview to update it.

If there is a model currently selected, it won't refresh when going from 2D to 3D or backwards.

This is consistent with the current code because the preview is only refreshed inside the `UnSelectAllModels()` call, which is called in `OnCheckBox_3DClick` only if `is_3d` is true and `selectedBaseObject` is null.

I can also confirm that my fix solves the problem. With the additional `UpdatePreview()` call, the preview updates every time the checkbox is clicked.

AzGilrock:
Yeah adding the call doesn't bother me.  Most of us will never uncheck the box anyways.

federicobond:
https://github.com/smeighan/xLights/pull/1484

keithsw1111:
Can you make a video of the problem. It concerns me that I cannot reproduce the problem. It seems unlikely that this would be an OS specific problem ... not impossible ... but unlikely

federicobond:
@keithsw1111 I have uploaded a video in the description of this PR: https://github.com/smeighan/xLights/pull/1484

PR #1484, opened 2019-02-21
Update preview when toggling 3D in Layout tab

federicobond:
Video of the bug (xLights 2019.13): https://www.dropbox.com/s/q40hw7rkeok3cbw/xLights%20bug.mov?dl=0

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -5750,6 +5750,8 @@ void LayoutPanel::OnCheckBox_3DClick(wxCommandEvent& event)
     }
     obj_button->Enable(is_3d);
 
+    UpdatePreview();
+
     wxConfigBase* config = wxConfigBase::Get();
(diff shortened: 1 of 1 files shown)


keithsw1111:
Its hard to tell from the video but it looks to me like it updates ... just there is a delay because of the complexity of your layout.

federicobond:
That's not the the case here. All updates in the video are triggered by clicks, either to the checkbox or the preview. No update occurs delayed.

keithsw1111:
This code should not call UpdatePreview as it has the side effect of marking the rgbeffects file dirty. If OSX has this bug then you should be calling RenderLayout()

dkulp:
I'm not seeing this issue on OSX either...  I can keep clicking  on 3D checkbox and it keeps toggling back and forth fairly quick.

keithsw1111:
This has been here for weeks with no response to objections ... so I am closing it pending an answer to the concerns.

PR #1489, opened 2019-02-24
fixed issue 1488: Linux can get stuck in Entry() loop

### LINKED ISSUE #1488: Linux: xLights hangs on shutdown starting with 2018.52
Test Environment:
* Ubuntu 18.04.2 in Virtual Machine with 4 virtual cores (also reproduces with 2 cores). 
* Reproduced with Chris's PPA (currently 2019.09)
* Reproduced with compile from source using 2019.14 Tag

Tests attempted: 
*  xLights -r sequence.xml   
* run xLights and then do File->Quit

Result: xLights GUI completes, but xLights process doesn't terminate.  It ends up consuming all of one CPU. (Prior to the shutdown, CPU usage is very low so it isn't that a thread is misbehaving all the time.) 

I've used gdb to capture a back trace of all threads at time of hang as well as captured a system call trace.    Unfortunately, neither is leading me to the root cause.   Anybody see what I'm missing? 

[strace.txt.gz](https://github.com/smeighan/xLights/files/2897275/strace.txt.gz)

[gdb_backtrace.txt](https://github.com/smeighan/xLights/files/2897276/gdb_backtrace.txt)

ghormann:
Fixes #1488 

In Linux, it is possible for this set of events to occur.
Stop() called
Entry() started
Stop() changes _stop value

Entry() needs to used the current version of _stop during the lop check rather than a cached version.

### DIFF
--- a/xLights/xLightsTimer.cpp
+++ b/xLights/xLightsTimer.cpp
@@ -321,7 +321,7 @@ wxThread::ExitCode xlTimerThread::Entry()
     int fudgefactor = _fudgefactor;
     long long last = wxGetLocalTimeMillis().GetValue();
 
-    while (!stop)
+    while (!_stop)
     {
         if (_suspend)
         {


keithsw1111:
I am closing this without merging it because the fix is incomplete. I have included this plus a more complete removal of the unnecessary local copy of stop in the function.

Issue #1493, opened 2019-02-26
Cube Model - String Settings


priemd:
I have built an 8 x 8 x 8 cube and have not been able to accomplish the wiring layout that I used with the current string settings.

My cube was built where each layer is wired like a matrix model with zig zag each layer stacked on top of each other and all connected with the output of prior layer being the input of the next layer.

Here is the sample node layout for a 4 x 4 x 4 using the same wiring pattern:

Layer 1 (starts on back bottom left):

| 1 | 8 | 9 | 16 |
|-------|-------|-------|-------|
| __2__ | __7__ |__10__ |__15__ | 
| __3__ | __6__ |__11__ |__14__ | 
| __4__ | __5__ |__12__ |__13__ |

Layer 2 (stacks on top of layer 1, connect to output of layer 1 N16 on the back):

| 32 | 25 | 24 | 17 |
|-------|-------|-------|-------|
| __31__ | __26__ |__23__ |__18__ | 
| __30__ | __27__ |__22__ |__19__ | 
| __29__ | __28__ |__21__ |__20__ |

These layers would continue to stack on top of each other with the last node of each layer output to 1st node of the layer above it.

I was able to also create a custom 3d model which works perfectly for how my cube is built which I've attached.  I tried all combos of wiring options for 'cube' model >> export as custom model >> import and view 'model data' but none matched the layout I needed and created with custom 3d model.

Am I missing something in terms of the wiring or is this layout not supported with the current set of options?

[Cube.xmodel.zip](https://github.com/smeighan/xLights/files/2906906/Cube.xmodel.zip)

keithsw1111:
Yeah ... its not a currently supported wiring model for the cube. You managed to avoid one of the 184 patterns we do support.

AzGilrock:
Yeah the only reason I didn't drop the hammer about building something not supported is it did seem like the logical way anyone would wire up a cube if they had one long string of pixels.

priemd:
Honestly I assumed that is what Start: Back/Bottom Left, Direction: Horizontal Front/Back, zigzag meant, but instead it seems to mean vertical slices from left to right.  I guess you might call the Direction option for what I've built 'Layered Front/Back' / 'Layered Left/Right'?

[...]

PR #1496, opened 2019-03-02
add stacked wiring option to cube model

priemd:


priemd:
this commit resolves #1493.  I add a new Direction option for 'Stacked' wiring, essentially X matrixes stacked on top of each other with width x height for each layer.

priemd:
no clue why the continuous build is failing, i changed a single cpp file which runs perfectly fine locally on my branch, any help from you gurus would be appreciated?

PR #1515, opened 2019-03-17
Fix linux build for libcurl

ghormann:
Fixes Linux by linking libcurl.   

Also adds bin/xSMSDaemon to .gitignore.


ghormann:
In addition to this patch, We'll have to ask @cjd to add libcurl to the master version of https://www.adebenham.com/wp-content/uploads/xlights/build/Recipe.deps and update the xLights Docker image.   


Greg

PR #1557, opened 2019-05-15
Add a "Fold" transition

KMojek:
This isn't ready to merge but I want to get Keith's opinion on it. Clyde Lindsey contacted me a couple of weeks ago with this idea. At first, I thought it would be a trivial addition but this one is a little different since the effect would need a copy of the RenderBuffer below it (technically two layers since the warp's initial RenderBuffer is just a copy of the layer below). This is just a proof-of-concept at this point with no settings hooked up (I think that assigning a value curve to the fold point might be interesting).

Here's a quick video of the warp over an image over a butterfly effect: https://vimeo.com/336004349

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,7 @@ Issue Tracker is found here: www.github.com/smeighan/xLights/issues
 
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
+   -- enh (kevin)   Add "Fold" transition type
 2019.26 May 24, 2019
    -- enh (scott)   When prompting for missing audio include the original audio filename to make it easier to find
    -- enh (scott)   When audio is missing look down 1 folder and grab it from there if found
--- a/xLights/PixelBuffer.cpp
+++ b/xLights/PixelBuffer.cpp
@@ -194,7 +194,10 @@ namespace
          if ( rb1 == nullptr )
             return tex2D( cb0, uv0.x, uv0.y );
 
-         return progress < 0.5 ? tex2D( cb0, uv0.x, uv0.y ) : tex2D( *rb1, 1-uv0.x, uv0.y );
+         if ( progress < 0.5 )
+            return tex2D( cb0, uv0.x, uv0.y );
+         xlColor c = tex2D( *rb1, 1-uv0.x, uv0.y );
+         return ( c != xlCLEAR ) ? c : tex2D( cb0, uv0.x, uv0.y );
       }
       return xlBLACK;
    }
--- a/xLights/PixelBuffer.cpp
+++ b/xLights/PixelBuffer.cpp
@@ -168,6 +168,42 @@ namespace
 
    const float PI = 3.14159265359f;
 
+   xlColor foldIn( const ColorBuffer& cb0, const RenderBuffer* rb1, double s, double t, float progress )
+   {
+      const double CAMERA_DIST = 2.;
+      Vec3D ray( s*2-1, t*2-1, CAMERA_DIST );
+      float phi = progress * PI;   // rotation angle
+
+      // rotated basis vectors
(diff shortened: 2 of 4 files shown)


keithsw1111:
First impressions ... from reading the diff only it seems to make lots of assumptions and breaks the existing canvas concept. it is more like a transition or active blend style. would it make more sense there?

KMojek:
Yeah, I agree it makes more sense as a transition. I was under the impression that transitions were only capable of providing masks and/or fades but it looks like that may not be accurate; that's just how it works with the existing transitions? I'll close this out and look at implementing as a transition if that makes sense.

keithsw1111:
I am not saying transitions are a silver bullet. There will be challenges in making it work there. I just know that:
1. When I first built the transitions I tried building them as an effect and Dan set me straight at that time.
2. To the user they make more sense as a transition ... it is a more logical place to find them.
3. It does not further clutter the handling of layers and buffers in the layer rendering code.
It is also unfortunately likely to place some limits on the functionality but i think that is something we are going to have to live with.

KMojek:
Okay, still a work-in-progress but somewhat-working as a transition. I'm not familiar enough with rendering to know why there seems to be an extra blank layer in there, or how to identify it as such, assuming that's what it is. https://vimeo.com/337202094

keithsw1111:
Sorry Kevin ... really caught up on another problem and have had almost no time to look at this. I dont really understand your question re extra blank layer.

KMojek:
No problem... this is low-priority stuff. To clarify a bit, I poked at it a bit more tonight and seems related to this: https://github.com/smeighan/xLights/blob/8e6919dcb2f94220aa2bc4c61be9af45b87ab2e4/xLights/Render.cpp#L47 I think I was assuming the numLayers in PixelBufferClass and the size of the passed-in validLayers param would be in sync... likely just a misunderstanding on my part of how the layer blending works.

keithsw1111:
So are we ready to merge?

KMojek:
I've got a bit more work to do on the "in" version of the transition. I'll try to wrap that up within the next 24 hours. I'll just merge it once that's done unless you spot any obvious problems with what's in here currently.

keithsw1111:
Ok.. I will skip the next release and look to merge for the next one.

keithsw1111:
There s something wrong with this. I put it on a matrix as an in transform and even after the transition the pixel buffer is weirdly distorted
image

keithsw1111:
Here are the effect settings:

CopyFormat1	0	1	1	0	-1	NO_PASTE_BY_CELL
Bars	E_CHECKBOX_Bars_3D=0,E_CHECKBOX_Bars_Gradient=0,E_CHECKBOX_Bars_Highlight=0,E_CHOICE_Bars_Direction=up,E_SLIDER_Bars_BarCount=1,E_TEXTCTRL_Bars_Cycles=1.0,T_CHOICE_In_Transition_Type=Fold,T_SLIDER_In_Transition_Adjust=50,T_TEXTCTRL_Fadein=0.5	C_BUTTON_Palette1=#FFFFFF,C_BUTTON_Palette2=#FF0000,C_BUTTON_Palette3=#00FF00,C_BUTTON_Palette4=#0000FF,C_BUTTON_Palette5=#FFFF00,C_BUTTON_Palette6=#000000,C_BUTTON_Palette7=#00FFFF,C_BUTTON_Palette8=#FF00FF,C_CHECKBOX_Palette1=1,C_CHECKBOX_Palette2=1,C_CHECKBOX_Palette3=1	10250	13700	1	-1000	NO_PASTE_BY_CELL

keithsw1111:
Model is a default matrix

KMojek:
I should've caught that... a fix is pushed up and merged.

PR #1566, opened 2019-05-24
Add effects-grid representation of transitions

KMojek:
This is a first stab at adding an effects-grid representation of transitions.

The intent here is not to provide any effects-grid control of the transition lengths but just to provide some indication when transitions exist on effects. My first idea here was to alpha-blend the transition representation over the effect background but that can cause goofy results over some effects... red out transition over yellow on-effect in image below shows as orange.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -1,4 +1,4 @@
-xLights is a show sequencer and player/scheduler designed to control
+﻿xLights is a show sequencer and player/scheduler designed to control
 USB/DMX/sACN(e1.31)/ArtNET(e.1.17)/DDP controllers.
 xLights also integrates with the Falcon Player.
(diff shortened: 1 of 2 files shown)


KMojek:
transitions_on_effects_grid

KMojek:
transitions_on_effects_grid_2

Updated to make transitions less intrusive.

keithsw1111:
What happens when transition in + transition out > effect length ... does it go yellow in the middle.
What about when underlay is the same colour as the transition colour ... maybe a black line underneath would ensure it remains visible

KMojek:
Added the "black line underneath" and yellow for overlapping in/out transitions.

PR #1636, opened 2019-08-05
Add Logarithmic X axis option to Music Effect

PeteMatthews:
Replicated from the VU Meter to limit the heavy skew to the left hand
side.


keithsw1111:
I dont see a wxs file in the change set suggesting to me you changed generated code manually. This will result in all your changes being lost the first time someone touches the panel in wxSmith.

PeteMatthews:
You're absolutely right - Hadn't spotted that one. Quite glad it exists though... I was getting unpleasant flashbacks to the early days of Java+Swing doing it all by hand :)

Reworked with the UI designed as required.

PR #1807, opened 2019-11-03
removing restriction in unicast and unicast csv that wouldn't sync fo…

gsrunion:
…r media


keithsw1111:
So has this now been tested?

gsrunion:
It tested with Unicast. Not sure what UnicastCSV is so I didn’t cover that path.

PR #1915, opened 2020-01-15
Fix further memory leaks

PeteMatthews:
Seeing as the last one caused a few problems, I'll try
to be more explicit about the changes here.

Tracelog - The per-thread trace buffers aren't tidied up,
and some still have log messages present. These were leaking, so
they're now freed up in the destructor.

ViewObjectManager - The view objects were never removed. Invoke the
existing "clear" function from the destructor.

XLightsMain UI Elements - We're leaking button elements for some
reason - Added delete calls, matching the pattern already present for
CheckBoxLightOutput.

XLightsMain XmlSetting - Need to delete elements that we've obtained
through "RemoveChild", as per the wxWidgets docs.

Testing performed -
 - Opened layout
 - checked layout in 2d/3d
 - opened sequence
   - rendered
   - played
   - changed
   - played
   - saved
- closed xLights
- Repeated the above to verify changes saved

When I have saved a sequence, I do see an access violation on
shutdown, but that doesn't seem to be new. I see it on the current
master branch, and in local builds of released versions back to 2020.1.
Not sure if this occurs in the MinGW builds too, or if it's a VS thing,
but I'll have a poke around. Entirely understand if you don't want to
take this PR without that resolved, although I don't think it's
related to my changes.

### DIFF
--- a/xLights/TraceLog.cpp
+++ b/xLights/TraceLog.cpp
@@ -12,13 +12,17 @@ thread_local std::list<std::string> *threadLog = nullptr;
 
 class TraceLogHolder {
 public:
-    TraceLogHolder() {TRACE_LOG_VALID = true;}
-    ~TraceLogHolder() {TRACE_LOG_VALID = false;}
-    
-    
     std::mutex MESSAGE_MAP_LOCK;
     std::map<std::thread::id, std::list<std::string>*> LOG_MESSAGES;
-    
+
+    TraceLogHolder() {TRACE_LOG_VALID = true;}
+    ~TraceLogHolder() {
+        TRACE_LOG_VALID = false;
+        for (auto it_thread = LOG_MESSAGES.begin(); it_thread != LOG_MESSAGES.end(); ++it_thread) {
+            delete it_thread->second;
+        }
+        LOG_MESSAGES.clear();
+    }
     
     void ClearTraceMessages() {
         if (!TRACE_LOG_VALID) {
--- a/xLights/ViewsModelsPanel.cpp
+++ b/xLights/ViewsModelsPanel.cpp
@@ -821,7 +821,18 @@ void ViewsModelsPanel::OnButton_RemoveAllClick(wxCommandEvent& event)
 void ViewsModelsPanel::Clear()
 {
     ListCtrlModels->ClearAll();
+
+    for (int i = 0; i < ListCtrlNonModels->GetItemCount(); ++i)
+    {
+        Element* e = (Element*)ListCtrlNonModels->GetItemData(i);
+        if (e != nullptr && e->GetType() == ELEMENT_TYPE_MODEL && e->GetSequenceElements() == nullptr)
+        {
+            delete e;
+            ListCtrlNonModels->SetItemPtrData(i, (wxUIntPtr)nullptr);
+        }
+    }
     ListCtrlNonModels->ClearAll();
+
     ListCtrlViews->ClearAll();
     ValidateWindow();
 }
--- a/xLights/models/ViewObjectManager.cpp
+++ b/xLights/models/ViewObjectManager.cpp
@@ -15,6 +15,7 @@ ViewObjectManager::ViewObjectManager(xLightsFrame* xl) : xlights(xl)
 ViewObjectManager::~ViewObjectManager()
 {
     //dtor
+    clear();
 }
 
 BaseObject* ViewObjectManager::GetObject(const std::string &name) const
--- a/xLights/xLightsApp.cpp
+++ b/xLights/xLightsApp.cpp
@@ -184,6 +184,7 @@ void InitialiseLogging(bool fromMain)
 
 					logger_base.info("    %s : %s", (const char *)(*it)->getName().c_str(), (const char *)levels.c_str());
 				}
+                delete categories;
 			}
             catch (log4cpp::ConfigureFailure& e) {
                 // ignore config failure ... but logging wont work
--- a/xLights/xLightsMain.cpp
+++ b/xLights/xLightsMain.cpp
@@ -2406,6 +2406,23 @@ xLightsFrame::~xLightsFrame()
     }
 
     delete CheckBoxLightOutput;
+    delete ButtonPasteByTime;
+    delete ButtonPasteByCell;
+
+    delete Button_ACDisabled;
+    delete Button_ACSelect;
+    delete Button_ACOn;
+    delete Button_ACOff;
+    delete Button_ACTwinkle;
+    delete Button_ACShimmer;
+    delete Button_ACIntensity;
+    delete Button_ACRampUp;
+    delete Button_ACRampDown;
+    delete Button_ACRampUpDown;
+    delete Button_ACFill;
+    delete Button_ACCascade;
+    delete Button_ACForeground;
+    delete Button_ACBackground;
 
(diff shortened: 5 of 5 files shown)


keithsw1111:
Yeah ... not going to merge this until you tell us more about this crash. I dont believe it is an existing problem unless it is rare and just happens to be present in your setup. In which case I want to see your setup to try to fix it.

PeteMatthews:
Fair enough - Am looking at it now, in fact.

It seems to be happening if I open xLights (with my default show folder), create a new sequence, drag an effect to any model, then close xLights, saving the new sequence on the way. It manifests as an access violation in the CRT process shutdown. 

It doesn't happen if I open and close an existing sequence.

I'm trying to isolate exactly what's going on at the moment - Will update when I've got more information.

In case it does matter - are the released versions built with VS or MinGW? I'm using VS, and can't get the same behaviour with released binaries, so was wondering if there was a difference there.

Thanks for your time.

PeteMatthews:
Continuing to look into the crash, but have opened issue #1917 in case anyone else has any bright ideas.

AzGilrock:
Yeah these memory leak fixes always worry me.  I've had a lot of times I try to add cleanup code in destructors and I end up with crashing on exit and can't figure out why.  And right now I just ran into a crash if I try to create a downloaded model and immediately close the vendor dialog its crashing on cleanup.

PeteMatthews:
Yeah, I know it's a little unusual, but I find them weirdly therapeutic... There's an automated tool telling you something's wrong, and you need to go fix it. The answers are often not obvious (especially heap corruption), so it's a bit of a challenge. It's equal parts confusing, infuriating and satisfying when you finally get there. 

It probably helps that "real job" is in consumer-facing low resource embedded stuff which needs uptimes measured in months or people complain... Any kind of resource leak immediately makes me twitchy :)

AzGilrock:
Oh I see Visual Studio telling me about the leaks every time I exit.  I set out to clear them all once and got through some but not a lot of them.  The bright side is in 6 years I've never seen memory seem to run out even without rebooting so I've just figured Windows probably recovers memory when an applications quits anyways kinda like the C# garbage collector where they never worry about deleting stuff.  By the way my "real job' is low level embedded code but we follow a set of "rules" that prevents problems.  Its a little easier because typically embedded software never needs to restart between power cycles.  So for one we never allow dynamic memory allocation all memory must be allocated during construction.  So I've had to create my own memory manager classes to get around that.

PeteMatthews:
Getting off topic here, but - ooooh - you're in proper embedded :) 

I'm in satellite broadcast kit at the moment, which is all embedded linux so we're not at custom allocator level, but I used to work in industrial process control kit with vxWorks and all the static allocation, custom mem maps and the works. It was deeply unpleasant :)

(And yeah, Windows cleans up on a process level when you exit, so it's not a system level thing. You can certainly make a case for it not being all that important, especially when you've got 16/32Gb of RAM to play with :) )

dkulp:
Periodically, I do full memory profiles on xLights while doing full sequence rendering and playback.   However, I specifically create a "snapshot" once xLights is started and then another right before exiting and compare those so it's only comparing the allocations that occur during usage and not "startup leaks".   They aren't really leaks.

PeteMatthews:
> They aren't really leaks.

You're not wrong, they just pollute the otherwise-potentially-useful CRT process shutdown report. If the thing's spewing errors left, right and centre at shutdown anyway, no one notices when they accidentally add a leak or two. If there's normally no report and it suddenly shouts at you, there's a better chance you'll go fix the code you've just changed to clean up properly, and that's wildly preferable to trying to hunt them down after the fact years later :)

AzGilrock:
Hey any you fix I'm all for it.  I was surprised the ViewObjectManager clean() call wasn't already there because I wrote all that.  I could swear I had tried calling it during destruction and had crashing.  One thing you gotta make sure you test is switching Show folders.   As a developer I tend to switch Show folders alot and I'll find crashes that normal users don't see often.

AzGilrock:
I got one right now frustrating the crap outta me.  It just started yesterday when I wanted to add DMX models to the vendor downloader. Well I'm back to main branch code with none of my changes and it still crashes every time I close the vendor download dialog.
image

AzGilrock:
And this is why this has me on edge....look at the comment in the wxWidgets code that is causing my crash
image

AzGilrock:
Well I got it to hit a breakpoint on that line and it steps over it just fine.  It seems to crash after that.  Here's the messages I get if anyone has any ideas.  This is really making it hard to troubleshoot adding Vendor Download content.
image

PeteMatthews:
Just tried a trivial switch of show folders with these changes applied, and all seems happy - Will exercise it a little more later - Thanks for the tip.

Your thing looks more like a heap corruption than anything. Unable to reproduce it here (the words every developer wants to hear), but will poke at it more for fun later. I can't guarantee I'll be any help, but I do enjoy a good heap corruption, so if you want to send me show directory stuff, feel free - you can find me on Facebook or whatever for contact details.

AzGilrock:
Yes its definitely heap corruption.  Its not the show folder because it happens when starting with a blank show folder.  All I do is click the down arrow in layout to create a model, the vendor downloader dialog opens, and I immediately close it and it crashes every time but only in debug.  No crash in release build.

AzGilrock:
Problem is I can't troubleshoot the importing of a downloaded DMX model.

AzGilrock:
Well I've narrowed it down to here.  If I comment out those 2 lines no crash.  Interesting that is has no issue deleting the entries that are nested down an extra level because I did set breakpoints and made sure it hits up there...it does for the 46" Chromabulbs.
image

AzGilrock:
So I can put a copy of Boscoyo.xml on my own website and reduce it to one model and it doesn't crash even though its hitting those same statements.  So now it seems dependent on what's in the XML.  I wonder if Keith purged his download cache before he tested what I said was crashing.

AzGilrock:
Well at least rerouting to a slimmed down Boscoyo file is allowing me to work on my own XML since the crash is gone.  Getting back to your Pull Request here I was deferring to Keith for seeing if he's ready to pull it in.

keithsw1111:
DrMemory is amazing for heap corruption errors.

PeteMatthews:
I've heard Dr Memory is good... Tried it out yesterday when trying to diagnose that shutdown crash, and it immediately crashed on me :D

Might try it again if it comes recommended :)

(Thanks for the merge... Will try and do something more useful to the project soon :) )

keithsw1111:
If you plan on changing any significant behaviour etc I suggest you raise a github first and let us discuss it. That changes with time but your first few changes it saves a lot of grief on merge when folks violently disagree with the change.

PeteMatthews:
Absolutely 100% understood. Plan was to come ask you as before and try not to get so interrupted I disappear this time. Fully understand that there's a wider project here with goals and targets that don't need unsolicited major change dreamt up by inexperienced "helpers" :)

Fundamentally. I'm just grateful this project (and others in the space, notably FPP) exists, and keen to help out if I can without getting in the way :)

AzGilrock:
Heck I was trying Dr Memory and Application Validator.  Dr Memory I'm not sure how you're supposed to use it when you want to run VS debug sessions.  Application Validator didn't see any problems at all when the app was crashing.

PR #1916, opened 2020-01-16
Vixen3import

robness:
Fixed problem of not finding models to map in Vixen 3 Import.


computergeek1507:
You need to rebase you pull request

robness:
I did rebase the local branch from upstream/master before making the commit so I'm confused why this happened. Althought this is my first time working with forks and github instead of Enterprise Bitbucket repos and their pull requests so any advice is greatly appreciated.

PR #1925, opened 2020-01-23
Retain locked words/phonemes when breaking down lyrics

PeteMatthews:
As requested at http://nutcracker123.com/forum/index.php?topic=6931.0

Previous behaviour -
Effects layers below the current selection were deleted (words and phonemes for breakdown phrases, just phonemes for breakdown words).
New layer(s) were created and populated as appropriate.

New behaviour -
Effects layers are purged, rather than deleted. All non-locked effect elements are deleted.
Layer(s) are repopulated, although new words/phonemes that would clash with existing (locked) ones are not created.

There is an open question over the repopulation semantics here - At the moment, we just don't insert a given word/phoneme if there would be a clash. It might be possible to try to work out which words/phonemes already exist and insert others around them. This is probably difficult to work out, and not that useful... A bunch of strategies could well end up with broken down words in the wrong order, etc. I'm more than happy to discuss and implement something if anyone has strong preferences. Feel free to discuss on this PR, or open an issue assigned to me.

### DIFF
--- a/xLights/sequencer/RowHeading.cpp
+++ b/xLights/sequencer/RowHeading.cpp
@@ -1140,7 +1140,28 @@ void RowHeading::BreakdownTimingPhrases(TimingElement* element)
     EffectLayer* layer = element->GetEffectLayer(0);
     if( element->GetEffectLayerCount() > 1 )
     {
+        bool found_locked = false;
+
         for( int k = element->GetEffectLayerCount()-1; k > 0; k--)
+        {
+            EffectLayer* purge_layer = element->GetEffectLayer(k);
+            for (int ef = 0; ef < purge_layer->GetEffectCount(); ef++)
+            {
+                Effect* e = purge_layer->GetEffect(ef);
+                if (!e->IsLocked())
+                {
+                    found_locked = true;
+                    break;
+                }
+            }
+        }
+
+        if (found_locked)
+        {
+            return;
+        }
+
+        for (int k = element->GetEffectLayerCount() - 1; k > 0; k--)
         {
             element->RemoveEffectLayer(k);
         }
@@ -1158,12 +1179,32 @@ void RowHeading::BreakdownTimingPhrases(TimingElement* element)
 
 void RowHeading::BreakdownTimingWords(TimingElement* element)
 {
+    EffectLayer* word_layer = element->GetEffectLayer(1);
+    EffectLayer* phoneme_layer = nullptr;
+
     if( element->GetEffectLayerCount() > 2 )
     {
+        bool found_locked = false;
+        phoneme_layer = element->GetEffectLayer(2);
+        for (int ef = 0; ef < phoneme_layer->GetEffectCount(); ef++) {
+            Effect* e = phoneme_layer->GetEffect(ef);
+            if (!e->IsLocked())
+            {
+                found_locked = true;
+                break;
+            }
+        }
+
+        if (found_locked)
+        {
+            return;
+        }
+
         element->RemoveEffectLayer(2);
     }
-    EffectLayer* word_layer = element->GetEffectLayer(1);
-    EffectLayer* phoneme_layer = element->AddEffectLayer();
+
+    phoneme_layer = element->AddEffectLayer();
+
     for( int i = 0; i < word_layer->GetEffectCount(); i++ )
     {
         Effect* effect = word_layer->GetEffect(i);
--- a/xLights/sequencer/SequenceElements.cpp
+++ b/xLights/sequencer/SequenceElements.cpp
@@ -2064,7 +2064,18 @@ void SequenceElements::BreakdownPhrase(EffectLayer* word_layer, int start_time, 
             {
                 word_end_time = end_time;
             }
-            word_layer->AddEffect(0,words[i].ToStdString(),"","",word_start_time,word_end_time,EFFECT_NOT_SELECTED,false);
+
+            // Don't clash with any existing words - it makes a mess
+            if (!(word_layer->GetAllEffectsByTime(word_start_time, word_end_time).size()))
+            {
+                word_layer->AddEffect(0, words[i].ToStdString(), "", "", word_start_time, word_end_time, EFFECT_NOT_SELECTED, false);
+            }
+            else
+            {
+                // If there's only a clash for part of the word, can we fit this one in anyway?
+                // Semantics of this are up for grabs... TBD.
+            }
+
             word_start_time = word_end_time;
         }
     }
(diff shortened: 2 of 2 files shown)


AzGilrock:
I need to wait till later to fully digest this.  I wrote most of that original logic.  One thing to note is there are two different ways to cause a phrase or word to breakdown.  You could be at the row level telling xLights to breakdown the entire row of phrases or words or you could be click specifically on a single phrase or word.  If you were selecting a single item it was not deleting any layers that only happened when the entire row was being broke down.  My opinion was people should not be selecting those breakdowns unless were want to replace the items so I'd need to think about this whole idea of allowing locked items.  I would almost tend to tell them they cannot breakdown the row if there are any locked items at all on the row below it.  Same with an individual phrase or word.  If anything is locked then they get a message saying we cannot breakdown due to locked items.   My original thought was the user didn't need to select those actions if they didn't want the whole row replaced and I was presenting them with a warning.  Problem is most users couldn't figure out that they could select a single phrase or word to be broken down on its own.

AzGilrock:
Yeah the more I think about it I kinda like the strategy of just preventing the entire operation if anything locked would block it.  So if a row breakdown is selected any locked item on the row(s) below it would trigger a warning message saying it could not be done.  If a user has locked items under a particular word or phrase then further breakdowns must be down by selecting a single word or phrase which would follow the same rules of a warning message if anything below them is locked.

PeteMatthews:
Sorry it's taken me a few days to get back to this - While I messed with some lyrics last year, I signed up for the Facebook singing faces effort a few weeks back... I thought I'd go do my bit for that and get myself some deeper real world experience, which has been no end of fun :)

First up - If we're interpreting the requested feature that way, no problem. Clearly, that's actually less effort than the PR here, and fundamentally comes down to "If there's anything in the way that's locked, disallow (or at least re-confirm) the whole operation". I'll happily go back and rework the changes if that's where we land. 

I'll be honest though - it feels like this could be more useful to users. My experience over the last few days has been something like:

* Start new sequence
* Import lyrics copy/pasted from Google, etc.
* Space Phrases correctly-ish
* Breakdown phrases
* Respace words within phrases
* Inevitably, get to a phrase that has an issue - Spelling error, incorrect lyric, badly positioned - that I didn't notice before
* Correct that phrase, and possibly a number after it (Some sources are especially bad for incorrect chorus elements, etc that repeat throughout) or around it (when I've screwed up the spacing and need to re-jig a few around it)

At this stage, re-running "Breakdown phrases" on a row-level is clearly not an option, so I've got to go through, delete the broken-down words for the relevant phrases (or we'll currently get a bunch of weird overlapping words, which this PR also prevents), then break down the individual selected phrases. Selecting large numbers of words for deletion also seems awkward on my install, although I'm totally prepared to believe that's cos of crappy trackpad/crappy trackpad drivers.

In this situation, it would be handy to lock the work I've done as I go (also saving me from user-error where I accidentally move things I didn't mean to) and then have the ability to re-run the breakdown on a row-level. I'll admit it's a little "sledgehammer to crack a nut", but it's a bunch easier than fiddling around deleting stuff out of the way. 

(And the same applies to Word -> Phoneme breakdowns. Hopefully you'll have caught most of the errors by then, because selecting/deleting a big pile of phonemes is even less fun :) )

I know I'm "outsider poking around in source with huge legacy" here, so please don't think I'm trying to start some kind of argument... I've all the respect in the world for this project and it's contributors, and if you (and/or others) would rather this went in another direction (or just don't want to think about it any more), I'll happily revise the PR to what you're expecting. Just thought it was worth outlining where I was coming from. 

Thanks for your help and feedback - it really is appreciated.

AzGilrock:
You should really never need to run Breakdown Phrases more than once.  The first thing you should do is be happy with where your Phrases start and end before you breakdown phrases.  After that doing it from a single phrase at a time should be sufficient.  You say you're having to delete stuff below it or it becomes a mess.  Well maybe that's the issue is it not deleting everything below the phrase first?  If not then I'd say make that the change.  Whenever you do a breakdown it should be assumed everything underneath it is deleted.  And you should really never be dragging breakdown items outside the boundaries of the item above it that created it.  I think people just develop bad habits on how they like to use the feature.  Sean and I used to disagree on how to create a lyric track and then he watched me teach a class and admitted he liked my method better.  I think its just too much of a pain trying to remember what stuff is locked and not locked and making sure you lock all the right things.

AzGilrock:
I'd need to go home and see how its currently working.  Sometimes things change without me realizing someone change it.  My preference is that any breakdown operation clears the space below it whether it be at the row level or a single phrase or word.  If you want to make it so locking prevents that deletion that's fine because if I never do locking I'd never even notice the feature.  The question is what to do when there is partial locking below a phrase or word.

AzGilrock:
What I run into is sometimes users make complaints and other developers will make a change that I end up hating because it was really a difference of opinion on how something should work.  For instance I was just doing a copy/paste of a model and noticed the copy of my model is being placed to the right and below the ground level of the grid in 3D.  I originally had it do a paste in place which allowed me to drag the copy directly to the right for perfect alignment.  So I'm pretty sure someone complained they couldn't see their copy after they pasted it because it was on top of the original so someone decided to add an offset.  Well I hate the offset because I usually want my copies to be at the ground level and half the time they end up below my ground image where I can't see them.  I would have preferred the copy offset to be in the X direction instead of Y but I guess it was a person  using 2D.

PeteMatthews:
Understood, and I think you're right. There's definitely a certain amount of error (or inexperience) between chair and keyboard on my side.

If I've got this, I think we're heading for...
* Breakdown of anything should always delete the stuff in it's way first.
* If any breakdown finds a locked element in it's way, deny the whole thing.

(And yeah, paste in place would definitely be my preference, because you're almost always doing it to generate a bunch of things you want aligned on at least one axis, but there's no easy to tell which)

I'll go re-jig the commit and resubmit shortly.

AzGilrock:
Yep that sounds good.  BTW if you work without any timing tracks selected its pretty easy to drag the yellow dashed box around those words or phonemes and hit delete.

computergeek1507:
How do you plan on representing when something is locked? I can image users getting frustrated when some words/phenemes are locked and they do not know it. They will try to breakdown a word or phrase and "think" xLights is not working correctly if it does nothing. I see this all the time when tried to edit fixed timing tracks and cannot figure out why it doesn't work.

PeteMatthews:
The facility to lock words/phonemes is already there, it just doesn't do very much. Visually, it already does the same as locked effect elements - It makes the arms of the span red on selection rather than purple. We could colour locked text elements differently? Make the label background red rather than yellow? That said, given that we're building on implemented functionality here, it might be best to have that as a separate PR?

As far a user experience goes, the intention is to throw a dialog if the user attempts something blocked by a locked element. We could make the dialog pretty explicit - "The following words are locked: Christmas, time (and 23 others)" or something?

@AzGilrock - Thanks - will try it. Every day, school day, etc. ;)

AzGilrock:
Yep phrases, words, and phonemes internally are treated the same as other effects so I would expect the lock coloring of effects to work the same on them.  The same argument would apply to deleting effects from anywhere on the grid.

PeteMatthews:
(Apologies, don't merge this yet - commit was merely a sync between 2 systems, and I forgot it would update the PR too :) )

PeteMatthews:
Aborting this one for now, thanks to git-related mess. Will open a new one shortly.

PR #1934, opened 2020-02-06
Retain locked words/phonemes when breaking down lyrics.

PeteMatthews:
As requested at http://nutcracker123.com/forum/index.php?topic=6931.0

Previous full row behaviour -
Effects layers below the current selection were deleted (words and phonemes for breakdown phrases, just phonemes for breakdown words).
New layer(s) were created and populated as appropriate.

New full row behaviour -
Effects layers that would be deleted are checked for locked effects.
If any are found, abort the operation and inform the user.
Otherwise, delete them all and repopulate.

Previous individual phrase/word behaviour -
All effects in the target time range were deleted and repopulated

New individual phrase/word behaviour -
Effects in the target time range are checked.
If any are locked, abort the operation and inform the user.
Otherwise, delete them all and repopulate.

See previous discussion in #1925


AzGilrock:
Forgot to look at this last night.  I'll try to remember tonight.

Issue #1933, opened 2020-02-03
2020.4 Linux move to GTK3 tracking bug


cjd:
This bug is to track current move from WxWidgets 3.1.2 and GTK2 to WxWidgets 3.1.1 and GTK3 for the appimage build.
Current appimage is at https://www.adebenham.com/wp-content/uploads/xlights/xLights-2020.4-x86_64.AppImage
This is a flow on from https://github.com/smeighan/xLights/issues/1922

extremeskillz:
Just tried 2020.4 on Ubuntu 18.04lts and I'm getting a crash when add or deleting models. Everything else seems to be fine. Attached is my logs.
[xLights_dbgrpt-11787-20200203T195916.zip](https://github.com/smeighan/xLights/files/4150904/xLights_dbgrpt-11787-20200203T195916.zip)

matt-torrie:
to add onto @extremeskillz there is also a crash that happens when you attempt to switch to any layout

leighfiddes:
Another issue: Select a model on the House Preview with the mouse and then right-click on the model. The model will be rotated about 58 degrees anti-clockwise and scaled up by about 2x. This doesn't happen with the "normal" appimages.

leighfiddes:
My mistake: This issue occurs on all 2020.x appimages. I will raise a separate issue for this.

reenx:
I have been researching the crash when adding a new model and discovered that it goes away when using only one cpu. (From the README.linux: taskset -c 0 xLights). 

I am guessing that in between the creation of the model and the finalizing of the model an update is triggered which will trigger the assert in LayoutPanel::OnSelectionChanged because the model is invalid. 
I have no good solution yet. But at least xLights is workable for me again by using only one core.

cjd:
Closing off old linux issues - reopen if still occurring

PR #1978, opened 2020-03-22
Ignore selection of invalid model when editing layout

leighfiddes:
When adding or deleting items from the Layout, OnSelectionChanged() will
be called pointing to invalid models which was causing an assert.
Change to log and ignore.
This appears to be a similar problem to that described in wxWidgets ticket #18045.
This addresses one item in Issue #1933.
@cjd please consider.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -6812,6 +6812,11 @@ void LayoutPanel::OnSelectionChanged(wxTreeListEvent& event)
             ModelTreeData *data = (ModelTreeData*)TreeListViewModels->GetItemData(item);
             Model *model = ((data != nullptr) ? data->GetModel() : nullptr);
             if (model != nullptr) {
+                if (!xlights->AllModels.IsModelValid(model))
+                {
+                    logger_base.debug("Warning: Model at %p is Not Valid. This would have crashed. Ignoring.\n", model);
+                    return;
+                }
                 wxASSERT(xlights->AllModels.IsModelValid(model));
                 if (model->GetDisplayAs() == "ModelGroup") {
                     mSelectedGroup = item;


keithsw1111:
I would be more interested in a fix which identified why we were holding onto an invalid model pointer. Also this would not be a debug message but a error or critical message.
If you can reliably reproduce it maybe you could post a video of that.

leighfiddes:
Thanks for your response Keith. I have captured videos and log of the issue. I'm using 2020.11 on Kubuntu 18.04 LTS.
1) Adding a model group causes immediate crash. See video "add.mp4"
2) Deleting a model causes immediate crash. See video "delete.mp4"
I then built a debug build of 2020.11
3) Deleting a model causes an Assert message. See video "delete_debug.mp4"
I added my proposed patch. I have also added extra debugging (prefixed by "LF")
4) Deleting a model seems to work fine. See video "delete_patched.mp4". I have also extracted the relevant lines from "/tmp/xLights_l4cpp.log" in the file "log_extract.txt". You can see after the DELETE_MODEL_GROUP that the old models are still being referenced, for example "Arches" was at 0x557cd079e0f0 before the deletion but this pointer now does not point to a valid model. "Arches" has moved to 0x557cd0a67ee0.
[videos.zip](https://github.com/smeighan/xLights/files/4379866/videos.zip)

keithsw1111:
Do you have your show folder ... I wonder if this isnt related to the leading trailing space changes I have been trying to make.

leighfiddes:
This is the simple show I used for the "delete_debug" and "delete_patched" videos.
[testshow.zip](https://github.com/smeighan/xLights/files/4386555/testshow.zip)

keithsw1111:
So I added groups, deleted models ... no crashes

leighfiddes:
I thought you only ran Windows. Do you run Linux as well?

keithsw1111:
I run windows ... if the bug is linux specific then it is likely event related. I still need to find the cause ... not keen to just treat symptoms

leighfiddes:
I appreciate you efforts Keith. To me it appears to be a mismatch with the wxGTK interface library which behaves differently to the WxWindows library because obviously the Windows version doesn't crash. As I initially noted, this appears similar but not identical to the issue in [http://trac.wxwidgets.org/ticket/18045](url).

keithsw1111:
I have included this but in a way where it only runs on Linux as I am led to believe this is purely a linux issue,

PR #1989, opened 2020-03-30
Initial updates to FFmpeg

KMojek:
* Updating to most recent FFmpeg release
* Update AudioManager reading / decoding code to current FFmeg APIs
* Update AudioManager writing code to current FFmpeg APIs

Leaving the FFmpeg code in HinksPixExportDialog untouched for now; I plan on updating that and anywhere else where we get the warnings regarding deprecated APIs in a later branch.

Testing completed:
* Able to create new musical sequences using MP3 and WAV files that I had used previously in xLights
* Made a temporary modification to 'Prepare Audio' in order to test the updates to AudioManager::CreateAudioFile(). I then disabled that modification once I verified that exported audio was working as expected.

I also found what seems like a bug... attempting to create a musical sequence based on a short (5 seconds) low-bitrate MP3 results in silent audio. The same thing happens in `2020.12` though, so I don't think it's related to this branch. I didn't log it as an issue because it seems unlikely that users would be working with such a tiny bit of audio.

Not sure if my CodeBlocks settings are still not quite right; changes here are more evident by blocking out whitespace changes: https://github.com/smeighan/xLights/pull/1989/files?w=1


KMojek:
@keithsw1111 I would normally just merge this myself but since it's a rather large one, you might want to take a look. Thanks!

KMojek:
Well, I'm not entirely sure that this won't break things on Mac and/or Linux but this has been open for a couple of days now so I'm going to merge it.

PR #1994, opened 2020-04-02
Move FFmpeg dependencies out of HinksPixExportDialog

KMojek:
I added an `AudioLoader` class to hide the FFmpeg details of reading, decoding, and resampling audio. Eventually this will remove a bunch of duplicated FFmpeg code.


KMojek:
The output format is currently hard-coded in `AudioLoader::loadAudioData()` since that's all we've needed so far but that could be exposed when/if we need it.

Issue #1964, opened 2020-03-08
Right-click in House Preview corrupts Selected Model


leighfiddes:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
On the Layout page, right-clicking the mouse in the House Preview will corrupt the selected model is any.

**To Reproduce**
1. Go to layout tab.
2. Select a model on the house preview window.
3. Right-click anywhere in the house preview to bring up the menu
4. Observer that the selected model has been expanded and rotated.

**Expected behavior**
The right-click action should not corrupt the model.

**Screenshots**
Before and After screenshots:
![Screenshot_20200308_before](https://user-images.githubusercontent.com/59780576/76156029-8c715a80-6148-11ea-8bbf-05d533f51ac0.png)

![Screenshot_20200308_after](https://user-images.githubusercontent.com/59780576/76156025-68ae1480-6148-11ea-8457-1989da2e1237.png)

**Versions (please complete the following information):**
 - Kubuntu Linux 18.04 LTS
 - xLights version 2020.4

**Additional context**
Issue is observed on all xLights 2020.x
Reset or Undo does not restore the model.

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

keithsw1111:
I cant reproduce this on windows.

leighfiddes:
My Analysis:
I am running Kubuntu 18.04 LTS on a desktop PC (not a laptop) with xLights 2020.x built with wxWidgets 3.1.3 which is built with GTK3. Am I the only user running this scenario?

When I right-click inside the House Preview window, to get the pop-up menu, as well as the menu being shown, the following two events are generated in this order:
1) wxEVT_GESTURE_ZOOM with Zoom Factor = 1, Start = false, End = true.
2) wxEvt_GESTURE_ROTATE with Angle = 0, Start = false, End = true.

[...]

PR #2022, opened 2020-04-21
Validate gestures

leighfiddes:
This fixes issue #1964 on my system (Kububtu 18.04 desktop no touchpad).
I have no way to validate this on other environments so can someone with a touchpad verify that gestures are fully functional please? Perhaps @dkulp who added the gesture functionality?


leighfiddes:
Thanks Daniel. Sometimes it feels like shouting in the wind with Linux issues.

PR #2028, opened 2020-05-01
Improve custom model load and move performance

priemd:
Improve custom model, especially large models in terms of X/Y/X size, when opening xLights and when trying to move large models around in the layout.  GetCustomMaxChannel is called twice for each custom model on open and the same when you attempt to move a model in the layout tab.  GetCustomMaxChannel was looping through all layers/rows/columns to find the max node count and was taking ~5-6 seconds on a particular 3-D model which was 100x100x100, which is attatched.  This call is now down to ~.5 seconds on the same model.

Even for normal non 3d models performance is much improved.
   
[G-Sphere.xmodel.zip](https://github.com/smeighan/xLights/files/4562293/G-Sphere.xmodel.zip)


keithsw1111:
Not sure this will work if the largest channel is at the end of the line or end of a layer

PR #2048, opened 2020-05-14
Default wiring, custom model and node layout dialogs to maximized

priemd:
All 3 of these dialogs generally require maximum screen real estate.  Maybe I'm missing something as to why they are not, but I believe the first thing most do when opening these is to maximize them.

### DIFF
--- a/xLights/ChannelLayoutDialog.cpp
+++ b/xLights/ChannelLayoutDialog.cpp
@@ -66,6 +66,8 @@ ChannelLayoutDialog::ChannelLayoutDialog(wxWindow* parent,wxWindowID id,const wx
     Connect(ID_BUTTON_OPEN_IN_BROWSER,wxEVT_COMMAND_BUTTON_CLICKED,(wxObjectEventFunction)&ChannelLayoutDialog::OnButtonOpenInBrowerClick);
     //*)
     HtmlEasyPrint=new wxHtmlEasyPrinting("xLights Printing", this);
+    
+    Maximize();
 }
 
 ChannelLayoutDialog::~ChannelLayoutDialog()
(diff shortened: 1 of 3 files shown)


keithsw1111:
I disagree ... dialogs should never open maximised ... unless they were previously maximised.

priemd:
What good does this dialog in the pic do for you without having to do another click to make it useful in full screen?  Ticky Tacky but saved me a lot of clicks working on Rotate to just make it full screen, personally I know every time I open it for my own use the first thing I do is maximize it so I can actually see something and make it useful.  How about a user preference ;)

image

computergeek1507:
I'm with Keith, I don't think the dialogs should open maximize. The user can just double click the title bar to maximize.

priemd:
I rest my case after this and if you still don't agree then go ahead and close it but these dialogs are all Modal dialogs, the intent of a Modal dialog is to take attention away from the main program putting it in a sorta frozen state waiting for interaction from the user with the dialog.... hence when these are open you can't do a single thing in xLights besides interact with the dialog, on macOS if you attempt to click anywhere on the main app with the dialog open you get an annoying sound letting you know you can't touch it.  On windows the window frame of the dialog itself flashes 3 or 4 times, probably makes an annoying sound as well but I don't have audio connect to VM.

priemd:
Thanks for the feedback, appreciate it.

dkulp:
For the record, I don't think they should be maximized either.   I almost never have anything maximized on my screen.   Could the default be a bit larger?  Maybe as long as it still fits OK on the lower res screens.   But not maximized.

priemd:
Fair enough, case closed, but then they shouldn’t be modal dialogs and block interaction with the main app like they do.

priemd:
These folks wrote the book on UX not me, https://www.nngroup.com/articles/modal-nonmodal-dialog/

AzGilrock:
Good UI design says don't force things like that on the user leave them as they were last time the user opened it.

PR #2055, opened 2020-05-19
Updating Sync files for MIDI

hearmstinj:
MIDI sync was incorrectly calculating the hour mark by using 360,000 instead of 3,600,000 as the mod value.


keithsw1111:
Not sure what you have done but the deltas are strange ... way too many "changed" lines to feel comfortable merging it as is.

keithsw1111:
Not only that this is missing essential changes from the master branch that should not be missing.

hearmstinj:
Can you just replace all values of 360000 with 3600000 (Three hundred sixty thousand with 3.6 million) in the two files I mentioned? What's happening is if you have MIDI set with a program like Ableton, the moment 6 minutes pass, there's a resync event happening which resets the music playback to start from 0:00.

You've missed the mod calculation of an hour to 6 minutes for the buffer value for ms.

keithsw1111:
I did yesterday

PR #2057, opened 2020-05-20
Start of fixing modal dialogs disappearing on macOS.

### LINKED ISSUE #2050: macOS dialog lost when dragging to second monitor
On macOS if user has multiple monitors and Settings >> Mission Control >> 'Displays have separate spaces' is enabled (which it is by default on macOS) and you drag any dialog opened with showModal() to another screen the dialog disappears and interaction is blocked with main app as expected, you have to Force Quit / change screen resolutions / click on app icon and select 'Show Windows' etc... the usual workarounds for 'I lost a screen'.

I believe the problem is that in macOS by default apps cannot span multiple screens and there is a sorta 'virtual' space between monitors so when you drag an app from one screen to another the app stays hidden on the secondary monitor until ~80 of the app window from left to right is on the secondary monitor, hence the dialogs stay hidden.

This is not a problem on Windows as apps can span multiple screens, on Ubuntu 19.04, not an issue either as a modal dialog is not allowed to be moved/positioned on any other screen other than the one which xlights main is present on.

There are a few options I've found to resolve the issue and each essentially do the same thing which is to move the dialogs parent to the top most window of app, which is what native dialogs like 'Open File' are doing, just wanted to open a discussion before making mass changes to all the dialogs and submitting, maybe I'm missing something simpler....

Fix 1:  Add WXOSX if around the Create of each of the affected dialogs, setting parent to null and adding wxDIALOG_NO_PARENT flag WX then defaults parent to the top level window (interaction is still blocked)
```
#if defined(__WXOSX__)
    Create(nullptr, id, _("Custom Model Wiring"), wxDefaultPosition, wxDefaultSize, wxCAPTION|wxRESIZE_BORDER|wxCLOSE_BOX|wxMAXIMIZE_BOX|wxDIALOG_NO_PARENT, _T("id"));
#else
    Create(parent->GetParent(), id, _("Custom Model Wiring"), wxDefaultPosition, wxDefaultSize, wxCAPTION|wxRESIZE_BORDER|wxCLOSE_BOX|wxMAXIMIZE_BOX, _T("id"));
#endif
``` 

Fix 2:  Where ever these dialogs are instantiated LayoutPanel, SetupPanel, etc instead of passing the panel as parent pass the top level app, so for LayoutPanel this would be xlights->GetParent().  Not sure if/how this might affect Windows/Linux, but could also add #if definied(__WXOSX__) here too seems like same amount of copy/paste.

Fix 3: Didn't vet this yet but could also capture mouse dragging and then in OnMove event block moving the window off the main screen (sorta like Ubuntu does OOB) but that would also change behavior for Windows users so probably not so good.

I like #1 as it is simple Copy/Paste and doesn't affect how this is already handled on Windows/Linux but you all are the veterans here and again maybe I'm missing something easier?

priemd:
Changed dialog instantiation for all affected dialogs in LayoutPanel as a start.  If this is OK I will change and test the other modal dialogs where issue persists.  Passing NULL as parent to dialog Create() fixes #2050 and maintains the same functionality as before on both WSMSX and LINUX.  I tested all the dialogs on Windows 7, Windows 10, Ubuntu 19.04 and Ubuntu 20.04, their functionality remains the same, aka show a modal dialog, block interaction with main app and wait for user input.

Per wxWidgets docs it is OK to pass null as parent to dialog, and when done the apps top level window is defaulted as parent.  xlPasswordEntryDialog in fpp.cpp was already passing null as parent and is working fine.

For the record I'm also perfectly fine with this being squashed, as there really is no gain in being able to move a blocking dialog off the main app but it's allowed on WSMSX and WXOSX (and works fine in WSMSX) so should probably be handled properly on osx or it will become a nuisance on osx if users move the dialog off screen and lose interaction, most will force quit losing any recent changes.  I only discovered the issue adding persistence to dialog size/position.

Also to note is that dialogs which call native OS dialogs under the hood already alleviate the issue, ColorDialog, MessageDialog, FileDialog, DirDialog, etc.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -1519,7 +1519,7 @@ void LayoutPanel::UpdateModelsForPreview(const std::string &group, LayoutGroup* 
 void LayoutPanel::BulkEditDimmingCurves()
 {
     // get the first dimming curve
-    ModelDimmingCurveDialog dlg(this);
+    ModelDimmingCurveDialog dlg(nullptr);
     std::map<std::string, std::map<std::string, std::string>> dimmingInfo;
     for (size_t i = 0; i < modelPreview->GetModels().size(); i++)
     {
@@ -1650,7 +1650,7 @@ void LayoutPanel::BulkEditControllerConnection(int id)
     else if (id == ID_PREVIEW_BULKEDIT_CONTROLLERGROUPCOUNT) {
         ccbe = controller_connection_bulkedit::CEBE_CONTROLLERGROUPCOUNT;
     }
-    ControllerConnectionDialog dlg(this, ccbe);
+    ControllerConnectionDialog dlg(nullptr, ccbe);
     dlg.Set(cc);
     OptimiseDialogPosition(&dlg);
 
@@ -1779,7 +1779,7 @@ void LayoutPanel::BulkEditControllerName()
         cn.push_back(it);
         i++;
     }
-    wxSingleChoiceDialog dlg(this, "Choose the controller name", "Controller Name", cn);
+    wxSingleChoiceDialog dlg(nullptr, "Choose the controller name", "Controller Name", cn);
     dlg.SetSelection(sel);
     OptimiseDialogPosition(&dlg);
 
@@ -1910,7 +1910,7 @@ void LayoutPanel::BulkEditControllerPreview()
         }
         j++;
     }
-    wxSingleChoiceDialog dlg(this, "Preview", "Preview", choices);
+    wxSingleChoiceDialog dlg(nullptr, "Preview", "Preview", choices);
     dlg.SetSelection(sel);
     OptimiseDialogPosition(&dlg);
     if (dlg.ShowModal() == wxID_OK)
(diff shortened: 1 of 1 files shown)


keithsw1111:
I struggle with the idea that dropping the parent really has no effect at all. It doesnt even change focus behaviour?

priemd:
I am not seeing any change in behavior, focus is placed on the modal dialog, dialog is placed in the same default position on the screen, interaction with xLightsMain is blocked until any of these dialogs are closed with OK/Cancel/Close/etc.

I've also been very careful to check the actual dialog implementation to make sure there is no dependency/usage of parent being passed, all good for those instantiated from LayoutPanel.

wxFileSelector Dialog which is used for a bunch of the 'export' and 'select a file' actions all over actually has a 'parent' arg which is never passed...so under the covers wxFileSelector calls wxFileDialog with NULL as parent.  All of the wxFileSelector dialogs work great and don't dissappear.

dkulp:
I just pushed a small change to ModelDimmingCurveDialog that might be a better model for handling this.   It would only affect the parent on OSX and leaves everything as-is for the other operating systems.   It also keeps the parent passed into the constructor as-is so if the parent IS needed for something, we can save it to a field prior to calling AdjustModalDialogParent.   Can you take a quick look and see if things could be updated similarly for the others?

priemd:
Looking now, thanks.

priemd:
Works as expected, no hide and seek for modal dialogs on osx.  Closing this PR and will work through the dialogs with Dan's solution since it alleviates the concern of changing behavior on the other platforms.

Thanks for the help Dan!

PR #2071, opened 2020-05-31
Model Tree multi select

priemd:
Add multi-select for tree view for all types, model/group/submodel, and keep in sync with model preview selections.  There is a lot here, I think it's pretty solid and but will monitor any issues closely after the next release.

Aside from the obvious, multi-select for all tree view items synced with preview selections, there are a few minor changes/fixes/enhancements whatever you want to call them...

- context menu in tree is aware of multiselect, delete group vs delete groups etc.
- prior context menu 'Delete Group' actually deleted all models in the group as well which I don't think was the intent as it bit me a few times on accident vs 'Delete Group'.  It is now labeled 'Delete Model / Delete Model's' and will only show up if models are selected and will only delete models, not groups.
- Prompt to confirm Model / Group delete with a list of models/groups that will be deleted in prompt.
- SubModels when selected in tree now highlight in preview, prior sub models would only highlight if they were part of a group and the group was selected
- Submodels can be selected in tree and if 'Create Group' is selected with at least 1 model selected they will be added to the newly created group.  I plan to change logic in context menu to allow group creation when only subs are selected which may be more convenient then finding them in the model group panel for large displays.
- Create Group takes all tree selections into consideration vs only the models selected in preview so if you selected 2 groups in tree, 2 submodels and 2 models all will be added to the new group.
- Besides group creation, selected submodels are ignored for all other preview right click actions, bulk edit, align, active, etc as would be expected.
- Non-3D multi-select now mirrors the same 3-click approach that Gil did for 3d, before if you had a model selected in 2D, CTRL+Click another model and then CTRL+clicked the second model without the handles prior selection was lost, now this mirrors what happens in 3D world.

TODO
- Add in appropriate model preview right click options to model tree menu
- 3D Objects tab.  I made sure business as usual on the objects tab until multi-select on Models tab is deemed stable, will adopt multi-select for ViewObjectList when it is.
- Build a new panel for when multiple selections exist which shows the currently selected items, right now if multiple selections are made the group panel and property grid are hidden so there is just a blank space.  Tooltip for Model List does show a summary of current selections for now.
- Monitor and fix any bugs I may have caused ;)

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -1434,15 +1434,19 @@ void LayoutPanel::UpdateModelsForPreview(const std::string &group, LayoutGroup* 
     }
 
     // add in any models that were not in preview but belong to a group that is in the preview
-    std::string selected_group_name = "";
-    if (mSelectedGroup.IsOk() && filtering) {
-        selected_group_name = TreeListViewModels->GetItemText(mSelectedGroup);
+    wxArrayString selectedGroupNames;
+
+    if (mSelectedGroups.size() > 0 && filtering) {
+        for (int i = 0; i < mSelectedGroups.size(); i++) {
+            wxString selectedName = TreeListViewModels->GetItemText(mSelectedGroups[i]);
+            selectedGroupNames.Add(selectedName);
+        }
     }
 
     for (const auto& it : xlights->AllModels) {
         Model *model = it.second;
         bool mark_selected = false;
-        if (mSelectedGroup.IsOk() && filtering && (model->name == selected_group_name)) {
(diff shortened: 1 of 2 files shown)


KMojek:
Looks good overall but I see one weird thing... start with a layout of just three models:
* select first model in preview --> selected in preview and list
* ctrl-select second model in preview --> both selected in list but only the first in preview

priemd:
I will take a look in the morning for sure, I think I tested this scenario a thousand times as it’s the basic test in terms of syncing preview and tree but maybe I broke it working on other issues I’ve came across.  Thanks for the heads up!

PR #2072, opened 2020-05-31
fix prop grid blanking + highlight of deep nested groups

priemd:
- Fixes prop grid blanking when selection is cleared.
- Fix deeply nested group models not being highlighted when selected
- Add back ifdef LINUX I had commented out for testing


priemd:
still not quite right in 2d

PR #2073, opened 2020-05-31
Fixes for multi select in tree

priemd:
Fixes:
- Revert 2d click behavior to what is was prior
- Check if selections > 0 in OnSelectionChange to resolve prop grid not showing issues
- Fix highlighting of deeply nested models

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -2772,45 +2772,37 @@ bool LayoutPanel::SelectMultipleModels(int x,int y)
     int modelCount = FindModelsClicked(x, y, found);
     if (modelCount == 0)
         return false;
-    
-    bool treeSelectionChanged = false;
+
     propertyEditor->Freeze();
     clearPropGrid();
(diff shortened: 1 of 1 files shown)


keithsw1111:
Create group is still broken

keithsw1111:
click on a model on layout, ctrl click a second model ... the second model does not highlight on layout but it does in the list and the panels all blank out ... that also seems wrong.

keithsw1111:
Should I just reverse all these changes out ... they seem to be more trouble than they are worth right now.

priemd:
You’re the boss, your call. I did a ton of testing on MacOS and could never recreate the group issues that you are seeing on Windows, I shared a video and a gif.  I am happy to continue digging in now that I know this seems to only be a problem outside of macOS but as I said earlier I’m sorry but this probably won’t be until tomorrow.

keithsw1111:
I believe I have unwound all the changes from the master branch and moved all the pending changes into another public branch where it can continue to be worked on until we are all happy with it.

PR #2074, opened 2020-06-01
Multi model select in list

priemd:
The issue with MSW is when TreeListViewModels->Select() and Unselect() are called they do not fire the OnSelectionChanged() Event which is where all the logic is to keep track of current selections and keep things in sync.

Seems to be a bug with wxWidgets for MSW.  Maybe there is a better way to solve this, maybe creating a synthetic event and calling OnSelectionChanged() manually.


priemd:
my bad, didn't mean to pull on master

PR #2083, opened 2020-06-09
Model Tree MultiSelect Fixes/Cleanup

priemd:
lots of fixes/cleanup for model tree multi select to try and ensure there are no changes or hinderance to existing layout panel functionality and experience is business as usual for all existing preview MultiSelect capabilities across platforms.  This also adds the existing right click model preview options to tree view for single and multi selection.

There are a few open items which I have thoughts/questions on and didn't want to make the changes without feedback from the rest of you all which I'll comment on this PR shortly.

### DIFF
--- a/xLights/ControllerConnectionDialog.cpp
+++ b/xLights/ControllerConnectionDialog.cpp
@@ -459,10 +459,10 @@ void ControllerConnectionDialog::Get(wxXmlNode* node) {
             if (CheckBox_PixelDirection->IsChecked()) node->AddAttribute("reverse", wxString::Format("%d", PixelDirection->GetSelection()));
         }
     }
-    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RGBEFFECTS_CHANGE, "ControllerConnectionDialog::Get");
-    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_MODELS_CHANGE_REQUIRING_RERENDER, "ControllerConnectionDialog::Get");
-    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RELOAD_PROPERTYGRID, "ControllerConnectionDialog::Get");
-    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RELOAD_MODELLIST, "ControllerConnectionDialog::Get");
+//    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RGBEFFECTS_CHANGE, "ControllerConnectionDialog::Get");
+//    xLightsApp::GetFrame()->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_MODELS_CHANGE_REQUIRING_RERENDER, "ControllerConnectionDialog::Get");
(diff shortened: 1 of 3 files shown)


priemd:
My thoughts up for discussion on a few thoughts/questions I had while making these changes:

1 - CUT in preview for current release currently cuts/deletes all selected models while COPY/PASTE only deals with primary/base selected model.  Should Copy/Paste handle multiple models like CUT or should CUT only handle primary/base selected model while another key like 'DELETE' should handle all selected models, today both CUT and DELETE behave the same, delete all select models.

2 - Distinguish between Tree 'Create Group from Selections' which includes Groups/Models/Submodels vs 'Create Group' in preview right click only is concerned about selected models.  Just don't want to confuse folks with these, maybe change 'Create Group' in preview to 'Create Group from Selected Models'?  Maybe a combo of both?.....Regardless of where menu is initiated whether tree or preview have 2 options 'Create Group from Selections' and 'Create Group of Selected Models' or something like that.  I'm a fan of create group from selections as it makes it really easy to add a new group visually making selections in preview vs fine tuning in the model group panel but want to make sure folks aren't confused with the difference.

3 - Combine  OnModelsPopup and OnItemContextMenu handlers into one, just wanted to make sure that tree menu is all good with everyone before doing so and reducing the redundancy.

keithsw1111:
There is still weirdness in this change. If i ctrl select a bunch of models then ctrl select a model other than the focussed model instead of removing it from selection it seems to reset the selection mode. That aint right.

AzGilrock:
Not sure if you mean in the tree or the actual models but they should work as a tri-state.  If you ctrl click a model that is already selected but not the focused model it should become the focus.  Then the next click removes it from the selection group.

keithsw1111:
Yeah ... that is what is broken. Clicking the models on the layout dont work that way ... at least not all the states.

priemd:
In 2D current release it is not tri-state.  If you Ctrl-Click a bunch of
models and then Ctrl-Click one of the selected models the other selections
go away and only the last clicked model is highlighted.  I tried to keep
things exactly as they were before for click behavior.

AzGilrock:
Whether it worked that way before it doesn't sound right.  A control click should never deselect any other models than what you are clicking.  It should add or remove it from the group.  I'm not sure why it needs to be different from the 3D behavior] other than the feature when if you continue to click it would cycle through models that overlapped.  I think the answer is 2D should be deleted...lol.

priemd:
Here is 20.23 2d behavior

2d-click-20-23

I'm happy to make 2d behavior mirror 3d if that is what is desired.

AzGilrock:
Yeah I believed you I'm just not sure why we would have had it that way.  There may be a good reason for it but I don't know.

priemd:
Ok I take that back.  The behavior above is how it is in OSX 20.23 and prior, it was a bug.  I always thought it was strange in 2D on osx that this was the behavior.  It is because TreeListCtrl on MSW doesn't fire OnSelectionChanged when Tree->Select() is called so the subsequent UnSelectAllModels() doesn't get called like it does on OSX.  I will fix it.

priemd:
On 20.23 and prior on MSW here is what I think is a bug, but want to confirm in case I am wrong and this is expected behavior.  When selected model is Ctrl-Clicked it gets unselected but the Active handles don't move to one of the other selected models.  

Bug? or Expected Behavior?

msw-2d-click-bug

AzGilrock:
It probably should pick one to be the focus.

PR #2085, opened 2020-06-10
setup tab align/show status led, osx fix show dir btn label cut off

priemd:
- LedPing was not showing until controller selected and was not aligned with the other buttons.
led-ping-before
led-ping-after

- On osx when temp show directory btn was set to 'Restore to Permanent' the 'nt' was cut off.
label-cut-off


KMojek:
Did some quick testing on Windows; looks good to me.

PR #2092, opened 2020-06-12
fix/restore 2d msw and osx click behavior

priemd:
This should fix 2d click and mirror the same behavior of 3d across platforms.  I was trying to be very careful not to disrupt click behavior and followed osx behavior for 2d which is different than msw in current release, my apologies.  This also fixes prior bug with msw where there is no selected primary object with handles.


keithsw1111:
It still does not work ... I am going to have to fix this myself.

PR #2104, opened 2020-06-17
fix matrix scaling on bulkedit resize

### LINKED ISSUE #2044: Matrix Model Resize - Match Width/Height/Size makes 2nd model larger than original
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
When resizing a matrix in the layout window, selecting Resize > Match Height, Width or Size causes the 2nd and subsequent matrixes to be taller and sometimes wider than the original (depending on the Direction set. Matrixes set to be in the Horizontal direction come out taller than the original and the same width, whereas Vertical direction matrixes have both their height and width widened.

**To Reproduce**
Horizontal Direction Matrix:
1. Create a matrix in the layout window.
2. Ensure the direction is set to Horizontal
3. Copy / paste this matrix, and align roughly next to the other one. They should look to be the same height/width
4. Select one of the matrixes, ctrl+click on the other matrix
5. Right click on the 2nd matrix
6. Select Resize > Match Height or Match Size (doesn't matter)
7. 2nd matrix will be resized slightly taller than the first, even though it was already the same height.

Vertical Direction Matrix:
1. Create a matrix in the layout window.
2. Ensure the direction is set to Vertical
3. Copy / paste this matrix, and align roughly next to the other one. They should look to be the same height/width
4. Select one of the matrixes, ctrl+click on the other matrix
5. Right click on the 2nd matrix
6. Select Resize > Match Height, Match Width or Match Size
7. 2nd matrix will be resized slightly taller AND slightly wider than the first, even though it was already the same height/width

I have quickly tested other models (tree, candy cane, arches) and these do not appear to be affected by this.

**Expected behavior**
Both matrixes should be the same height and width after resizing.

**Screenshots**
Horizontal_before_resize

Horizontal_after_resize

vertical_before_resize

vertical_after_resize NOTE: 2nd model was copy/pasted and moved to under the original after the resize to show the width difference.

 - OS: Windows 10
 - xLights version 2020.19, but was also noted in 2020.17

**Additional context**
NIL

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced: N/A
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file: N/A
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file: N/A

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

priemd:
saw this in my own display, fixes #2044.  If selected base object is matrix and resizing another matrix to match then also set matrix scale to match.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -4641,8 +4641,9 @@ void LayoutPanel::PreviewModelResize(bool sameWidth, bool sameHeight)
     int selectedindex = GetSelectedModelIndex();
     if (selectedindex < 0) return;
 
+    std::string selectedType = modelPreview->GetModels()[selectedindex]->GetDisplayAs();
     std::vector<std::list<std::string>> selectedModelPaths = GetSelectedTreeModelPaths();
-
+    
     CreateUndoPoint("All", modelPreview->GetModels()[selectedindex]->name);
 
     if (sameWidth)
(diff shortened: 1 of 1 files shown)


keithsw1111:
Did you test this on trees .. which are also matrices? Is it only relevant to matrices ... what about other box sized models?

priemd:
I did test a tree quickly and visually it looked spot on, however looking at the real numbers scale does end up 1/10 off for x, and 3/100 off for Y/Z in my display, will check other boxed size models now and adjust.

PR #2105, opened 2020-06-17
Set MatrixScale for all models just in case

priemd:
Matrices were the only models I could visually see a difference in after resize but just in case set the ScaleMatrix props to be the same on all models being adjusted to match height/width or both, not just matrix as other models sometimes were 1/10th off, others were 3-5/100th off in scale properties which is why #2044 only reported matrix resize issues as they were not visible, but this should keep all things square on resize.

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -4641,8 +4641,9 @@ void LayoutPanel::PreviewModelResize(bool sameWidth, bool sameHeight)
     int selectedindex = GetSelectedModelIndex();
     if (selectedindex < 0) return;
 
+    std::string selectedType = modelPreview->GetModels()[selectedindex]->GetDisplayAs();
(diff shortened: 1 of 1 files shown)


AzGilrock:
Hmm...  I haven't looked at anything but my gut tells me you can't just take the scaling matrix from one model and stuff it into other models.  I'd be concerned the most about PolyLines and the two point and three point model types.  For models like that moving a point accomplishes the same thing as scaling so if you called something to set the width and you pulled a scale matrix that does something like doubling and stuff it in then you would end up with something twice as large.  For testing this you need to make sure the selected model has a matrix thats not 1,1,1 and then use all other model types with varying scale matrix values.

priemd:
bad idea here, Gil is correct. Visually I only see issues with matrix like the open issue also suggests, I had this limited to matrix only in prior PR, will recheck and resubmit that one.

PR #2109, opened 2020-06-20
match size/scale for boxed models of the same type

### LINKED ISSUE #2044: Matrix Model Resize - Match Width/Height/Size makes 2nd model larger than original

priemd:
A second attempt at fixing resize matching issue for boxed models, thank you Gil and Keith for the feedback on my failed last attempt.

This should fix #2044 which was more than just matrix (thank you Kieth for the tip), it was all boxed models...Matrix, Cube, Sphere, Tree, Custom and DMX.  I couldn't find an exact science to match size of boxed models of various types so this will only correctly match size/scale for boxed models of the same type or for custom 3d models those that have the exact same fingerprint.  If there is not a match in type/fingerprint for box models OR it is not a boxed model at all it defaults to prior behavior, just set W/H.

This also fixes the following issue for DMX models where bulk edit resize wasn't working at all, any models selected that weren't the base selection were sent off into the abyss as a tiny dot.  

dmx-resize

If I'm still wrong and there is a better way to go about this, that is unified across all models types I'm all ears...

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -4732,35 +4732,67 @@ void LayoutPanel::PreviewModelResize(bool sameWidth, bool sameHeight)
     std::vector<std::list<std::string>> selectedModelPaths = GetSelectedTreeModelPaths();
 
     CreateUndoPoint("All", modelPreview->GetModels()[selectedindex]->name);
+    
+    Model* selectedModel = modelPreview->GetModels()[selectedindex];
+    std::string selectedType = selectedModel->GetDisplayAs();
+    int width = selectedModel->GetWidth();
+    int height = selectedModel->GetHeight();
 
-    if (sameWidth)
+    bool isBoxed = false;
+    if ((dynamic_cast<ModelWithScreenLocation<BoxedScreenLocation>*>(selectedModel) != nullptr)) {
+        isBoxed = true;
+    }
+    
+    bool isCustom3d = false;
+    wxString customFingerprint = "";
+    
+    // check if custom 3d and set model fingerprint if so
+    if (selectedType == "Custom") {
+        if (wxSplit(customFingerprint, '|').size() > 1) {
+            customFingerprint = selectedModel->GetModelXml()->GetAttribute("CustomModel", "");
+            isCustom3d = true;
+        }
+    }
+    
+    for (size_t i = 0; i < modelPreview->GetModels().size(); i++)
     {
-        int width = modelPreview->GetModels()[selectedindex]->GetWidth();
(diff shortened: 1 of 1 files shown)


priemd:
and attempt #3 coming, didn't commit last change, sorry been a long week

PR #2119, opened 2020-07-03
Remove Black Cherry Cosmos

ghormann:
Remove shader that no longer works.


keithsw1111:
I fixed it instead of deleting it.

PR #2143, opened 2020-07-25
FSEQ VariableHeader parsing improvements

Cryptkeeper:
This PR includes several code hardening improvements for the `parseVariableHeaders` function. The goal of this is to help protect against corrupted FSEQ files (accidentally or maliciously) and improve logging output to assist with debugging potential issues.

Assuming there is interest in these changes, I'll upstream them to fpp's `FSEQFile` implementation.

- Zero length VariableHeaders are now ignored and print a warning. Some previous logic vaguely prevented this, with some issues. While xLights could parse these VariableHeaders all the same, the reason to not is simply a continuation of the current implementation and reduces needed handling.
- The data length of each VariableHeader is now checked against the `header` buffer. This prevents an incorrect length from reading out of bounds and unintentionally accessing memory, or crashing the program.
- VariableHeaders with unrecognized codes (not `mf` or `sp`) print debug messages to increase visibility into the read process for modified sequence files.
- An error is now logging if the data is not null terminated (for recognized VariableHeaders). This helps locate errors caused by wrongly encoded sequence files. 

**Misc. Changes**
- Magic number usage has been reduced and moved into a named const, `FSEQ_VARIABLE_HEADER_SIZE`. 
- Added debug output when reading VariableHeaders to improve visibility.

### DIFF
--- a/xLights/FSEQFile.cpp
+++ b/xLights/FSEQFile.cpp
@@ -414,30 +414,46 @@ inline bool isRecongizedVariableHeader(uint8_t a, uint8_t b) {
 const int FSEQ_VARIABLE_HEADER_SIZE = 4;
 
 void FSEQFile::parseVariableHeaders(const std::vector<uint8_t> &header, int readIndex) {
-    while (readIndex < header.size() - 5) { // todo: where is 5 from? off by one from 4?
-        const int variableLength = read2ByteUInt(&header[readIndex]);
+    const int VariableLengthSize = 2, VariableCodeSize = 2;
 
-        if (!variableLength) {
-            LogErr(VB_SEQUENCE, "VariableHeader has 0 length data: %c%c", header[readIndex + 2], header[readIndex + 3]);
+    // when encoding, the header size is rounded to the nearest multiple of 4
+    // this comparison ensures that there is enough bytes left to at least constitute a 2 byte length + 2 byte code
+    while (readIndex + FSEQ_VARIABLE_HEADER_SIZE < header.size()) {
+        const int dataLength = read2ByteUInt(&header[readIndex]) - FSEQ_VARIABLE_HEADER_SIZE;
+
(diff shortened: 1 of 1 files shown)


xLights/FSEQFile.cpp
```diff
@@ -404,20 +404,72 @@ void FSEQFile::preload(uint64_t pos, uint64_t size) {
...
+            // there is no reasonable way to recover from this error - the reported dataLength is longer than possible
+            // return from parsing variable headers and let the program attempt to read the rest of the file
+            return;
+        } else if (!isRecongizedVariableHeader(header[readIndex], header[readIndex + 1])) {
+            // avoid reading unrecongized variable headers
+            // each 2 byte length can consume up to 65k of memory when allocating vheader->data
+            LogErr(VB_SEQUENCE, "Unknown VariableHeader code: %c%c, length: %d bytes", header[readIndex], header[readIndex + 1], dataLength);
+            
+            // advance the length of the 2 byte code + the length of the data
+            readIndex += VariableCodeSize + dataLength;
```
dkulp, line 445: This is the only part of the change that I'm concerned about.   Having a list of "known"  variable headers and enforcing rules on those (like the null termination) makes a lot of sense.   However, ignoring the non-recognized headers is a bad idea.    For things like FPP Connect that copy the headers over to the new fseq files that it generates, those extra headers would be lost.     On the FPP side, the fsequtils command line utility can be used to convert fseq files from format to format (or compression level/type to another) and again, these headers would be lost.   While I'm not sure if anything is using the variable headers outside the known headers, the idea of losing data seems to be a step in the wrong direction.

Cryptkeeper: I wasn't aware it utilized in the sole in memory copy as its source of truth of re-encoding output. I'll move this from hard fail to a warning. :+1:

PR #2178, opened 2020-08-21
add disperse feature to layout

priemd:
Add 'Disperse' feature, hope that is the appropriate term, for models and 3d objects to layout panel.  The idea is to take 1 or more selected models and be able to 'disperse' them in any direction with an offset from the starting point of a target model.  Users can use the default disperse by direction options in right click menu which will disperse selected items with a default offset of 20 (tbd if this is appropriate) from the primary selection or use the 'Select Options' menu item which pulls up a dialog where they can rearrange the order to disperse as well as select any target model as the base for disbursement for selected models as well as the offset.  Dialog also prompts on single selection right click >> disperse so a user can select an existing target model to place the newly created/imported model next to in any direction.  For now the 'Select Options' is top on the menu as we don't track selection order (treelist just gives us what is selected in sort order) but if I can figure out a decent way to track selection order so users don't have to manually manipulate order this should probably dropped to the bottom.

### DIFF
--- a/xLights.xcodeproj/project.pbxproj
+++ b/xLights.xcodeproj/project.pbxproj
@@ -10,6 +10,7 @@
 		1ECB4F631FF4D014006D57AA /* BulkEditControls.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1ECB4F601FF4D013006D57AA /* BulkEditControls.cpp */; };
 		1ECB4F641FF4D014006D57AA /* BulkEditSliderDialog.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 1ECB4F611FF4D014006D57AA /* BulkEditSliderDialog.cpp */; };
 		3D585F311E7E541400A3F84F /* UtilFunctions.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 3D585F301E7E541400A3F84F /* UtilFunctions.cpp */; };
+		63B4360324EE23250000C90D /* DisperseOptionsDialog.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 63B4360224EE23250000C90D /* DisperseOptionsDialog.cpp */; };
 		6701999F1CE5A03200AE9B7E /* RenderProgressDialog.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 6701999D1CE5A03200AE9B7E /* RenderProgressDialog.cpp */; };
 		67025C6E20D7E80900BF1AC6 /* Assets.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 67025C6D20D7E80900BF1AC6 /* Assets.xcassets */; };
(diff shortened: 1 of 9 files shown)


keithsw1111:
So why wouldnt you use the distribute function and make that more 3D? conceptually ... you put a model at each end and the rest in the middle and it distributes them between the extremes in a straight line? This would seem easier than guessing a dispersion amount and hope the end one ends up in the right place?

There would be challenges I guess in terms of working out which is the end ... but a 3d line distance between model centres with the furthest from the main selected model selected as the end is likely the best strategy,

keithsw1111:
It would also allow arbitary dispersion directions defined by the line between the end models and it does not need more popup dialogs.

priemd:
The idea started with trying to move x number of models near another model where you can then fine tune placement dragging handles or yes using existing distribute features, for example I create a new star or 2 and I know i want them near an existing star in my layout that was already placed.  All of this is trivial in 2d but when you add in z world it takes a lot of mouse movement and zooming/panning in order to get one model lined up with another.  Today you can either remember x/y/z of the model you wish to move another near and manually key those in or Right Click >> align top >> then right click >> align left/right >> then right click align front >> then since there is an overlap you have to find the moved model in tree, click to highlight >> then you can make your fine tuned adjustments to make it above/below/left/right etc.  It's a short cut to get things quickly 'near' another model in x/y/z world where you can then use the existing features like distrubute or dragging to fine tune.

AzGilrock:
The way I originally designed it you could easily copy paste models and move them along one axis for near perfect spacing.  Then a couple people whined about the paste being in the exact spot as the copy so now it pastes offsets that I dislike.  I’d like to see a paste in place option to restore my original behavior.

priemd:
I have always disliked the offset as well, maybe not moving it right, but dropping it down.  I am going to close this for now as I still think it is very useful for those that create/import new models and want them close to another model in the display in 3D world, most user error not understanding how they can use the keyboard/mouse to pan/zoom/rotate to get a better angle to line things up but it's a common thing I hear as to why people are still in 2d, right or wrong on their par.  I called it 'disperse' because it's not a distribute in the sense of distribute with any software that has such feature like Illustrator/vectric/blender/etc will distribute selected items evenly across X or Y depending on if you do horizontal or vertical which is exactly what distribute does in xl today perfectly, this was intended to get models on the same x/y/z plane with an offset where you could easily then adjust x/y with the handles.  In fact align front/back doesn't currently really work for custom models which even though are 2d have a z-scale, especially for custom models with a large x/y footprint there are times when align front/back throws the model you are attempting to align to another whether front or back off into outer space, this at least solves that and I can implement fix. But to Keith's point I dislike dialogs and popups as well.  He has suggested tracking selection order so I am going to work on that and then this may be more user friendly not needing the dialog to adjust order when multiple models are selected, but as a last resort.  Although I don't think tracking selection order is an exact science when you factor in group selection, shift select in tree, hold shift and press UP/DOWN in tree or shift drag select in preview as treelistctrl gives you selections in the order in which they are currently in the tree, not order in which they were selected.

PR #2259, opened 2020-10-03
enh: add "Cut Effects" option to right-click menu for sequence elements

Cryptkeeper:
This PR introduces a "Cut Effects" function (for both the Row and Model right-click menus) as requested in https://github.com/smeighan/xLights/issues/2170. The code re-uses the preexisting `CopyModelEffects` functionality to avoid code duplication, before then creating an undo step and deleting the relevant effects. 

I strayed away from simply posting a `EVT_COPY_MODEL_EFFECTS` event and then duplicating the `ID_ROW_MNU_DELETE_ROW_EFFECTS` handling since the first event will be queued whilst the second would be sync, introducing a race condition and deleting the effects prior to copying.

I have included a call to `UnselectEffect` when cutting to avoid dangling `selectedEffect` pointers if the currently selected effect had been cut. This condition could be gated by a condition to only call `UnselectEffect` if it is within the cut effects, but I'm not sure if it is necessary.

I've successfully tested these changes without experiencing regressions, but given the wealth of various selection and deletion methods within the xLights codebase, additional eyes would be appreciated.

### DIFF
--- a/xLights/sequencer/EffectsGrid.cpp
+++ b/xLights/sequencer/EffectsGrid.cpp
@@ -7101,6 +7101,26 @@ void EffectsGrid::StretchAllSelectedEffects(int deltaMS, bool offset) const
     }
 }
 
+void EffectsGrid::CutModelEffects(int row_number, bool allLayers)
+{
+    CopyModelEffects(row_number, allLayers);
+
+    mSequenceElements->get_undo_mgr().CreateUndoStep();
+    if (!allLayers) {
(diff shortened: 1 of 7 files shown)


MarkA75160:
2020.42   Cut works on ROW, but not on MODEL

MarkA75160:
This is worse than I previously thought.  Cut not working is a minor nuisance but I now see that its erroneously deleting the top timing track instead of the model effects and that is very bad  :(

PR #2459, opened 2021-01-15
Added '--no-check-certificate' flag to wget command

jdye64:
Certain versions of wget and linux environments require the use of the '--no-check-certificate' flag to allow for log4cpp to be downloaded by the Makefile. If that flag is not present the the standard 'make' command enters an infinite loop of retrying to download and then failing to open the *.tar.gz since it obviously doesn't exist.

Another option is to have a user modify their ~/.wgetrc file to not require '--no-check-certificate' but there seems to be no negative consequence to just adding it in the Makefile as the standard.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -45,7 +45,7 @@ log4cpp: FORCE
 	@printf "Checking log4cpp\n"
 	@if test "`log4cpp-config --version`" != "1.1"; \
 		then if test ! -d log4cpp; \
-			then echo Downloading log4cpp; wget --no-verbose -c https://nchc.dl.sourceforge.net/project/log4cpp/log4cpp-1.1.x%20%28new%29/log4cpp-1.1/log4cpp-1.1.3.tar.gz; \
+			then echo Downloading log4cpp; wget --no-check-certificate --no-verbose -c https://nchc.dl.sourceforge.net/project/log4cpp/log4cpp-1.1.x%20%28new%29/log4cpp-1.1/log4cpp-1.1.3.tar.gz; \
 			tar xfz log4cpp-1.1.3.tar.gz ;\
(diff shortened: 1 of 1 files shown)


jdye64:
Hmm the build failure seems to be something linking related. I just attempted a docker build, using the method in the documentation from master and see the same error without this change. Anyone have any ideas?

priemd:
The build checks have been failing for awhile, nothing you did to break it :)

jdye64:
Ahh ok :)

- Do you know if they have only been failing on linux?
- Do most people use xLights with windows then?
- In expansion to question about windows, does anyone in the community possibly know a fix or is there little interest in fixing linux builds since most people are windows based?
- I would certainly be willing to help figure it out just curious of the current state there first.

jdye64:
@priemd thoughts on the above? I sense most people are using windows. Honestly if that is the answer I'll just bite the bullet and use windows, although I really don't want to do that =)

cjd:
There should be no need for a no-check-certificate (checked/confirmed from multiple locations - including autobuilds)
If you need that - then something/someone is getting between you and the sourceforge server it is downloading from.
The autobuilds are working currently without issue (see 
https://travis-ci.org/github/smeighan/xLights showing the build of 2021.03 working/releasing 2 days ago.)
Adding the --no-check-certificate would allow for folks to be MitM'd
A safer/better thing would be to check what is causing the error on your end - is there an old/broken certificate in the chain, or perhaps an ISP doing something a bit 'off'

jdye64:
Delayed response here, apologies. Turns out it was a new Ubiquiti Dream Machine Pro I had installed in my home network and setup to do "network sniffing" will close the issue.

PR #2461, opened 2021-01-16
avoid redundant model list reloads

priemd:
In RecalcStartChannels(), RELOAD_MODELLIST work was being added regardless of whether or not there were channel changes to models causing unnecessary/redundant reloading of the model tree list in Layout when app is launched/adding a new group/new submodels/etc.  These reloads become more of a pain as a display grows, especially on macOS/gtk with the treectrl implemenations.

### DIFF
--- a/xLights/models/ModelManager.cpp
+++ b/xLights/models/ModelManager.cpp
@@ -485,7 +485,9 @@ bool ModelManager::RecalcStartChannels() const {
 
     ResetModelGroups();
 
-    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RELOAD_MODELLIST, "RecalcStartChannels");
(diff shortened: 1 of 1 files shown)


keithsw1111:
Are you really sure about this. It is possible but one of the reasons i adopted the ASAP work approach is it was meant to make it unimportant that we request multiple updates ... the updates are all consolidated and then just performed the once and in the optimal order in the DoWork function.
That being said there have been times when doing work that we unnecessarily trigger the doing more work but in this case we seem to be recalculating start channels and when that happens the model list does need to be reloaded if the start channel changes. So excluding the work if the channels dont change might make sense ... but excluding it all the time runs the risk that there is a path in the code which hasnt already asked for it and was relying on this call.
I do have plans to totally rework the way we handle models in the first half of this year because this endless reloading of models is an abysmal way to do things.

PR #2489, opened 2021-02-02
import-from-xsq-package

priemd:
Import Effects from a Packaged Sequence.  Assets (images/videos/shaders) for mapped effects which exist in the package will be imported into the user's show directory if they don't already exist under a new subfolder called "ImportedAssets/<name of sequence>".   Also if a Faces Effect of type Matrix is mapped it will create the Face Definition on the target model. 

Existing .xsq/.xml only sequences work as they did before.

### DIFF
--- a/xLights.xcodeproj/project.pbxproj
+++ b/xLights.xcodeproj/project.pbxproj
@@ -539,6 +539,8 @@
 		67FE98551EB8C9BF00C5E804 /* ViewsModelsPanel.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 67FE98531EB8C9BF00C5E804 /* ViewsModelsPanel.cpp */; };
 		67FF398F1D57F5C000290DB4 /* MusicXML.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 67FF398E1D57F5C000290DB4 /* MusicXML.cpp */; };
 		67FF39931D57F5D000290DB4 /* NoteImportDialog.cpp in Sources */ = {isa = PBXBuildFile; fileRef = 67FF39911D57F5D000290DB4 /* NoteImportDialog.cpp */; };
+		FA9C5DE125D253CC00886D4A /* MediaImportOptionsDialog.cpp in Sources */ = {isa = PBXBuildFile; fileRef = FA9C5DDF25D253CC00886D4A /* MediaImportOptionsDialog.cpp */; };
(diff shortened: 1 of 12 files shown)


computergeek1507:
This is pretty cool, good work. I don't how we should solve this but PPD creates 'Matrix.avi' and 'Tree.avi' for each sequence and with just one "ImportedAssets" folder for all the imported resources they will override each over if the files are the same name.

priemd:
"Matrix.avi" != "Tree.avi".... what am I missing :) ??  They do get placed in Images/Shaders/Videos/etc subfolders but wouldn't help if a file has the exact same name.

computergeek1507:
They create a "Matrix.avi" for each sequence. If you import two sequences from PPD the second "Matrix.avi" will override the first.

priemd:
Ok I think I see what you are saying.....  2 different sequences, each have a Tree.avi.  This is handled already at least on import.... a subfolder is created under ImportedAssets by sequence name..... so ImportedAssets/<name of sequence>, but it is also possible that FixFile() already found a Tree.avi in the Show Folder when the xsq you are mapping was loaded which that code was already in place.  Not really sure what could be done to solve that......tell PPD to not name their files generically....

priemd:
I just didn't want to cause file bloat on the machine importing every single mapped asset with a new unique name......which is where FixFile() happening when the sequence to map is loaded comes in handy......if a file is already in the show directory then we won't import yet another copy.

dkulp:
If the "target" sequence that you are importing into is itself in a subdirectory of the show directory, would putting the assets for that sequence into a subdirectory of that directory instead of the show directory itself make more sense?

AzGilrock:
I haven't looked at what this does but I do organize my stuff like Dan suggests.  I like all assets associated with a sequence in its own subfolder.  I used to try to keep audio in a common folder but now I like to just be able to copy one folder and know I'm getting everything.

priemd:
We can do that, check if the target sequence is in a subfolder and then use that as the drop target for assets.  

What about files that were found/corrected by FixFIle when the sequence to map was loaded?  Ignore it unless the package doesn't have the asset in which case make a new copy in the sequence subfolder?

I think you'd end up with a little file bloat, most likely not for videos/images but the same shaders/faces are commonly used across multiple sequences coming from our download, but storage is cheap.

priemd:
I just pushed some updates based on the feedback.  There are default assumptions made based on target xsq location (whether it is in base of show folder, or in a separate folder) for the imported media targets and all can be modified by the user in the mapping dialog to suite there specific needs for how they manage their show folder.  I've laid out the options dialog to allow for other shortcuts/options and added a new class for import options that can easily be added to with additional feat requests that I may not have thought of.

PR #2503, opened 2021-02-12
fix memory/loop issue in effect preview generation

### LINKED ISSUE #2496: Presets Memory Issue
Large preset libraries cannot be previewed sequentially.  The more preset you preview, the more memory xLights uses until the system runs out of resources.  Also, it is impossible to import larger preset libraries.  The system eventually runs out of memory and locks up.  It appears that the memory allocated to display the preset is not being released.  xLights 2021.03

priemd:
This resolves #2496 

I found numerous instances in the preset files from the user who reported the issue where the effect presets had negative values for rows, such as a preset with 8 layers had values of -4,-3,-2,-1,0,1,2,3.  The negative row values where causing a near infinite loop in the loop on WriteGIFForPreset() in TabConvert.cpp which added the layers for the preview.  Negative int was converted to a HUGE number when being compared to size_t returned from GetEffectLayerCount() and memory didn't last long.

Not really sure where the (-) row values came from as I was unable to recreate trying to create a bunch of presets but they do exist and the code in EffectsGrid>>Paste() handles this situation on Applying the effect.  So figured the preview generation should as well as opposed to just ignoring the preview generation.

### DIFF
--- a/xLights/TabConvert.cpp
+++ b/xLights/TabConvert.cpp
@@ -1203,7 +1203,7 @@ void xLightsFrame::WriteGIFForPreset(const std::string& preset)
                 ModelManager mm(nullptr, this);
                 _presetModel = new MatrixModel(&n, mm, true);
 
-                Element* elem = _presetSequenceElements.AddElement(_presetModel->GetName(), "Model", true, false, false, false);
+                _presetSequenceElements.AddElement(_presetModel->GetName(), "Model", true, false, false, false);
             }
 
             size_t channels = PRESET_ICON_SIZE * PRESET_ICON_SIZE * 3;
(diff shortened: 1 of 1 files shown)


AzGilrock:
I did a fair amount of work on the presets dialog back in 2015.  If I remember right it applies the presets as if they were a paste.  I believe the negative values might be valid just not sure I'm remembering this correctly.  The way paste works is all effects pasted are based off a reference effect and I believe that reference effect is not guaranteed to be on the highest row.  I think it happens when a lower row has an effect earlier in the timeline than the top row of all selected effects.  Then that lower row becomes the reference as 0 and the rows above it have negative values.

PR #2510, opened 2021-02-18
Makefile, improve installing in an alternative location, add TCP_NODELAY

dfries:
I was able to install without sudo by installing it into an alternative location.  The changes to the Makefile make this easier and should not affect the default installation.

xLights playing to the network is clearly timing sensitive.  It would appear xLights will write a complete frame worth of LEDs to a controller in one write, assuming that is correct, adding TCP_NODELAY disables optimizations intended for applications that aren't fully buffering their write.  The LED frame is timing sensitive and needs to go out as soon as it is written.

In practice when I was using tcpdump to look at the packets coming out of xLights, running at 40 Hz even with TCP_NODELAY the packets weren't an even 25ms spacing.  Without TCP_NODELAY I was at times seeing two frames worth of data being sent in one frame.  I didn't look into it further.  My conclusion is TCP_NODELAY helps, and there are likely other timing issues to look into.


Makefile
```diff
@@ -1,5 +1,9 @@
+PREFIX          = /home/david/src/xLights/local
```
computergeek1507, line 1: I don't think it is good to have your user folder hard coded in the makefile

dfries: Thanks for pointing that out, I missed resetting it back to the default.

keithsw1111:
There are a couple of problems here preventing a merge. 1. You have two unrelated changes ... one I would likely merge but the other I wont as i really need Chris to evaluate anything related to the Linux build script. 2. You use .error for a success log ... in fact i am not even sure why we would log that at all. Logging is largely for trouble identification and occasionally progress/activity identification ... i dont think the success log needs to be logged at all here.

dfries:
1. I separated out the Makefile commit into a separate pull request.  I didn't include a README.txt file entry because it is a development time improvement not an end user visible change.

2. If setsockopt succeeds there isn't any output

Thanks for making xLights available, last season was my first Christmas LED light show of any scale, 1300 WS2812B fairy LEDs, 5 Raspberry Pi controllers, and a Linux laptop running xLights.  I started the setup with the intention of some fixed animation sequences, I kept seeing references to xLights, and with xLights around the world I ended up putting together three songs!

PR #2562, opened 2021-03-19
Added One Shader

TurtleThumper:
Added "Acid Jelly"  Tested in xlights prior to submission


dkulp:
This was already handled

PR #2630, opened 2021-05-03
Update shaders.xml

Old-Salt:
Removes all shader_xL.fs from the catalog.  Load the attached shader in xLights for the reason.
[shader removed1.fs.txt](https://github.com/smeighan/xLights/files/6417967/shader.removed1.fs.txt)

### DIFF
--- a/download/shaders.xml
+++ b/download/shaders.xml
@@ -91,6 +91,14 @@
 		<rating>3</rating>
 	</shader>
 	<shader>
+		<name>Aurora</name>
+		<id>604bcf6adf59c70014cdc5ff</id>
+		<link>https://www.interactiveshaderformat.com/sketches/604bcf6adf59c70014cdc5ff</link>
+		<download>https://www.interactiveshaderformat.com/sketches/604bcf6adf59c70014cdc5ff/download</download>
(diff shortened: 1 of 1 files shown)


keithsw1111:
Is there a reason why?

Old-Salt:
Yes.  These shaders have been removed from the ISF repository.  I'm replacing all of them with versions that do not reference xLights in any manner,

The reason for this is that some no longer behave as expected in xLights, and that extra controls which are added within xLights.  In some these are highly inappropriate. 

I'm firmly of the opinion that the changes within xLights are good for some shaders, but not for the vast majority of them.  They should be an opt-in choice, not one imposed by default.

Therefore I'm not going to re-write my shaders to work around the problem.  Going forward, I will not be adding any shaders to the xLights catalog. Others are free to add those I've added to the ISF repository, just like any other shader.  Others are welcome to copy the shaders I put in the ISF repository and modify the as they wish, one again. just like any other open source shader in existence.

This pull request cleans up my mess, as was described during the VCS.

I respect the xLights' development team to take it where ever they wish.  Similarly, I expect the team to respect where I'm taking my shaders.

PR #2652, opened 2021-05-18
A better implementation of Offset and Zoom

Old-Salt:
Does not display second controls for shaders already implementing them for offset and/or zoom.
Offset is detected by presence of a 'point2D' input
Zoom is detected by by a 'float' input containing 'zoom' in its name.


keithsw1111:
I dont agree with this change.

While I get the desire to remove unnecessary redundant parameters doing so makes a big assumption about the duplication which I dont think you can know is true ... and I definitely dont want users to need to write shaders for xLights specifically.

While zoom may be the same it is also entirely possible it is different ... in intent or implementation.

As for Point2D assuming this is for offset is an even bigger assumption as it is usually used to represent a mouse in most shaders I have looked at.

The idea to add a generic zoom and offset was a great idea and to the best of my ability i have implemented it in a way which does not require the shader to know about it while also letting a shader that does know about it to access the unaltered values.

If there is a benefit other than removing the redundant parameters or you can help me understand why they truly are always duplicates then I am happy to reconsider.

Old-Salt:
Offset and Zoom are mathematical transformations.  As such, the order of operations is critical.  Applying zoom, then offsetting the result does one thing, while applying an offset and then zooming has a different result.  Another problem with both offset and zoom transformations is the coordinate system's origin.  Most have the origin at the center, but when they don't - look out.  I learned this the hard way.

Add to these other mathematical transformations such as rotation about one or more axes, and mirroring.  Then throw in camera angles, depth of field, and the positioning of light(s) in 3D shaders to quickly make this an unmanageable situation, even if all xLights does is perform two simple hidden transformations at the start.

You are correct in that a point2D is usually used to represent a mouse.  The mouse is available in both the GLSL Sandbox, and Shader Toy implementations of Open GL.  The implementations are different, mouse buttons are available in one but not the other.  The ISF does not recognize the mouse, so it is usually implemented as a point2D by those performing conversions to this standard.  They often just leave the name alone to ease the conversion process.  Note that the mouse is used in the other implementations of Open GL to initiate and/or control a mathematical transformation.

Checking for the presence of point2D, zoom, or scale is the most reliable method that I could come up with to detect a shader that performs transformations.  As I think about it, I'm leaning towards leaving the coordinate system alone for shaders with any of these inputs. 

I'm in complete agreement that adding zoom and offset control to shaders that don't perform any transformations is a good idea.  I think the expectation that someone walk back the transformations xLights imposes to get a shader to work is unreasonable.

keithsw1111:
So your order of operations point is important ... are you saying I got them wrong ... I remember testing it and feeling the order I settled on was best but I am happy to reconsider.

As for your other points then I think you are arguing for my current implementation unless in the default state it isnt doing nothing which was the intent. Ie if the user does not touch the default settings then I expect the shader to render exactly as it did before.

Old-Salt:
No, they're not wrong - for most shaders.  I have encountered several shaders that rely on applying them in the opposite order.  I've also encountered shaders where  an offset or zoom would be inappropriate.

Recently, I've also added rotation to the mix.  That gives me eight possible combinations of transformations to apply, and then I need to decide on the order of operations.  That order of operations makes a huge difference in the final result.

As an example, go to https://editor.isf.video/shaders/609ae9f3df59c70014cdc6b2
This is a shader that has a lot of different action in it.
Check uSegShort - this allows you to select part of the action using the uSegSel slider.  (I like about 0.9)
Use offset to center the area you like best, and zoom as you wish.
Now add a little rotation, notice that the rotation is around  the center of the Talisman
Use the drop down  "Rotate Center' to select 'Viewport,'  now rotation is around your chosen viewpoint.
 
These are two of the possible outcomes from changing the order in which the transformations were applied.  In this case I thought both were useful, so I made the selection available.

With xLights imposing transformations there are three possible outcomes:
1. Shaders with no transformations will have them added, This is a good outcome 90% of the time.
2. Shaders with the same transformations, applied in the same order receive a duplicate set of controls, some labeled differently.
3. Shaders with the xLights' tranformations may require them applied in the opposite order.  They also may include other transformations  where those imposed by xLights must be applied between them.  These shaders may or may not have a duplicated set of controls, and rarely produce the expected results.

Since there is no method for xLights to determine the order in which the transformations should be applied, in my opinion xLights should leave the coordinate systems for any shader using transformations strictly alone.  'point2D' is an almost certain indication that transformations are used in the shader.  The words 'zoom' or 'scale' in an input's name are also very good indicators.  Offset and zoom are by far the most common transformations.  Shaders with other transformations often also use offset and/or zoom transformations.  While not precise, these are the best I can come up with for the xLights detection of  transformations in a shader.  Those creating shaders for xLights, including 'NoXform' or something similar in the name of a float, or color if it needs to be hidden, would be an easy way to indicate that xLights was not to modify the coordinate system.

This combination would cause the least number of problems with the shaders

PR #2675, opened 2021-06-13
Changed text from .xml to .xsq

bcarriveau:
Changed text in line 240 from .xml to .xsq


derwin12:
Thanks Keith .. Bill, Paul and I were going thru the process..

PR #2692, opened 2021-06-25
Fpp Connect Button Added to Controllers Page

bcarriveau:
Button to open fpp connect on the controllers tab , one button height spaced below the Discover button.

My original thought was to put it after upload output but it isn't really controller specific like the rest of the functions there are, but goes well with the flow of Upload inputs, Uploads outputs, fpp connect.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,7 @@ Issue Tracker is found here: www.github.com/smeighan/xLights/issues
 
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
+   -- enh (bill) Add FPP Connect button to Controllers page
    -- enh (bill) Added/change a few tool tips to help new users 
    -- enh (scott) Added the Ability to Import Views from RGBEffects File and Sequence Files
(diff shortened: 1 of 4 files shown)


bcarriveau:
FPP

keithsw1111:
I am not a huge fan of this change ... not because it is fundamentally a bad idea ... but where does it end ... do you add bulk controller upload and hinkspix export as well? What rules do we have for when we do and do not add buttons here. Right now it is controller related ... which is a rule I think we have followed so far. This makes the rule much greyer. Maybe now it is all hardware related items?

I know FPP is a bit of an exception although I think FPP connect really needs to change name at some point as some of what it does now is not really just about FPP but about pushing sequence data to controllers (some (to be fair most) of which are FPP).

computergeek1507:
I agree with Keith, this is a slippery slope of what is "important" enough to get a button on controller page.

derwin12:
Why is it called FPP Connect when the panel is coded as FPP Upload :-)

PR #2696, opened 2021-06-30
Fix for Double Value Curves Upgrades

### LINKED ISSUE #2695: .20 update causing strange single line speeds
**Describe the bug**
After opening a sequence on 2021.20, I noticed single lines I have on groups are doubling in speed.

I am guessing it was caused by this change. So it will probably effect shockwave effects as well
— enh (scott) Increase range of shockwave and single strand effect

**To Reproduce**
1. See attached sequence (open and render)
2. Note the speed of the effects.
3. To fix it -  if you pick the single line effect on the rosa grande,  just add another color and remove it again,   or remove and re-add the curve again ... now it will be correct.

**Expected behavior**
For the effects to be the same speeds they were in the previous version.

 - OS: Windows 10
 - xLights version 2021.20

**Attachments**
- [https://drive.google.com/file/d/1m1o_RqZJvzROspzOYLoHBIszu8T4eQ0Y/view?usp=sharing](https://drive.google.com/file/d/1m1o_RqZJvzROspzOYLoHBIszu8T4eQ0Y/view?usp=sharing)

computergeek1507:
I am trying to fix #2695. Setting the new max/min then Deserialize the old value, triggers the upgrade code during batch render in my testing. I am unsure if this will break other things. The VC code is very confusing so I hope someone can verify this is correct and test it more.

### DIFF
--- a/xLights/effects/RenderableEffect.cpp
+++ b/xLights/effects/RenderableEffect.cpp
@@ -266,7 +266,7 @@ bool RenderableEffect::SupportsRenderCache(const SettingsMap& settings) const
         {
             return true;
         }
-        
+
         // we want to cache rotations because of compute cost
         if (Contains(it.first, "VALUECURVE_Rotations"))
         {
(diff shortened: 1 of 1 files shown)


keithsw1111:
I dont know scott. Without a bunch of things to test it with this is really hard to be sure about.

PR #2706, opened 2021-07-14
Added xBackup App

computergeek1507:
I wrote this to help users restore their show folder from the backup directory. It will display the Controllers, Models, Sequences in the selected backup folder. WIP looking for feedback.


dkulp:
Why wouldn't this just be a tool on either the Tool menu or even a "Restore from Backup" option under the "Backup" option on the File menu?

computergeek1507:
I didn't know how risky it would be to pull all the files out from under xLights, It would have to unload the RGB Effects file and and networks file without saving.

dkulp:
Those two files are easy as you can just clear the "dirty" flag and then they won't save.  The only one that is really a concern is the keybindings file as I believe that always saves, but that should be fixed to just save when modified in the preferences panel, then that issue goes away.

computergeek1507:
I will look into that, Thanks.

PR #2784, opened 2021-08-31
Spelling Repair

bcarriveau:
fix Aternate


computergeek1507:
You need to write upgrade code for models with the old setting to convert it to the new name

bcarriveau:
show me if easy :-) 👍

PR #2797, opened 2021-09-08
Dialog wording update

MrPaulAR:


keithsw1111:
Happy for you to remove the xml file reference but you cant remove replace as that is what it does.

MrPaulAR:
I couldn't come up with a brief wording that made sense and used the word "replace".   I did say to open the other file instead.

AzGilrock:
I thought the original wording made sense.  The word "replace" at least alerts the user that they will lose something.

PR #2828, opened 2021-09-30
Create Lightish.xcontroller

emil-johansen:
Added a new controller

### DIFF
--- a/controllers/Lightish.xcontroller
+++ b/controllers/Lightish.xcontroller
@@ -1,0 +1,42 @@
+<Vendor Name="Lightish">
+    <AbstractVariant Name="BaseWLEDSettings">
+        <!--SupportsPixelPortCommonSettings/-->
+        <SupportsPixelPortDirection/>
+        <SupportsPixelPortColourOrder/>
+        <SupportsPixelPortNullPixels/>
+        <SupportsUpload/>
+        <MaxInputUniverses>9</MaxInputUniverses>
+        <MaxSerialPortChannels>0</MaxSerialPortChannels>
+        <MaxPixelPortChannels>4500</MaxPixelPortChannels>
+        <AllInputUniversesMustBeSameSize/>
+        <UniversesMustBeSequential/>
+        <SupportsAutoLayout/>
+        <PixelProtocols>
+            <Protocol>ws2811</Protocol>
+            <Protocol>tm18xx</Protocol>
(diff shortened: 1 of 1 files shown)


keithsw1111:
Is this actually any different to WLED? How broadly available is it. A google search doesn't show any obvious mentions.
We dont generally add controllers with no significant market share or distinguishing features from the base.

emil-johansen:
These particular controllers are sold mainly in the EU, thus the lower mentions in the forums. The E1-WiFi is actually quite popular, in primarily Denmark. We are working hard on new versions of controllers (Wled and FPP) with new abilities not seen on the market before, to be available for the European community. Therefore would we like to add the controllers to Xlights, to make it easier for setup, uploading outputs, and so on.

keithsw1111:
Which forums ... do you have a website. You dont say what the unique features of this board (ie in your proposed change) are that are not supported by the current WLED definition.

emil-johansen:
They are mainly discussed in our local forums over here. It's not the E1-WiFI that has any special features to brag about, but the new lineup on the way, fx we have a new controller called E8D-WiFi that is based on WLED with the ability to plug 2 normal differential receivers in. 
The creation for the "new vendor tab" is that we have some FPP-based controllers in the making, and would like to have them in the same place as the WLED controllers - If we are allowed to?

Yes, we do have a website, but we mainly sell to Nothern Europe.

But the main point is the price, the controllers from the US just cost a lot more over here and that is holding some people back from entering this hobby and I think that is sad.

keithsw1111:
Then come back to us when it needs something unique. In the meantime the current WLED configuration appears to meet your needs.

I dont understand why you are not posting a link to the website.

keithsw1111:
For future reference ... I finally found the website he was not willing to provide: https://computerstyretjulelys.dk/vare-kategori/controller/

emil-johansen:
I'm sorry that you got the impression that I didn't want to share the website :(  That wasn't the intention. 

I see why you want it to be added under WLED, and I will do so instead :)

Thank you so much for the help, and in general for supporting/developing Xlights - It's a fantastic program.

jefft0825:
Why is there a picture of a Kulp board on the site?

emil-johansen:
> Why is there a picture of a Kulp board on the site?

Course at the time of making the category the only controller in stock was Kulp. just forgot to change it, thank you for pointing it out:)

Issue #3009, opened 2021-12-14
Remote Falcon reports unable to upload playlist in .39 - rollback to .38 and works fine.


derwin12:
3 Folks and one in zoom .. simply rolling back made it work again. Is there anything obvious in the update that might have caused this or do you need a tester? log?

PR #3010, opened 2021-12-15
Hopefully Fix RF

bcarriveau:
Remove .a from dll


derwin12:
Is this perhaps the fix for #3009 ? If so it can be closed.

PR #3025, opened 2021-12-21
controllers:  Add support for BBB16v2

sslupsky:
Scott Hanson's BBB16v2 has an extra 4 ports available using the diff port.
Ports 17 to 20 do not appear in the xLights visualizer so it is
not possible to assign a model to those ports.

This commit adds a controller definition
that includes the additional ports available
using the diff port.

### DIFF
--- a/controllers/fpp.xcontroller
+++ b/controllers/fpp.xcontroller
@@ -294,7 +294,7 @@
             <fppSerialPort1>ttyS1</fppSerialPort1>
             <fppSerialPort2>ttyS2</fppSerialPort2>
         </Variant>
-        <Variant Name="w/Expansion" ID="BBB16v2" Base="FPP:FPPStringCapeTTYs">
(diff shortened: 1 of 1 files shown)


computergeek1507:
"BBB16v2-EXP" should be the ID of the version with an expansion

sslupsky:
> "BBB16v2-EXP" should be the ID of the version with an expansion

Yeah, I noticed that.  When I was looking at how Dan configures his boards I noticed that he doesn't change the ID between the different variants, just the variant name changes.  Do you want to change the ID as well?

sslupsky:
@computergeek1507 Can we get this merged so it is included in the next release build?

sslupsky:
@computergeek1507 Is there some outstanding issue preventing this from being merged?  I made the change you requested so I think this is ready to be merged.  It's a bit of an annoyance to have to go and manually update the fpp.xcontrollers file every time there is an update.

PR #3029, opened 2021-12-22
wxSpinCtrl:  work around macOS issue

sslupsky:
The wxWidget wxSpinCtrl() checks value in the spinner
against the min and max continuously.  If you attempt
to key a value directly and the min value is greater
than 1, it prevents the user from entering a value.

This commit fixes four instances of this issue in xLights by
changing the min value to 1 and performing min value
checking when wxSpinCtrl returns a value.

### DIFF
--- a/xLights/CustomTimingDialog.cpp
+++ b/xLights/CustomTimingDialog.cpp
@@ -41,7 +41,7 @@ CustomTimingDialog::CustomTimingDialog(wxWindow* parent,wxWindowID id,const wxPo
 	FlexGridSizer1 = new wxFlexGridSizer(0, 3, 0, 0);
 	StaticText1 = new wxStaticText(this, ID_STATICTEXT1, _("Frame interval (msec):"), wxDefaultPosition, wxDefaultSize, 0, _T("ID_STATICTEXT1"));
 	FlexGridSizer1->Add(StaticText1, 1, wxALL|wxALIGN_CENTER_HORIZONTAL|wxALIGN_CENTER_VERTICAL, 5);
-	SpinCtrl_Interval = new wxSpinCtrl(this, ID_SPINCTRL_Interval, _T("33"), wxDefaultPosition, wxDefaultSize, 0, 15, 1000, 33, _T("ID_SPINCTRL_Interval"));
+	SpinCtrl_Interval = new wxSpinCtrl(this, ID_SPINCTRL_Interval, _T("33"), wxDefaultPosition, wxDefaultSize, 0, 1, 1000, 33, _T("ID_SPINCTRL_Interval"));
 	SpinCtrl_Interval->SetValue(_T("33"));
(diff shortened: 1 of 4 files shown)


sslupsky:
@dkulp @keithsw1111 @computergeek1507 Any feedback on this?

AzGilrock:
So you’re setting the min’s to 1 but then have code to force the desired min values right?

dkulp:
I've specifically not merged this as I know there are a bunch of wxNumberValidator and wxSpinCtrl  and wxTextCtrl fixes in wxWidgets/master that I want to test with first.   We don't do wxWidgets upgrades late in the year due to all the other changes in wxWidgets so this is something I hope to look at in Jan/Feb at some point.

dkulp:
It's also likely that:
https://github.com/wxWidgets/wxWidgets/pull/2380
is the real PR we want to grab.

sslupsky:
Yeah, that wxWidgets PR might to the trick but beware some of the other comments in the bug report ...

_ As the 10th birthday of this bug report is getting close, perhaps the simple workaround should be applied ASAP, until a complex solution is found and agreed upon._

sslupsky:
> So you’re setting the min’s to 1 but then have code to force the desired min values right?
> 
> 

Yes.  There were only a small number of instances I found where the min value was set to a value that caused the bug.  So, I changed all it two and did the validation with additional code.  The two instances I did not change were because I was unsure of what was going on with those particular pieces of code so thought some discussion would be helpful.

sslupsky:
I agree with Dan though, it is a wxWidgets bug and if it is fixed there we should grab that.  Only problem is that is it might take forever.

AzGilrock:
You can still use the up/down arrows to set your value right?

sslupsky:
> You can still use the up/down arrows to set your value right?

Not in a practical sense.  Say you want to set the metronome to 500ms ... you would have to click 250 times.


There is one thing that works though.  If you open an editor, enter a value into the editor then cut and paste the value into the SpinCtrl, you can enter a value.

sslupsky:
Progress:
https://www.wxwidgets.org/news/2022/01/goodbye-trac/

sslupsky:
Recent nighties of codeblocks require a more recent version of wxWidgets.  We should consider updating to something after this commit:  https://github.com/wxWidgets/wxWidgets/commit/391080e77d7f78d5d136f30fc5ac02ee89e6ec2e

dkulp:
> Recent nighties of codeblocks require a more recent version of wxWidgets. We should consider updating to something after this commit: [wxWidgets/wxWidgets@391080e](https://github.com/wxWidgets/wxWidgets/commit/391080e77d7f78d5d136f30fc5ac02ee89e6ec2e)

I tried flipping to newer builds a few weeks ago but ran into a bunch of issues (particularly with the bitmap button drawing).   They've fixed those issues, but haven't had time to really do a better/more complete set of testing.   3.2.0 is due soon so we may just wait for that.

dkulp:
Closing.   At this point it's obsolete.

Issue #3018, opened 2021-12-18
Feature Request:  Default Gamma


sslupsky:
**Is your feature request related to a problem? Please describe.**
It would be helpful to be able to have a default setting for Gamma.

**Describe the solution you'd like**
Similar to the "Default Brightness" setting for controllers that support FPP, add a "Default Gamma" setting as well.

**Describe alternatives you've considered**
Manually change Gamma using FPP output settings or model dimming curves.

tmblue:
+1

PR #3041, opened 2021-12-30
controller:  add support for default gamma

sslupsky:
This commit adds a Default Port Gamma setting to the controller
settings dialog of the controller tab.

The setting works similar to the Default Port Brightness.

Address issue #3018 

Note, I do not have a Falcon controller and so I am unable to test the changes against the Falcon controller specific code.  Thus, this commit only supports FPP controllers at the moment.


keithsw1111:
I am not convinced this is actually a useful change. I just cant imagine a common requirement for a default other than 1,0.

PR #3055, opened 2022-01-04
xlDo in MacOS

gbonesso:
Changes to allow build and execution of xlDo in MacOS

### DIFF
--- a/xLights/automation/automation.cpp
+++ b/xLights/automation/automation.cpp
@@ -46,11 +46,11 @@ void xlDo_Output(const std::string& script, const std::string& resp, bool verbos
                     for (const auto& it : val.GetMemberNames()) {
                         std::string set = "@set";
                         #ifndef __WXMSW__
-                            set = "EXPORT";
+                            set = "export";
                         #endif
                         std::string c;
                         if (val[it].IsString()) {
-                            c = wxString::Format("%s %s=%s\n", set, it, val[it].AsString());
+                            c = wxString::Format("%s %s=\"%s\"\n", set, it, val[it].AsString());
                         } else if (val[it].IsLong()) {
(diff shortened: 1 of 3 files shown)


dkulp:
xlDo is not needed for MacOS or Linux.  Call the xLights executable directly with  the -xlDo flag (and then the xlDo options).  With 2021.01 (when available) a symlink of xlDo to the xLights executable will also work directly.

gbonesso:
Hi Daniel,

That's great, I didn't know that. The -xlDo option doesn't appear in the xLights help (/Applications/xLights.app/Contents/MacOS/xLights -h).
Anyway the modification in automation.cpp to generate the script with "export" in lowercase is necessary to run the script in MacOS. The string output in MacOS should be between quotes in the case of spaces in the string.
Now the script generated in MacOS is:

> EXPORT output=/var/folders/_5/9t40z6jd01z9m3f30typtd340000gn/T/xLightsCheckSequenceTSQher.txt
> EXPORT msg=Sequence checked.
> EXPORT res=200

And it should be:

> export output="/var/folders/_5/9t40z6jd01z9m3f30typtd340000gn/T/xLightsCheckSequenceBEeWeR.txt"
> export msg="Sequence checked."
> export res=200

I'm trying also to make MacOS versions of the scripts that works in Windows and are in the git repository. The following script is trying to do the same work as "checkASequence":

> #!/bin/bash
> $xLights_path/xLights.app/Contents/MacOS/xLights -xlDo -v -c "{\"cmd\":\"checkSequence\", \"seq\":\"$1\"}" -s result.sh
> 
> if [ $? -eq 0 ]
> then
>   chmod 755 result.sh
>   source result.sh
>   open -t $output
> else
>   echo "Error level: $?"
> fi

But it works only if the file "result.sh" exists in the directory... Looks like a permission issue, maybe related to calling the xLights inside the .App...

PR #3061, opened 2022-01-09
3048 - xSchedule - cloning a step makes it loop oddly until restart

robfallone:
This addresses: https://github.com/smeighan/xLights/issues/3048

The issue here was that during a clone process the "_id" member was getting copied when in fact it should have been left alone (set in the constructor to be unique). To fix this I created a new clone/copy method that conditionally copies these fields during copy vs. clone.

While fixing this I debated several approaches and settled on this one. However, I'm not married to it and would be happy to change it. I think its more of a preference/style thing. Feel free to ask for changes.

### DIFF
--- a/xSchedule/PlayList/PlayListItem.cpp
+++ b/xSchedule/PlayList/PlayListItem.cpp
@@ -119,11 +119,14 @@ std::string PlayListItem::GetNameNoTime() const
     return "<unnamed>";
 }
 
-void PlayListItem::Copy(PlayListItem* to) const
+void PlayListItem::Copy(PlayListItem* to, const bool isClone) const
 {
-    to->_id = _id;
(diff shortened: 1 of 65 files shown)


xSchedule/PlayList/PlayListItem.cpp
```diff
@@ -119,11 +119,14 @@ std::string PlayListItem::GetNameNoTime() const
     return "<unnamed>";
 }
 
-void PlayListItem::Copy(PlayListItem* to) const
+void PlayListItem::Copy(PlayListItem* to, const bool isClone) const
 {
-    to->_id = _id;
-    to->_lastSavedChangeCount = _lastSavedChangeCount;
-    to->_changeCount = _changeCount;
+    if (!isClone) {
```
robfallone, line 124: This "if" statement is the crucks of the fix.

PR #3064, opened 2022-01-10
3063 - 3D mesh throwing unhandled exception

robfallone:
I found the OBJ file (3D mesh) doesn't have any texture coordinates because it does not have any textures. The code assumes there always exists some textures. Given that I generated this file using FreeCAD and did not modify it, I'm inclined to believe that having a 3D mesh without textures is a valid thing. If so, I believe this PR addresses the issue.


robfallone:
I screwed up the PR. I will open up a new one with correct commits.

PR #3068, opened 2022-01-10
Add support for Twinkly lights

aarcangeli:
Hello xLights Team.

I developed a feature (initially for me) which permits to control Twinkly LED strings using XLights.

To do this I used the [unofficial api](https://xled-docs.readthedocs.io/en/latest/rest_api.html).

The api is relatively simple, there are some http call to change the mode and UDP packets to send the lights data in real time to the device.

## Brief

Twinkly string is an app controlled chain of RGB or RGB+W leds, each pixel can be controlled individually.
The android application can be used to map each pixel to the corresponding location.

Luckly the mapping is stored on the device, so in can be retrieved with api.

Homepage: https://www.twinkly.com/

## Feature details

This PR adds a new protocol type and a layout type.

### Controller tab

<details>
<summary>Screenshots</summary>

image

</details>

The controller can be added as usual with the "Add Ethernet" button and choosing "Twinkly" in the protocol field.

The IP must be a network reachable twinkly controller.

This configuration can be done with the official application in station or access point mode.

The "Get layout" button (which is showed only if a twinkly controller is selected) downloads the mapping from the device and creates a new layout for it.
This is the only way to create this kind of layout.

### Layout tab

<details>
<summary>Screenshots</summary>

image

image
</details>

The layout tab can be used to move and resize the model.

## Known bugs:
- "Get Layout" is not pushed on undo history (I don't know how to do it)
- If the ip is wrong, the connection times out after 5 seconds (which is quite boring)
- In the layout editor, the red arrow of the "top" handle doesn't take any effect (the green handle changes the height of the model)
<details>

image
</details>

### DIFF
--- a/xLights/TabSetup.cpp
+++ b/xLights/TabSetup.cpp
@@ -58,6 +58,9 @@
 #include "../xFade/wxLED.h"
 
 #include <log4cpp/Category.hh>
+#include <outputs/TwinklyOutput.h>
+#include <models/TwinklyModel.h>
+#include <LayoutPanel.h>
 
 const long xLightsFrame::ID_List_Controllers = wxNewId();
 const long xLightsFrame::ID_NETWORK_ADDETHERNET = wxNewId();
@@ -1576,6 +1579,8 @@ void xLightsFrame::SetControllersProperties() {
         ButtonOpen->Enable(false);
         ButtonControllerDelete->Enable(false);
         LedPing->Disable();
+        ButtonControllerDownloadTwinklyLayout->Hide();
+        FlexGridSizerSetupControllerButtons->Layout();
 
         if (_outputManager.GetGlobalFPPProxy() != "") {
             Button_OpenProxy->Enable();
@@ -1675,6 +1680,13 @@ void xLightsFrame::SetControllersProperties() {
 
             // one item selected - display selected controller properties
             controller->AddProperties(Controllers_PropertyEditor, &AllModels, expandProperties);
+
+            if (controller->GetProtocol() == OUTPUT_TWINKLY) {
+                ButtonControllerDownloadTwinklyLayout->Show();
+            } else {
+                ButtonControllerDownloadTwinklyLayout->Hide();
+            }
+            FlexGridSizerSetupControllerButtons->Layout();
         }
     }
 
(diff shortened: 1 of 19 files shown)


dkulp:
I'm OK with the output code (although using curl would definitely be an upgrade, does twinkly support keepalives?), but I'm a bit against having a separate Twinkly model.     I would greatly prefer an update/button on the Custom model dialog to support loading data via twinkly and just use the Custom model.

sslupsky:
If the lights are ws2811, why even bother with supporting their non standard controller?  Just connect the lights to a Kulp/Falcon/etc controller?

robfallone:
Having support for Twinkly would be a very nice add. I agree with some of @dkulp comments. However, I think this is a great start and I think we should continue to pursue this and improve it. Regarding not supporting their controller, my opinion is that we support many controllers (not sure what "standard" means) and why not add this to the list of supported controllers.

As a side note, I have a neighbor who has invested in Twinkly and we have a talked about ways to synchronize our shows.

sslupsky:
By "standard" I am referring to is FPP.  Though not strictly necessary, I think it is accurate that most pixel based displays use FPP.  xLights is able to interface with controllers directly so it is possible to support another controller protocol within xLights.  It is not clear to me if the Twinkly API is compatible with xLights.  My understanding is models are independent of the controller.  Dan pointed out the proposed solution requires using a "Twinkly model" which appears to tie the model to the controller.

I agree though this is a good thing to pursue if it can be done in a manner that allows all users of xLights to benefit.  If we can seamlessly integrate Twinkly pixels into xLights and FPP (and possibly the controller) then the consequence for the existing suppliers of pixels will be profound.  The existing suppliers will be forced to compete with pixels that are in several retail channels.  Moreover, Twinkly does appear to have some degree of safety and quality control that the existing suppliers just don't have (though they are self certified).

aarcangeli:
Lights are connected to the controller with only two wires, so I don't think they are based on ws1811.
Also, lights cannot be unplugged from the device (unless you cut the wires).

The only supported way to control the lights is with the official app, but the effects are limited and it cannot be integrated with external applications.

This implementation uses an unofficial API.

I agree that it is preferred to use an already existing model for the layout, I'm not sure if the custom model can be used to store the position of 600 lights one by one.

sslupsky:
> I'm not sure if the custom model can be used to store the position of 600+ lights one by one.
> 

I think you can do that by selecting the "model data" in the custom model.

AzGilrock:
I’ll probably work on a true 3D custom model that will work for this

keithsw1111:
I am about to merge the twinkly output code with some enhancements. I have moved all the wxhttp and the roll your own curl code to our existing curl functions which do everything required and I have reverse engineered the discovery protocol and so added that as well. So when you click the discover button it should find the twinkly controllers.

PR #3171, opened 2022-03-31
Fix the Linux build (update .cbp file)

KMojek:


KMojek:
I didn't see those, no. I'll clean it up if this is adding duplicate code. Yes, I do plan on moving the dialog content to effect assist but I don't anticipate that affecting the effect itself.

AzGilrock:
I took a quick look earlier but couldn’t tell if the existing class would work or had too much extra specific stuff.   I know it had features to support having a matrix to do the scaling, rotation, etc.  Its what I used to do the Bézier curves for the PolyLine model.

PR #3187, opened 2022-04-06
Sketch Effect - allow paths to be dragged in the SketchCanvasPanel

KMojek:


AzGilrock:
I'm going to need to utilize all this path drawing work since I'm going to have a Moving Head effect and one feature will be the ability to draw a motion path.  So if you do anymore work keep in mind all the path drawing and manipulation would be best if it was in a common file both effects could utilize.

KMojek:
>  all the path drawing and manipulation would be best if it was in a common file both effects could utilize

Right now the SketchCanvasPanel and SketchAssistPanel are pretty tightly coupled... it's only a slight improvement over when everything was in one giant SketchPathDialog class. I can clean that up though, but not right now.

PR #3194, opened 2022-04-08
SketchAssistPanel - add 'continue path' button

KMojek:


AzGilrock:
One enhancement I'd suggest is the ability to be able to toggle existing segments from lines to curves.  Unless I missed it I think you can only change the type while your first drawing it.

KMojek:
Yup, that would likely be next up on a list of possible enhancements. This change just makes it so you don't have to trace out the entire path only to find out you made a mistake early on. I may take a break from it for a little while until it makes it into a release and see whatever other feedback comes in... assuming it's a worthwhile effect, that is.

PR #3208, opened 2022-04-18
Fix crash handling

robfallone:
This PR fixes and consolidates all our crash handling. The fixes include the following:

1. Building and reporting an accurate call stack reliably.
2. Handling crashes/exceptions for spwaned threads (i.e. JobPoolWorker, SequenceData, DoPostStartupCommands).
3. Handling SEH exceptions on Windows.

Associated MacOS PR is: https://github.com/xLightsSequencer/xLights-macOS/pull/4


common/xlBaseApp.h
```diff
@@ -0,0 +1,136 @@
+#pragma once
```
robfallone, line 1: This file defines new base classes for all apps to inherit crash handling from.

common/xlStackWalker.h
```diff
@@ -0,0 +1,92 @@
+#pragma once
```
robfallone, line 1: This file is a better version of MSWStackWalk.h (which was deleted). This file has some fixes for handling exception cases and uses wxWidgets where possible.

xLights/JobPool.cpp
```diff
@@ -68,12 +69,22 @@ class JobPoolWorker
 };
 
 static void startFunc(JobPoolWorker *jpw) {
+    try
+    {
+        xlCrashHandler::SetupCrashHandlerForNonWxThread();
```
robfallone, line 74: This is a critical fix to handle exceptions within spwaned threads.

Issue #3211, opened 2022-04-20
Default port brightness missing on f16v4


computerdave200:
In controller settings the default port brightness and gamma settings on the f16v4 is missing in 2022.08. I jumped on zoom room and verified with those online this evening. 

_
Version .07
<img width="687" alt="2022 07 brightness control visible" src="https://user-images.githubusercontent.com/51300984/164131123-a67bf9b0-fac7-4e9c-ba51-6cb15aa0c668.png">

version .08
<img width="698" alt="2022 08 default port brightness missing" src="https://user-images.githubusercontent.com/51300984/164131124-05142a50-2109-4a7c-8d7a-eea570c01622.png">

computergeek1507:
fixed with 548c02cb02e68307d82e152c4a1e2fc5f8fe4b60

PR #3213, opened 2022-04-20
Update falcon.xcontroller

TrekErs:


TrekErs:
fix #3211

computergeek1507:
I don't think the V4 should inherit the V2\V3 base configuration. The V4 supports a different feature set and different pixel types

TrekErs:
I wasn’t sure either so my first fix was the same way you fixed it. Keith happened to pop into the zoom room and confirmed that it should incorporate the base settings.  Of course he didn’t put it quite so delicately.  :)

computergeek1507:
If he wants to incorporate the base falcon type, then the v3 settings need to be abstracted out to another layer. The base settings currently has the GECE pixel types and the universe/channel count of the V3.

TrekErs:
I did a little testing and including the BaseFalconSettings appears to work since the protocols and max input universes are defined in the BaseV4FalconSettings those appear to overwrite the BaseFalconSettings.  Not saying one way is right or wrong, just wanted to share what I found.

While testing I did notice that the protocols with spaces (lines 190 192 195) do not appear as available protocols.  But this is something I do not know how to fix or even if it is an issue.

Issue #3258, opened 2022-05-24
Tree Model with Horizontal Strands


merryoncherry:
**Is your feature request related to a problem? Please describe.**
I wrapped a matrix net around a tree, like a vertical cylinder.  The current xLights tree model offers the correct physical shape, but I did horizontal wiring on the prop, and the tree model currently only supports vertical.

**Describe the solution you'd like**
I would like a checkbox on the tree model property grid to allow horizontal wiring, otherwise the behavior should be the same (much like it is for the underlying matrix model).

**Describe alternatives you've considered**
For 2021, I just used the matrix flat, without the 3D cylinder behavior.  A custom model can also work here, though using xLights to make the model is the most convenient way.

**Additional context**
I have a longer description with pictures here:
https://merryoncherry.org/2022/05/23/using-tree-model-for-a-cylindrical-matrix/

I attempted to implement this on a fork/branch here:
https://github.com/merryoncherry/xLights_arch_tree_models/tree/tree_horizontal

It seemed to work, but I am not sure what to do to make sure it is thoroughly tested, nor do I know how one updates the manual.  If the developers agree that this would be useful for xLights, I will be happy to make a PR, address comments, see if I can update the manual and tests, etc.  If the developers think this is a one-off, I will just export my "tree" as a custom model from the hacked-up xLights, enjoy having tinkered with it, and drop the matter, no worries.

merryoncherry:
Will have to get better at rebasing PRs.  In lieu of that, attaching patch since this is a very simple one.

`diff --git a/xLights/models/MatrixModel.h b/xLights/models/MatrixModel.h
index 72ffd36c1..592668338 100644
--- a/xLights/models/MatrixModel.h
+++ b/xLights/models/MatrixModel.h
@@ -19,7 +19,7 @@ class MatrixModel : public ModelWithScreenLocation<BoxedScreenLocation>
         virtual ~MatrixModel();
         virtual int GetNumStrands() const override;
 
-        virtual bool StrandsZigZagOnString() const override { return true;};
+        virtual bool StrandsZigZagOnString() const override { return true; }
 
         bool SupportsChangingStringCount() const override{ return true; };
         bool ChangeStringCount(long count, std::string& message) override;
diff --git a/xLights/models/TreeModel.cpp b/xLights/models/TreeModel.cpp
index 7f12212ed..8d4740587 100644
--- a/xLights/models/TreeModel.cpp
+++ b/xLights/models/TreeModel.cpp
[...]

PR #3259, opened 2022-05-24
Tree horizontal

merryoncherry:
For #3258 - Tree Model gets horizontal strands


AzGilrock:
I kinda don't like the new attribute being used as a string everywhere requiring string comparisons.  The Matrix used a string but only because it was the "DisplayAs" field which was already a string.  Most other attributes use integers to store the value in the XML.  See how the TOP_BOT_LEFT_RIGHT variable in the MatrixModel is used.  The strings are only used for display purposes but the rest of the program uses the enumeration values and its a wxEnumProperty.  Its not a super big deal I'm just always thinking of what's faster and strings are usually slower.

merryoncherry:
> I kinda don't like the new attribute being used as a string everywhere requiring string comparisons.

I would be happy to follow the other example you cite and switch to integer.  Let me try it this afternoon.  Gotta say, I was surprised how much information is sometimes stored in the "Display As" field, like matrix direction for matrix, degrees for tree...

The argument for using the string in the xml is for any other tool that reads the xml, if any, or if someone reorders the enum, or anything like that, then it will still read properly.

AzGilrock:
Yes I thought of that as well but we probably have tons of items that are tied to enum orders already.  I'm used to working with real time embedded code.  I'd wait and see if any other dev's have an opinion.  I'm not going to make a fuss about it if it stays as is.

computergeek1507:
It's also not a "hard rule" I am sure you can find instances when I have used string compares.

computergeek1507:
all the examples I found we convert bools to strings for saving

merryoncherry:
Seems the XML document uses string for everything... but we still get to decide if we want "0"/"1" or "Horizontal"/"Vertical".

In favor of text (possibly others, this is just what I see doing a quick skim of rgbeffects.xml):
layout="horizontal", type="ranges", bufferstyle="Default", type="subbuffer", StartSide="T", Dir="R", StringType="RGB Nodes", StarStartLocation="Bottom Ctr-CCW", CandyCaneReverse="false"

In favor of numbers (possibly others, as noted):
Antialias="1" (comes from "Pixel Style" enum)
Locked="1", selected="0"

Antialias jumps out as the most elaborate enum I found.  The numbers version includes a lot of code like this, rather than string compares, but it is easy enough to do.
 wxString::Format(wxT("%i"), (int)_pixelStyle)

Whatever is consensus, let me know.

PR #3294, opened 2022-06-23
Update ArtNetOutput.cpp

JanWull:
Updated ArtNetOutput.cpp to set the source port to 0x1936 for ArtNet data.  This is required for some devices, and required by the ArtNet spec.

### DIFF
--- a/xLights/outputs/ArtNetOutput.cpp
+++ b/xLights/outputs/ArtNetOutput.cpp
@@ -41,7 +41,8 @@ void ArtNetOutput::OpenDatagram() {
     else {
         localaddr.Hostname(GetForceLocalIPToUse());
     }
-
+    localaddr.Service(ARTNET_PORT); //force the source port to 0x1936.  Required as per ArtNet specs and required by some devices.
+    
     _datagram = new wxDatagramSocket(localaddr, wxSOCKET_NOWAIT);
     if (_datagram == nullptr) {
         logger_base.error("Error initialising Artnet datagram for %s %d:%d:%d. %s", (const char*)_ip.c_str(), GetArtNetNet(), GetArtNetSubnet(), GetArtNetUniverse(), (const char*)localaddr.IPAddress().c_str());
--- a/xLights/outputs/ArtNetOutput.cpp
+++ b/xLights/outputs/ArtNetOutput.cpp
@@ -43,7 +43,7 @@ void ArtNetOutput::OpenDatagram() {
     }
     localaddr.Service(ARTNET_PORT); //force the source port to 0x1936.  Required as per ArtNet specs and required by some devices.
     
-    _datagram = new wxDatagramSocket(localaddr, wxSOCKET_NOWAIT);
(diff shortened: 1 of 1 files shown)


keithsw1111:
Exactly where in the specs is this required. I am afraid I dont believe it is. If this were true then only one piece of software on a machine could send artnet at any time and there is no reason why that needs to be true. More likely it is a defect in the devices firmware. This change would limit a whole bunch of valid use cases so I cannot merge it.

JanWull:
From the spec found at https://art-net.org.uk/resources/art-net-specification/ on page 10
port

Secondly, an error is produced when using the packet analyzer in DMX-Workshop while xLights is running:
artnet2

I do agree that it is odd that a particular source address is required, but maybe the implementers of the ArtNet protocol had a reason for that.  I believe the proper way to implement ArtNet support in a piece of software is through the ArtNet driver (ArtNetAcn.dll) included in the SDK.  This driver would negotiate between different processes using ArtNet on a single machine.  I understand that implementing this in xLights would take some work.  I may have time for this during the summer.  I have in the past (5+ years ago) written my own basic light control software using this ArtNet driver so I do have some experience with this.  

I understand now that my change proposed yesterday is a quick/dirty solution and not the real answer.   But I do think that xLights needs a change in order to properly support the ArtNet protocol because it will fail to work with devices that follow the ArtNet spec religiously.

Cheers,
Jan

AzGilrock:
I searched for a spec yesterday also because I didn’t believe it either but I did find that port number listed as a requirement in the spec.

dkulp:
Three years ago when this was also reported to FPP I was a bit surprised to see it in the spec.  It's even more of an issue with something like FPP where we need to accept ArtNet packets as well as send them as using the SO_REUSE doesn't work quite as simply as the incoming packets may come in on sockets not ready for reading.    Was definitely a pain to implement.

keithsw1111:
The standard is not very well written.

The statement you are referring to is: 

"Protocol Operation
A Node operates in one mode, each Node having a unique IP address derived from its
Ethernet MAC address. The UDP port used as both source and destination is 0x1936."

So I looked up the terminology and a node is defined on page 8 as:

"Node: A device that translates DMX512 to or from Art-Net is referred to as a Node"

So clearly a controller receiving artnet is a node and should follow this rule (although I doubt they all do). What is much less clear is is a computer sending artnet a node. It may be.

If you look up the definition of a port you get on page 9

"Port: Actual data transmission on Art-Net uses the UDP protocol that operates ‘on top
of’ the TCP/IP protocol. UDP data transfer operates by transferring data from a specific
IP:Port on a Node or Controller to a second specific IP:Port on a second Node or
Controller. Art-Net uses only one port of 0x1936. "

So now you need to look up a controller ... not to be confused with what xLights calls a controller ...

"Controller: A generic term describing an Art-Net device with the primary task of
generating control data. For example, a lighting console."

So clearly this includes consoles but is silent on computers ... but it seems reasonable to assume computers = consoles for this purpose.

That said this still makes no sense. I have had many situations where I have had multiple programs running concurrently on a computer sending out artnet packets and there seems no good reason for a remote device to reject them based on the source port. The best I could agree to is to make it an option on the artnet output to force the source port to be 0x1936 for the handful of devices that implement these checks and leave the full flexibility for others.

PR #3337, opened 2022-07-13
Fix issue 3329 and added enhancement 3332

ruskin0611:


merryoncherry:
I was wondering if we are supposed to delete any PixelData we remove from the container.  (This is done inconsistently.)
I also wonder what should happen if the start channel for "Set pixel ranges" matches but the length does not match existing - it seems the behavior here could be quite confusing.
It would be fine with me to save these thoughts for a future PR.

ruskin0611:
> I was wondering if we are supposed to delete any PixelData we remove from the container. (This is done inconsistently.) I also wonder what should happen if the start channel for "Set pixel ranges" matches but the length does not match existing - it seems the behavior here could be quite confusing. It would be fine with me to save these thoughts for a future PR.

Yeah, I was wondering that as well.
Basically I refactored the code so as not to change any existing functionality.

BUT: yeah, there may be an issue with multiple requests with the same start channel.
I've updated the commit to only include a break if the channel is zero OR the channel counts match. Otherwise the break outside of those conditions means it only checks the first overlay for a matching channel count (and excludes any additional overlays with the same start channel but different channel counts). 

Thus if you started at channel 1 for 600 channels at red. Then started at channel 1 for 300 channels at blue, then half would be blue and the other half would be red and you would have two overlays both starting at channel 1 (this was existing functionality).
The problem with this, is if you have start channel 1 with a channel count of zero, which overlay would it remove (likely the first one even though you'd probably want it to remove the last one). This is because the break statement would prevent it removing both overlays with the same start channel.

merryoncherry:
The iterations aren't set up to continue after removal; leaving the existing behavior is probably fine until we have good reason to change it.

ruskin0611:
> The iterations aren't set up to continue after removal; leaving the existing behavior is probably fine until we have good reason to change it.

My thoughts as well.
We don't want to impact existing behaviour as others may be using this feature.

PR #3403, opened 2022-08-31
Shorten CAD labels for close props. Standardized # Strings label and …

cybercop23:
…added Ctrl Port to make it more obvious (although it is in the hint message lots of guest on Zoom still don't get it)

### DIFF
--- a/xLights/cad/CADModel.cpp
+++ b/xLights/cad/CADModel.cpp
@@ -48,11 +48,11 @@ void CADModel::LoadModel(Model* model, bool twoD, bool addNodeNumbers, bool addC
 
 	if (addConnections) {
 		if (!model->GetControllerName().empty()) {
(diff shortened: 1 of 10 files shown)


computergeek1507:
You changed "Pixel Port" to "Port" but not "Serial Port" to "Port". It should be consistent. I don't agree with the "Ctrl. Ports" with the strings, while this is true with standard pixels, it is not always true with single color/AC lights.

AzGilrock:
Every time I've changed text to try to reduce questions it never seems to solve the problem.  Some people are just always going to need help.  # Strings doesn't really need to match the number of ports unless you care about auto configuring your controllers.

cybercop23:
Yes, we're trying to get everyone new (and seasoned) to move to auto layout unless there's a good reason why not and then we try to use shadow models.

PR #3429, opened 2022-09-16
Change default port brightness

cybercop23:
Lots of people do not know about this option a \nd leave it to 100. We always tell them to change it when we help them set up in the Zoom room.
They also change all of the modeles to 30/40 vs just this option.

### DIFF
--- a/scripts/BatchRenderExportModel.lua
+++ b/scripts/BatchRenderExportModel.lua
@@ -1,0 +1,34 @@
+models_str = RunCommand('getModels', {})
+models = models_str['models']
+
+sel_model = PromptSelection(models,'Select Model')
+Log(sel_model)
+
+seqs = PromptSequences()
+for i,seq in ipairs(seqs) do 
+    properties = {}
+    properties['seq'] = seq
+	properties['promptIssues'] = 'false'
+    result = RunCommand('openSequence', properties)
+
+    properties = {}
+	properties['highdef'] = 'true'	
+	RunCommand('renderAll', properties)
+
+    properties = {}
+	properties['model'] = sel_model
+	properties['format'] = 'avicompressed'
+	
+	media = string.gsub(result['media'],".*\\","")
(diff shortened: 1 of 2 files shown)


keithsw1111:
I am not sure I agree with defaulting away from 100%. Every controller I am aware of default their brightness to 100% or have default pixel counts that allow for this without overloading their outputs.

cybercop23:
Ok, agree. However, everytime we help someone with xLights and FPP in the Zoom room, we start reviewing the controller setup, and always suggest (and always is accepted) to lower the default to 30. People do not know about this and how this works and they set all their props to 30% when they only have to do this.
I know it would be hard to determine, but I'm sure the vast majority does soemthing <100 by either using this option or setting it at the model level.

AzGilrock:
There’s usually a big group of folks that argue 100% is the only way to go.  Not me but I’ve seen the arguments

cybercop23:
Hey guys. Is there a way to make this data driven? Do you have any xLights shared usage logs that can show how users are setting up controllers (I wouldn't be opposed to have share that) or maybe a way to see/get it from FPP sharing stats.

bcarriveau:
I feel most users in the zoom room that need help from the ground up are a
small percentage of the xLight users that dont understand the brightness's
or setup anyways, its why they are helped and setup the way they are 😇 🙈.

Padre653:
Could we not leave the field blank when the controller is added and have it as a required field like controller model and IP address?

cybercop23:
> What about leaving the field blank when the controller is added and have it as a required field like controller model and IP address?

I think that's a great idea given that we don't have any other info to support setting it to a specific value.

And since this just for new/novice users that don't know any better, it should help a lot.

dartunder (changes requested):
Can we close this request?  There are conflicts and I don't think it's good to default away from 100%. As brightness can be controlled multiple ways ( controller settings, dimming curves, fpp plugins, schedule settings, etc ) you can't mix them, or you get 30% of 30%.

PR #3563, opened 2022-11-14
Color Wash Effect update

KMojek:
Addresses https://github.com/smeighan/xLights/issues/3517 by adding a reverse option for the horizontal and vertical fades. I'm not 100% sure this is how Clyde was intending it to work but this seemed to me a reasonable version of what a reverse option should do. Short demo video here: https://www.screencast.com/t/vl19Aqx1oSE

### DIFF
--- a/xLights/effects/ColorWashEffect.cpp
+++ b/xLights/effects/ColorWashEffect.cpp
@@ -8,9 +8,6 @@
  * License: https://github.com/smeighan/xLights/blob/master/License.txt
  **************************************************************/
 
-#include <wx/checkbox.h>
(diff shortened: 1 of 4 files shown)


computergeek1507:
I don't see the benefits of removing the wxsmith auto generated stuff. Now we have to add everything manually and now it's inconsistent with all the other effect panels

KMojek:
> I don't see the benefits of removing the wxsmith auto generated stuff.

Fair enough; I'll revert that. My misguided understanding was that we were trying to move away from that since it's kind of clunky to install a separate IDE just for the GUI-builder component.

PR #3570, opened 2022-11-18
Add Dig-Octa-Brainboard-32-8L

bcarriveau:
Add Dig-Octa-Brainboard-32-8L


bcarriveau:
With Quinn always making new boards im not sure if i should have made this a variant like i did or just a controller?

bcarriveau:
Something is not right with github or my client , i pushed, and the Varant indent was wrong in the preview here, so i edited it directly on github but now shows wrong on Pc. I will start over and just push once.
2022-11-17 21_52_31-GitKraken

PR #3668, opened 2023-01-05
Addition of DynamicPixels Controller configs

OnlineDynamic:
Addition of new config file for my DynamicPixels D24 controller

### DIFF
--- a/controllers/dynamicpixels.xcontroller
+++ b/controllers/dynamicpixels.xcontroller
@@ -49,4 +49,48 @@
             <SupportsLEDPanelMatrix/>
 		</Variant>
 	</Controller>
+		<AbstractVariant Name="BaseWLEDSettings">
+        <!--SupportsPixelPortCommonSettings/-->
+        <SupportsPixelPortDirection/>
+        <SupportsPixelPortColourOrder/>
+        <SupportsPixelPortNullPixels/>
+        <SupportsUpload/>
+        <MaxInputUniverses>9</MaxInputUniverses>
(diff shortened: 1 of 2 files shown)


keithsw1111:
We dont include every new controller vendor in xlights. You have to have some level of market share first then we will consider it. A search on your name and xlights right now returns absolutely nothing on google. In the meantime you can distribute your own xcontroller files to your customers particularly as these are just variants on other existing controllers.

OnlineDynamic:
Can you please change the logic of the controller setups then to allow the controller config to define the base software controller in the .xcontroller files - I had to hard code the model base controller as your current logic assumes its fpp for all controllers where you haven't recognised the vendor in whats hard coded - but one of mine is wled based.  

If you are searching for me I'm ElfStuart on FPP developers vendor forum, OnlineDynamic in Github.

OnlineDynamic:
@keithsw1111 if it helps its the code at line 94 in BaseController.cpp that is causing me issue for my WLED based controller - as my gpio pinout is different from the default WLED controller so I need my own .xcontroller definition file but this code defaults to thinking its fpp based:

`        logger_base.warn("Vendor not recognized ... assuming it is a FPP based vendor : %s.", (const char*)vendor.c_str());
        bc = new FPP(ip, proxy, caps->GetModel());`

OnlineDynamic:
@dkulp - just seen your commit the following day to fix this to have a 'ConfigDriver' driving logic rather than vendor - much appreciated ;)

Issue #3696, opened 2023-01-28
Need variables exposed to Shader


ClutchplateDude:
When writing a shader, I'd like to be able to support a slider like `Cycles` or `Revolutions`, so that it's easier to have the effect do loops instead of copy/pasting multiple shader nodes.

Looking at the code, I'm not sure what units the `Effect->startTime` is, but since it's just an `int` I assume it's frame based. If my reading of `FRAMEINDEX` is correct (and it's the frame number in the effect, starting at zero), this could work if you just exposed `DURATION` (in frames).

Alternatively, I'd like the start time (`TIMESTART`) and end time (`TIMEEND`), or start (`TIMESTART`) and duration (`TIMEDURATION`) of the effect being rendered to be available in the shader, so I can map the `TIME` to a 0..1 range. This would require that `TIME` and these new variables be in the same time frame (i.e. milliseconds from sequence start).

If the number of exposed variables is a concern, you could add a `TIMELOCAL` variable that goes from 0...1 during the effect playback.

PR #3700, opened 2023-01-30
Shader enhancement

ClutchplateDude:
This PR addresses Issue #3696 and makes the Effect Duration available as `XL_DURATION` to shaders. Shaders that would like to offer parameters based on repetitions during the effect currently have no means to do so. Some built-in effects offer parameters like `Cycle Count` or `Rev's`, which repeat things during the effect duration (independent of the actual time). With this PR, shader authors can also offer this and normalize `TIME` to a `0...1` range and create effects that span the duration, no matter how long it actually is.

### DIFF
--- a/xLights/effects/ShaderEffect.cpp
+++ b/xLights/effects/ShaderEffect.cpp
@@ -152,6 +152,7 @@ namespace
         "uniform vec2 RENDERSIZE;\n"
         "uniform vec2 XL_OFFSET;\n"
         "uniform float XL_ZOOM;\n"
+        "uniform float XL_DURATION;\n"
         "in vec2 vpos;\n"
         "in vec2 tpos;\n"
         "out vec2 texCoord;\n"
@@ -1138,6 +1139,9 @@ void ShaderEffect::Render(Effect* eff, const SettingsMap& SettingsMap, RenderBuf
     if (!si->SetUniform1f("XL_ZOOM", zoom)) {
         logger_base.warn("Unable to bind to XL_ZOOM");
     }
+    if (!si->SetUniform1f("XL_DURATION", (GLfloat)((buffer.GetEndTimeMS() - buffer.GetStartTimeMS()) / 1000.0))) {
+            logger_base.warn("Unable to bind to XL_DURATION");
+    }
     if (!si->SetUniform1f("TIME", (GLfloat)(_timeMS) / 1000.0)) {
         if (buffer.curPeriod == buffer.curEffStartPer && _shaderConfig->HasTime()) {
             logger_base.warn("Unable to bind to TIME\n%s", (const char*)_shaderConfig->GetCode().c_str());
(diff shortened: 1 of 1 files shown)


merryoncherry:
That's pretty cool.  Could you provide a shader that uses the parameter for testing purposes?

ClutchplateDude:
Sure, this is shader I've been developing and hope to add to the catalog if this PR gets approved. It does a Shockwave-like effect in a linear direction. I often use this kind of effect on straight pixel runs (like house outlines), this given much better control over that specific use case than Shockwave.
```
/*{
	"DESCRIPTION": "Bounces a gradient between upper and lower boundary, optionally as a dual",
	"CREDIT": "by Clutchplate",
	"ISFVSN": "2.0",
	"CATEGORIES": [
		"TEST-GLSL"
	],
	"INPUTS": [
		{
			"LABEL": "Color",
			"NAME": "color",
			"TYPE": "color",
			"DEFAULT": [0.5,0.5,0.5]
		},
		{
			"LABEL": "Thickness",
			"NAME": "thickness",
			"TYPE": "float",
			"DEFAULT": 0.2,
			"MIN": 0.01,
			"MAX": 5.0
		},
		{
			"LABEL": "Rotation",
			"NAME": "angle",
			"TYPE": "float",
			"DEFAULT": 0.0,
			"MIN": 0.0,
			"MAX": 360.0
		},
		{
			"LABEL": "Dual",
			"NAME": "dual",
			"TYPE": "bool",
			"DEFAULT": false
		},
		{
			"LABEL": "Cycle Count",
			"NAME": "cycles",
			"TYPE": "float",
			"DEFAULT": 1.0,
			"MIN": -16.0,
			"MAX": 16.0
		}
	]
}*/

float frac(float f)
{
    // The epsilon is to prevent the final frame to revert to the beginning value
    return f<0.0 ? 1.0-(-f-floor(-f+0.000001)) : (f-floor(f-0.000001));
}

// loc is -0.5 to 0.5
float mixcolor(float loc){
    float normTimeInEffect = frac(TIME * cycles / XL_DURATION) - 0.5;
    // normTimeInEffect maps to -0.5 to 0.5 now

    float midc = (1.0 - 2.0 * abs((loc - normTimeInEffect) * (1.0 / thickness)));
    return midc;
}

void main()	{
    float p2x = isf_FragNormCoord.x-0.5;
    float p2y = isf_FragNormCoord.y-0.5;
    float rads = angle*3.1415927/180.0;
    float px = p2x*sin(rads)+p2y*cos(rads);
    float py = p2x*cos(rads)-p2y*sin(rads);
    // px,py are in -0.5 to 0.5 range
    
    float mult = mixcolor(py);
    float mult2 = mixcolor(-py);
    vec4 col1  = vec4(mult * color[0],mult * color[1],mult * color[2],1.0);
    vec4 col2 = vec4(mult2 * color[0],mult2 * color[1],mult2 * color[2],1.0);
    gl_FragColor=dual ? 
      vec4(
        max(col1[0],col2[0]),
        max(col1[1],col2[1]),
        max(col1[2],col2[2]),
        1.0
      )
      :
      col1;
}

```

Issue #3712, opened 2023-02-04
Icons in Import Effects - Map Channels - Available are off a bit


merryoncherry:
**Describe the bug**
The icons in the Import Effects... Map Channels... dialog, on the right, under "Available", are off a bit.
(See screen shots)

**To Reproduce**
1. Create a new sequence.
2.  Import from package zip, making sure that the zip being imported contains a number of interesting models (star, candy cane, etc.)
3. Look at the icons on the right under "Available"

**Expected behavior**
The icons should coincide with those used on the Layout tree panel.

**Screenshots**
Current:
![MapChannelsWrongIcons](https://user-images.githubusercontent.com/102841136/216772959-8167fb22-2766-4d66-b733-d3698697cde9.png)
Expected:
![MapChannelsRightIcons](https://user-images.githubusercontent.com/102841136/216773009-4690623f-e1be-49f3-9fd9-e7bb5abcccff.png)

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version: 2022.19 - 2023.02, daily

**Additional context**
I ran this in the debugger.  LayoutUtils is ignoring the returned index given by imageList->Add, that return value goes 0, 2, 3, ..., somehow skipping number 1.

I will submit two PRs with patches that potentially fix this (choose one, the other, neither, but not both :-D )
Option 1: Remembers the map indexes, and uses them.  This follows the current object lifecycle model but has a bunch of cut-n-paste code
Option 2: Consolidates the image list into one object that can be shared.  This changes the object lifecycle, and gets rid of the cut-n-paste code.  However, I don't know if this is safe.  I get nervous because for some reason all the icons are black for me sometimes, so something is behaving incorrectly, but this is not directly related to my change...  If someone knows why this happens I'd love to understand it at a deeper level.

keithsw1111:
Pull request merged

PR #3713, opened 2023-02-04
#3712 option 1

merryoncherry:
Keeps track of the indexes assigned to the icons, and uses that when populating the "Available" table on the import dialog.


keithsw1111:
Per you comment on the ticket I think I will go with the safer option for now ... I dont have the problem with consolidation but I equally dont know this code.

merryoncherry:
Thank you

Issue #3740, opened 2023-03-01
Poor render quality on warp effect (among others)


merryoncherry:
**Describe the bug**
At certain render sizes, warp effects, effects that rotate the layer buffer (via Transformation, not roto-zoom), and probably others get static noise in their renders.

![UnstableRender](https://user-images.githubusercontent.com/102841136/222185274-2b458fc3-9703-445c-9acb-74a45af5539f.png)

**To Reproduce**
Render the attached sequence.

This is due to an unsuitable use of parallel_for in one of the core algorithms in RenderBuffer that has multiple threads accessing the same memory locations concurrently.  This SetNodePixel routine copies data from the underlying nodes to initialize the RenderBuffer, so the bug can affect any layer that reads the layer below in a different way than the previous layer, but is particularly noticeable when the buffer is big enough to warrant parallelism but small enough that the threads step on each other.  In my case, the "sweet spot" for reproducing is in the few thousand nodes range.

**Expected behavior**
The effect should look nice, without any "salt and pepper" noise, pixels in places other than where they should be, etc.

**Screenshots**
See description.

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version: <2022.19 - 2023.03 / present

**Additional context**
I will submit a PR that turns off parallelism in this section of code.  I think it should be approved as a "quick fix" for the issue.

One may wonder what the effect on performance will be.  I was unable to discern a difference when rendering all sequences in my 2022 show.  Here is the performance result for 3 renders of Greatest Show, which I've heard is a standardized benchmark :-)
**_Release build 2023.03_**
163.8
164.4
157.8

**_Special build w/ parallel_for disabled in submitted patch:_**
162.7
158.7
161.5

Also, one other sequence picked essentially at random (Carol of the Bells, Epic Version):
**_Release build 2023.03_**
66.192
58.439
56.357

**_Special build w/ parallel_for disabled in submitted patch:_**
62.158
59.037
56.998

The performance difference, if any, is in the noise here... the patch is worth it for better render quality.  If anyone can find a case where the performance difference matters, we should go back and look for a better approach.

**Attachments**
[LargeWarp.zip](https://github.com/smeighan/xLights/files/10863113/LargeWarp.zip)

PR #3741, opened 2023-03-01
Parallel_for in RenderBuffer::CopyNodeColorsToPixels causes bad renders

merryoncherry:
See #3740 for a description of the resulting problem.

The easy fix is to take out the parallel_for.  I didn't notice any performance difference in a practical setting (my 2022 show bulk render), but it fixed a number of glitchy effects.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,18 @@ Issue Tracker is found here: www.github.com/smeighan/xLights/issues
 
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
+2023.03 February 22, 2023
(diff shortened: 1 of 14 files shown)


dkulp:
I think the other option would be to change "std::vector<bool> &done" to "std::vector<uint8_t> &done"

merryoncherry:
I will be happy to try that.

merryoncherry:
Right, vector<bool> is special and has the threads fighting for values in the same byte.  Otherwise it's possibly fine (on x86).
It seems to work on my simple test case from the original issue.  I will try a more extensive test next.

merryoncherry:
Seems to work as well in the quick tests, updating the PR.

merryoncherry:
Thanks for the review @dkulp

PR #3746, opened 2023-03-08
Confirmation prompt when cancelling the Submodels creation dialog window

MrPierreB:
The submodels dialog is where a lot of time gets spent and accidentally hitting the escape key will close the window without any prompt. It is one of the few windows in xlights that doesn't have any kind of confirmation when closing.

I have added a confirmation dialog when you press the escape key on the keyboard or using the cancel button on the screen.    

We could potentially just disable the escape key being used completely if needed.

Here is a screenshot.
ClosingSubmodels

### DIFF
--- a/xLights/SubModelsDialog.h
+++ b/xLights/SubModelsDialog.h
@@ -316,9 +316,9 @@ private:
     void OnInit(wxInitDialogEvent& event);
     void OnNodesGridCellRightClick(wxGridEvent& event);
     void OnCheckBox_OutputToLightsClick(wxCommandEvent& event);
-    void OnCancel(wxCloseEvent& event);
     //*)
 
+    void OnCancel(wxCloseEvent& event);
     void OnPreviewLeftUp(wxMouseEvent& event);
     void OnPreviewMouseLeave(wxMouseEvent& event);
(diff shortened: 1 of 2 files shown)


xLights/SubModelsDialog.h
```diff
@@ -316,6 +316,7 @@ class SubModelsDialog : public wxDialog
     void OnInit(wxInitDialogEvent& event);
     void OnNodesGridCellRightClick(wxGridEvent& event);
     void OnCheckBox_OutputToLightsClick(wxCommandEvent& event);
+    void OnCancel(wxCloseEvent& event);
     //*)
```
merryoncherry, line 320: This change appears inside the wxSmith block and I don't see a change to the .wxs file?

merryoncherry:
I like it.  I accidentally / instinctively hit escape when I want to undo a change to a node list cell and the whole dialog disappears and takes my work with it... so this will be good.

MrPierreB:
> I accidentally / instinctively hit escape when I want to undo a change to a node list cell and the whole dialog disappears and takes my work with it..

I do this a lot.  I did a massive set of submodels and realised my mistake a split second after i pressed escape.

Issue #3756, opened 2023-03-19
Easier submodel creation for models with rotational symmetry


merryoncherry:
**Is your feature request related to a problem? Please describe.**
For stars, wreaths/circles, and custom models with rotational symmetry, it would be nice to draw one submodel and have xLights make the rest.  It would also be nice to get rid of nodes that appear in the resulting submodel more than once.

**Describe the solution you'd like**
There is a description here:
https://merryoncherry.org/2023/02/28/new-utilities-for-adding-submodels-to-spinner-props/

**Describe alternatives you've considered**
This can all be done manually, it is very time-consuming however.

**Additional context**
Preliminary PR available.  It has undergone some testing.

computergeek1507:
Your WordPress link requires a login...

merryoncherry:
It is not supposed to need that any more but if it asks it is 'spinner'.

PR #3757, opened 2023-03-19
Add some utilities to Submodel Dialog

merryoncherry:
See issue #3756 and https://merryoncherry.org/2023/02/28/new-utilities-for-adding-submodels-to-spinner-props/ for a complete description of the provided functionality.

Makes it easier to:
1. Create rotational copies of submodels
2. Split a stranded submodel into multiple
3. Join submodels as a single strand
4. Remove duplicates in submodel strands or replace with blanks
5. Sort submodel nodes geometrically
6. Space out the nodes in a stranded submodel evenly
7. Pivot the strand rows/columns to become columns/rows (same effect as vertical buffer layout, but easier to see what is happening)

One bit of feedback from beta testers was the desire to support ",0," instead of ",," for blank nodes to make it easier to tell what is going on.  I added this, but am not sure what the thoughts are on that.


xLights/SubModelsDialog.cpp
```diff
@@ -961,21 +986,100 @@ void SubModelsDialog::OnNodesGridLabelLeftClick(wxGridEvent& event)
...
         //NodesGrid->GoToCell(event.GetRow(), 0);
-        wxMenu mnu;
 
         mnu.Append(SUBMODEL_DIALOG_REMOVE_DUPLICATE, "Remove Duplicates");
-
-        mnu.Connect(wxEVT_MENU, (wxObjectEventFunction)&SubModelsDialog::OnNodesGridPopup, nullptr, this);
-        PopupMenu(&mnu);
-    }
+        mnu.Append(SUBMODEL_DIALOG_ELIDE_DUPLICATE, "Elide Duplicates");
+        mnu.Append(SUBMODEL_DIALOG_SORT_POINTS, "Sort Strand Points Geometrically...");
```
computergeek1507, line 995: what is the difference from Elide to Remove? "Elide" normally means "omit"(I think) so is Elide removing all instances of a node if there are duplicates?

merryoncherry: For elide, we are leaving a space so that the timing isn't affected.
1,2,2,3 becomes 1,2,3 with remove, or 1,2,,3 with elide.

merryoncherry: Certainly open to suggestions on a better word for elide, but I took it from a definition I saw, meaning "to suppress" or "to strike out".

merryoncherry: Give me a different word Gil and I'd be more than happy to do a find/replace for you.  Other people had the same reaction to the use of the word "elide", but they didn't have a suggestion and said it was fine.  Would be great to have a good suggestion.

AzGilrock: I haven't dug into this so I can't figure out why as programmers having ",0," vs ",," makes anything easier.  We should be able to deal with either.  And does this change work with all existing models.  Some folks will have the old format now does all code need to deal with both versions?

merryoncherry: The more I think about it, the more I prefer the term "suppress" vs. "elide".  What do you folks think about that change.  Remove duplicates is 1,2,3; suppress duplicates would be 1,2,,3.

AzGilrock: Yep "suppress" was what I had in mind also when I read your comment about how you found the "elide" term.

merryoncherry: OK, I changed that.

If you would like any explanation or discussion of anything else, you can always email me (any address at the merryoncherry.org domain goes to me) and we can set up a zoom or something.

Issue #3775, opened 2023-04-05
Value curves that don't get reset between effects


merryoncherry:
**Describe the bug**
In some cases, if you create two effects of the same type, one with a value curve, and one without, when you select the one with the value curve, then the one without the value curve, it shows the value curve as active when it shouldn't be.  A previous example of this was #3637 (see the instructions for reproducing the issue there), but I found the following additional 7 cases:
Bars - Center
Liquid - Gravity & Angle
Ripple XC, YC
Shape velocity and direction

**To Reproduce**
Follow the steps in #3637, but substitute the "Center" curve for bars effect, or the gravity angle for Liquid, and so on.

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version: 2023.05, daily, many prior versions

**Additional context**
I have a PR for this, which I will submit shortly.

PR #3776, opened 2023-04-05
Reset Value Curve buttons

### LINKED ISSUE #3759: Debug assertion failure: fpp_app_icon_svg is getting strlen'd but not null terminated
**Describe the bug**
Nondeterministic behavior on xLights startup, prevents startup in the debugger.
fpp_app_icon_svg is getting strlen'd but isn't null terminated.

**To Reproduce**
Start xlights under the debugger.

 - OS: Windows 10
 - xLights version: 2023.04 / master

**Additional context**
This is probably a case of using the wrong constructor in BitmapCache.cpp, it could pass the length as well as the pointer so that there is no strlen call:

```
 wxBitmapBundle BitmapCache::GetFPPIcon() {
-    return wxBitmapBundle::FromSVG((const char*)fpp_app_icon_svg, wxSize(16, 16));
+    return wxBitmapBundle::FromSVG((const wxByte*)fpp_app_icon_svg, sizeof(fpp_app_icon_svg), wxSize(16, 16));
 }
```

merryoncherry:
A few cases where the value curve bitmap button wasn't reset when resetting the panel via SetDefaultParameters().  See issue #3775.
Also fixes #3759.
Verified that this doesn't cause any rendering differences in a small test set that tries to do a little bit of everything.

### DIFF
--- a/xLights/BitmapCache.cpp
+++ b/xLights/BitmapCache.cpp
@@ -601,7 +601,7 @@ wxBitmapBundle BitmapCache::GetModelGroupIcon() {
     return CreateBitmapBundleFromXPMs(16, "ModelGroup", {model_16, model_64, model_64, model_64, model_64});
 }
 wxBitmapBundle BitmapCache::GetFPPIcon() {
-    return wxBitmapBundle::FromSVG((const char*)fpp_app_icon_svg, wxSize(16, 16));
+    return wxBitmapBundle::FromSVG(fpp_app_icon_svg, sizeof(fpp_app_icon_svg), wxSize(16, 16));
 }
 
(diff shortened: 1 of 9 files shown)


xLights/BitmapCache.cpp
```diff
@@ -601,7 +601,7 @@ wxBitmapBundle BitmapCache::GetModelGroupIcon() {
     return CreateBitmapBundleFromXPMs(16, "ModelGroup", {model_16, model_64, model_64, model_64, model_64});
 }
 wxBitmapBundle BitmapCache::GetFPPIcon() {
```
merryoncherry, line 603: The former call is for a string terminated with \0, but this fpp_app_icon_svg does not end in \0.  Drives the autodebugger nuts.

Issue #3779, opened 2023-04-07
Enhancements for the Ripple Effect


merryoncherry:
The ripple effect has great potential, but has a few limitations.  The ripple spacing isn't configurable, and behavior differs depending on which built-in shape it is.  It only has two available motions, implode and explode.  Rotation only works on some shapes.  There aren't that many shapes.  Etc.

I have a prototype that adds some new features to it; you can change the size, motion, twist, direction, etc., and the colors are more flexible.  Here is a draft updated documentation page (subject to change with feedback):
https://drive.google.com/file/d/1Rfn6FzpwgwS-ocoh_BleKX51YVude6Vr/view?usp=share_link

Here is a model preview video from the updated effects:
https://drive.google.com/file/d/1RA4YI_58XZQlwxmHtA9mjYdCh-LrK814/view?usp=share_link
I tried to make it as independent from the matrix resolution as I could, so it can be a pretty good stand-in for shockwave, which is notoriously finicky.  10x resolution on the exact same effects as above:
https://drive.google.com/file/d/168En-l_YMrz5ZrmFk6jFdPNbb73Pgrdq/view?usp=share_link

SVG is one of the fun new features:
Medium-low res matrix:
https://drive.google.com/file/d/1eMlQIiTIsPI5rl7uHA3oQ3kL97v96tNE/view?usp=share_link
Same sequence on high res matrix:
https://drive.google.com/file/d/1ebdCCY-5vG02MQ53Ak8Kybi-4rBDBnzA/view?usp=share_link

As usual, if you don't want to discuss here for whatever reason, email any address at the merryoncherry.org domain.

I will post the draft PR with known issues/sticky points shortly.

derwin12:
Can this be closed then?

PR #3780, opened 2023-04-07
Ripple Effect enhancements

merryoncherry:
See issue #3779 for the draft documentation and links to example video.

There are a couple of outstanding issues found by the testers:
1. The effect panel is wider than desired, and if you expand it, it won't shrink fully.  This is common to some other panels and dialogs.  Will keep searching.
2. This expands the range of the rotation value curve, which has an effect on existing sequences that use the rotation value curve.  There may not be many, since this VC was essentially broken until just a couple months ago.  Nevertheless, I am investigating better upgrade for value curves.

Beta testing continues.  There are probably other issues.

A few other points noted inline below.


xLights/wxsmith/RipplePanel.wxs
```diff
@@ -45,12 +98,38 @@
 					<size>14,14</size>
 					<bg>wxSYS_COLOUR_BTNHIGHLIGHT</bg>
 					<style>wxBU_AUTODRAW|wxBORDER_NONE</style>
```
merryoncherry, line 100: I cannot get wxSmith to stop clobbering the action handlers.  In the end, I let it win, and resorted to adding the connect call manually, see the .cpp file where they are carefully moved outside the generated section.

xLights/effects/RipplePanel.cpp
```diff
@@ -203,31 +372,60 @@ RipplePanel::RipplePanel(wxWindow* parent) : xlEffectPanel(parent)
 	FlexGridSizer57->Add(-1,-1,1, wxALL|wxALIGN_CENTER_HORIZONTAL|wxALIGN_CENTER_VERTICAL, 5);
 	FlexGridSizer57->Add(-1,-1,1, wxALL|wxALIGN_CENTER_HORIZONTAL|wxALIGN_CENTER_VERTICAL, 5);
 	SetSizer(FlexGridSizer57);
+	FlexGridSizer57->Fit(this);
+	FlexGridSizer57->SetSizeHints(this);
 
 	Connect(ID_CHOICE_Ripple_Object_To_Draw,wxEVT_COMMAND_CHOICE_SELECTED,(wxObjectEventFunction)&RipplePanel::OnChoice_Ripple_Object_To_DrawSelect);
```
merryoncherry, line 378: These are all present below outside the generated section.  wxSmith kept forgetting about the action handlers in the .wxs file and propogating that, so putting them outside the generated section seemed much more reliable.

PR #3842, opened 2023-06-14
Moc batchrender

merryoncherry:
Give -r/--render option to xLights a little TLC.
  If there are any deferred actions from startup that abort render, retry once.
  A few things to suppress when in this mode: splash, missing item errors.
  Small thing - "Unable to bind to XL_DURATION" message in log produced by shader is just noise, take it out.


xLights/effects/ShaderEffect.cpp
keithsw1111, line 1: I dont agree with suppressing this warning. Maybe thats not the real problem ... why is the binding failing? Why would this be a problem and XL_ZOOM not a problem ... there is more likely a different problem here.

merryoncherry: I had that thought... what makes some different than others.

XL_ZOOM is used in vsSrc calculations that are prepended to the user code to determine the xl_FragCoord, and therefore is always used.

XL_DURATION is not used in this.  XL_DURATION was added last year for the user's option (say, to get a fixed cycle count in the time interval of the effect).  So, the reason why XL_DURATION may not be found is that it could have been optimized out.

Now, do we know that XL_DURATION is actually working?  In the unit test suite, I added a test with the clutchplate.fs shader provided by the contributor, so I know that it does work when used.  So the code that binds it (when it exists) is functioning.

PR #3853, opened 2023-06-23
Fix FalconV4 16+4+2 to be 1024

cybercop23:
Had wrong 704 max pixel/port vs 1024
image


computergeek1507:
Are you sure the diagram isn't wrong, no understanding is that would be basically 2 expansions

cybercop23:
I don't think so. All mixes are 1024 until you actuall add the 2nd expanssion and that goes to 704. That makes sense.
image

cybercop23:
@computergeek1507  Yeah... Doc on the sales page is wrong. Thanks. Closing.

Issue #3868, opened 2023-07-04
Single-strand chase effect direction in preview window doesn't always match selection and final render


merryoncherry:
**Describe the bug**
If you click a single strand effect in the sequencer, and toggle left-right or right-left a bit, you may notice that the direction shown in the preview only goes one direction, and doesn't match the panel settings.  This doesn't always happen, but when it does if you click "Render All" then the preview will do the right thing.

**To Reproduce**
This can be seen when clicking the last chase effect in the attached sequence (at least on my machine).

**Expected behavior**
The direction should match the dropdown setting (once given enough time to render).

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights 2023.10, master, 2023.07

**Additional context**
I have traced this back to what I believe is an outdated code construct used only in single strand chase.  By deleting this temporary value and doing the simple direction calculation at the beginning of each render frame, the problem goes away.

I will submit the PR with the fix shortly.

**Attachments**
[FinickyEffects.zip](https://github.com/smeighan/xLights/files/11951908/FinickyEffects.zip)

merryoncherry:
This was fixed as #3869

PR #3869, opened 2023-07-04
Chase direction fix for #3868

merryoncherry:
These variables were mostly unused, except for in chase effect.  The maintenance and lifecycle were unclear, but seemed to cause an issue, so best idea I had was to remove them, and this made the problem go away.


merryoncherry:
After further research, I think this had been used to keep the auto reverse state in the past.  This is not the case any more, direction is calculated in a stateless way so that the effect can render an arbitrary time interval (CanRenderPartialTimeInterval returns true), so it is fine not to cache it with the buffer.

PR #3882, opened 2023-07-22
Fix Moving Head Export to include slew limits

MrPierreB:
The export of moving heads didn't include the slew limits in the xmodel.  The import of the xmodel already looks for the slew limits.


AzGilrock:
Yeah we probably could improve all these import/export routines to use routines like the Serialize call in Servo.cpp which iterates through all the fields of the ModelXml.

computergeek1507:
And prevent duplicates...

AzGilrock:
Yeah when I wrote that I know I stole it from some other class and said why aren't we doing this everywhere...lol

PR #3901, opened 2023-08-09
allow Preview to be selected when showing 3D models

djulien:
xLights currently disables the dropdown Preview choices when the 3D checkbox is turned on.  This will hide models that have a non-default Preview setting, and there's no way to see those models while the 3D checkbox is set. 

This change allows the Preview dropdown choice list to be used while the 3D checkbox is turned on - maybe it should be a UI option?


AzGilrock:
In reality I’d like to delete the preview feature but it’s greyed out for a reason.  A ton of code is broken that needs to be fixed I’d people try to use previews in 3D mode.

Issue #3904, opened 2023-08-09
Heap corruption crash in large single strand FX


merryoncherry:
**Describe the bug**
Heap corruption crash on sequence render.  Appears to be caused by FX effect DecodeMode.

**To Reproduce**
I hit this crash in the PPD "Bang!" sequence after mapping to my layout.  (I think you will not need it to reproduce to accept the code fix.)

**Versions (please complete the following information):**
 - OS: Windows 11
 - xLights version: 2023.11; nightly

**Additional context**
There appears to be an initialization for DecodeMode that caches the effects list.  This is not protected by a mutex, and is generally OK, but if there is a large enough effect that it gets put into the render pool and multiple threads initialize it at once, it will corrupt the heap. 

PR inbound.

PR #3905, opened 2023-08-09
 FX DecodeMode not threadsafe but used in render pool

merryoncherry:
See #3904


AzGilrock:
Looks like the same issue in the DecodePalettes function below the one you changed

merryoncherry:
Thanks, I'll look at that too.

PR #3913, opened 2023-08-12
Make color count available to shaders

ClutchplateDude:
When working with multiple colors in a shader, the shader has no way of knowing how many colors the user actually selected. This makes that number available.


ClutchplateDude:
#3914

Issue #3774, opened 2023-04-04
Please add controller support for Hanson P28D Hat in 2 and 4 port DPI Pixels Modes


gergmchairy:
In FPP v6.3 the Hanson Electronics HAT for Raspberry Pi has 3 output variants:

"rPi-28D"with 2 WS281x Outputs and 1 x WS280x Output.
"rPi-28D-DPIPixels" with 2 x WS2811 Outputs.
"rPi-28D-DPIPixels-4" with 4 x WS2811 Outputs.

Please could the Controller driver for the HAT be updated in xLights to match the options available in FPP.

Thanks

Greg

anubis-code:
+1 on this please

keithsw1111:
There is not a lot of information in this so I took a cut at it ... if it is wrong it is because this requests is missing information. I would prefer in future that such requests come as a pull request updating the xcontroller file as we have no access to these controllers or knowledge of the controller capabilities.

PR #3920, opened 2023-08-15
Fixes issues with DPI-Pixels uploads for Hanson rPi-28D from #3774

mjunek:
Closes https://github.com/FalconChristmas/fpp/issues/1538 - fixes issue introduced in bb8ea3c

Updates controller definition to use the DPIPixels Abstract Driver, rather than the default RPIWS281X driver

### DIFF
--- a/controllers/hanson.xcontroller
+++ b/controllers/hanson.xcontroller
@@ -131,7 +131,7 @@
             <fppSerialPort1>ttyAMA0</fppSerialPort1>
             <fppStringFileName>co-pixelStrings</fppStringFileName>
 		</Variant>
-		<Variant Name="rPi-28D-DPIPixels" ID="rPi-28D-DPIPixels" Base="FPP:BaseFPPSettings">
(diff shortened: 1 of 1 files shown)


SarahDods:
Thank you!

Have also found tonight that uploading a sequence from xLights to the licensed RPi28d+ cape using FPP connect also causes a crash with Error 503.  Is this the same issue and already resolved, or a new bug I should report separately?

mjunek:
> Thank you!
> 
> Have also found tonight that uploading a sequence from xLights to the licensed RPi28d+ cape using FPP connect also causes a crash with Error 503. Is this the same issue and already resolved, or a new bug I should report separately? This is the error message back (box address is 192.168.25.125), and last line is the relevant bit.

I'd hazard a guess this is an unrelated issue. The fix above was to change the Driver from the old RPIWS281x to the newer DPIPixels in the controller definition. 
The 503 message will be coming back from FPP -so raise it there first and we can take a look.

PR #3933, opened 2023-08-22
Fixes the hyperlink for FPP behind proxy in FPP Connect Dialog.

dartunder:
Stores the proxy info in the discovered instance of FPP
And fixes the location link in FPPConnectDialog.cpp


dartunder:
@computergeek1507 

This was added because it does not work without this part. 
                fpp->proxy = res->proxy;

 As you can see later, this field is updated ... so it's logical that it should be initialized too.
                setIfEmpty(fpp->proxy, res->proxy);
                
                
Alternative.... in
FPP::FPP(const std::string& ip, const std::string& proxy, const std::string& model) :
maybe this function should store the proxy value instead of only passing it to the base controller initializer.

I'm not sure why there is a "_fppProxy" variable in BaseController AND a "proxy" variable in FPP

dartunder:
I didn't see your prior commit...  [added proxy to constructors]
 That should work.

Issue #3989, opened 2023-09-25
Timeline zooming centered on mouse marker (QoxL)


hoodlumj3:
Have the ability to use the waveform timeline mouse marker position ( vertical blue line on waveform ) to be the center of the timeline zoom in / out.

I'm offering to build this, wait, I have built this already [Branch:timeline-zooming](https://github.com/hoodlumj3/xLights/tree/timeline-zooming)

**Is your feature request related to a problem? Please describe.**
No, but yes - timeline editing efficiency problem.

**Describe the solution you'd like**
Ability to change timeline zoom level centered target (CTRL + Mouse Scroll Wheel) based on the mouse marker position (currently its the start play marker)
Preference option to select if you want timeline zooming to zoom in/out using Mouse Marker or Play Marker 
(naming to be considered / discussed)

**Describe alternatives you've considered**
No alternatives as there is only one way this can be workflow'd.
The current zoom in on play marker can get you disorientated and lost in the timeline in an instant (as its a hard zoom in / out) if you crazy zoom in or scroll the timeline window left or right (SHIFT +  Mouse Scroll Wheel) away from the play start marker.
Only way to recover / find where you were is to zoom all the way out, quickly scan the timeline and attempt to recognize were you think you were editing (as its not always on the play start position because scroll bars), click on timeline and zoom in again.

**Additional context**
I class this as Quality of xLights (Life) Enhancement ;)
Current zoom feature is not an effective use of time and as 95% of software that i've used does this OOTB already.
I believe this feature will increase the qualilty of xlights and life in general.

please discuss and feedback welcomed ;)

keithsw1111:
Raise a pull request. Is this controlled by an option or are you changing the behaviour for all users.

hoodlumj3:
Pull request incoming.

awww you ask such hard questions... ;D 

The zoom setting, on windows at least, is stored in the HKCU registery by xLights so its a local setting per user.
The default setting is set to use current behaviour - zooming on the start play marker.
Zooming setting is in preferences under the View tab.
I tracked down the colour setting i added to the xlights_rgbeffects.xml file so its show (folder) based, part of the layout.

hoodlumj3:
PR #3991

PR #3991, opened 2023-09-25
Timeline zooming centered on mouse marker (QoxL)

hoodlumj3:
#3989 

hope i'm doing this correctly, this is my first ever PR

### DIFF
--- a/xLights/ColorManager.h
+++ b/xLights/ColorManager.h
@@ -57,6 +57,7 @@ class ColorManager
             COLOR_LOCKED_EFFECT,
             COLOR_EFFECT_SELECTED_DISABLED,
             COLOR_REFERENCE_EFFECT_DISABLED,
+            COLOR_WAVEFORM_MOUSE_MARKER,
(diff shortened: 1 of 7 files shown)


keithsw1111:
I dont see the wxs file? Did you not use codeblocks? If not then all your UI edits will get lost the next time someone edits the panel.

hoodlumj3:
@keithsw1111 nice catch, appologies latest commit now includes the wxSmith file

AzGilrock:
That StaticText6 object can be set to member="No" unless you need to programmatically change the text later.  I believe the default should have been No so you must have specifically changed it to Yes.  When its No it defines it locally in the constructor of the cpp file instead of needing to be in the header.  Not a big deal just pointing it out since it looks like it was changed from the defaults.

hoodlumj3:
@AzGilrock I did wonder why it was sitting out in the class instead of local all of a sudden - thx for the hot tip, I shall correct for consistency.

PR #3992, opened 2023-09-25
show more info so user knows how important are missing models

djulien:
This change lists the model names and #effects, so the user knows how important are the missing models.  -helps avoid user panic if there are actually no effects on those models  :)


xLights/sequencer/tabSequencer.cpp
```diff
@@ -504,13 +504,21 @@ void xLightsFrame::CheckForValidModels()
 
     logger_base.debug("CheckForValidModels: Remove models that already exist.");
 
-    size_t missingModelCount = 0;
+//    size_t missingModelCount = 0;
+    std::vector<std::string> missingModelCount;
```
AzGilrock, line 508: I'd suggest renaming that variable to maybe missingModels since its not really a count anymore.  Also I'd just delete those 2 lines you commented out.  We can look at history if we need to know what was there.

djulien: okay, I made those updates.

PR #3996, opened 2023-09-27
Add K8-Pi4 ctrl type for dual virtual matrix support

cybercop23:


computergeek1507:
One of the video buffers is used by the DPI pixel driver, I don't think this would work. Have you tested it on a real controller ?

dkulp:
Yea.. this won't work.

PR #3998, opened 2023-09-27
work-around for empty Select Effect panel

djulien:
The choice lists on the Select Effects panel are not populating for me.  I've tried various 22.x and 23.x versions of xLights and they all behave the same way.  There are many GTK warnings and errors on the console, so maybe my environment is not fully supported (Ubuntu MATE 22.04, x64).  However, the rest of xLights works okay so the warnings/errors seem to be benign.

If it's an environmental issue, I have no idea how to track this down in a time-efficient way.  This simple work-around allows the user to force the Effects Type drop-down list to populate if it's empty by clicking the Refresh button.  If ListCtrl_Select_Effects is already populated, this change won't do anything so it won't affect users who are not experiencing the problem.


keithsw1111:
This should be LINUX specific. Please using #ifdef to only compile it in the Linux build

djulien:
added #ifdef as requested

djulien:
thanks!

Issue #4008, opened 2023-09-30
Add RGB support to Image Model


lasershane:
I use many props that are not naked lights, mostly consisting of some shape illuminated by an external flood light or an internal light.
The image model can represent these props when using a single light color, but if a prop is lit by an RGB flood light, it is not well represented by a pixel model (which doesn't show the image shape/texture) nor by the image model (which doesn't show the color).

I propose adding RGB support to image models. When the default "single color" string types are selected, it would behave exactly as before for compatibility. If an RGB pixel type is selected, then the model will be recolored to match the pixel color in the preview.

I plan to implement this behavior and have a hack that appear to be working using a second image as a color overlay, but I'd appreciate any input/insight/feedback from the core developers on this topic.

<img width="441" alt="Capture" src="https://github.com/smeighan/xLights/assets/1802067/78fd1ace-a0ee-43f6-8589-f6d52b6bca08">

keithsw1111:
I am not sure I would use a second image ... can you not colour alter the image in line with the RGB colour? Unless you mean you would have original image and a copy you colour ... that would be an implementation efficiency I guess.

lasershane:
Maybe "second image" isn't the best description. I have not found a good way to draw a colorized version of a texture using the graphics library, and modifying the color of every pixel each frame would be very slow. 
What I am currently doing instead is drawing a grayscale version of the chosen image, then drawing a semi-transparent color texture on top. I do this by converting the original image to grayscale, then making a duplicate texture that copies the grayscale data to the alpha channel. Drawing these on top of each other gives a very good looking result. 
I have posted a draft PR with the planned changes.

PR #4009, opened 2023-10-01
Add support for rgb color image models

lasershane:
Implements enhancement from issue #4008 

Image models can now have string types that allow color, and images will be colored in the preview to match.

Known issues: transparency is not handled correctly for color image models


lasershane:
Here is a sample image of the preview with a few candy gumdrops and giant lightbulb stakes to show the effect. The supports for the giant bulbs are separate images.
Bulb
Gumdrop

image

keithsw1111:
Is this going to move from draft?

PR #4030, opened 2023-10-07
Don't resolve bad.address to 255.255.255.255, eliminate duplcate FPP …

### LINKED ISSUE #3942: FPP Connect showing duplicates because of case?
image

Not sure where it would get all the variations .. but can it be coded not to occur?

.LOCAL
.Local
.local

dartunder:
1. In "xLights/utils/ip_utils.cpp" 
"bad.address" was being resolved as 255.255.255.255.  Mirroring the behavior of "0.0.0.0" we can ignore this result and return the original value.

2. In "xLights/controllers/FPP.cpp" function "FPP::GetInstances"
  To prevent unexpected duplicate entries in the FPP connect dialog, we can attempt to resolve the IP or at least use the lowercase version of the address.  Currently, FPP01.local, fpp01.local, and 192.168.1.25 if entered as proxies for different controllers, can appear in the FPP connect dialog as separate devices when they all resolve to the same IP.    
  NOTE:  The result of this change is that the FPP connect dialog would almost always only show:
     single HOSTNAME and IP address:   FPP01 -192.168.1.25
          vs.
     multiple HOSTNAME and DNS names:  FPP01 - FPP01.local, FPP01 - fpp01.local, FPP01 -192.168.1.25
  
3. In "xLights/controllers/FPP.cpp" function "FPP::UploadUDPOutputsForProxy"
In order for UDP Out Mode: "Proxied" to work correctly currently, the proxy address of the controller must exactly match the address as displayed in FPP Connect.  So, if 2 is implemented we also need to resolve the controller proxy addresses when determining if the device should be a proxy.  In theory, we could just compare the lowercase result of ResolveIP of the c->GetFPPProxy() and "ipAddress".  However, without having enough experience and test data, I'm not sure this covers all cases and exceptions, so I compared both addresses and the resolved IPs.  If either match, then enable proxy for UDP.  This may be paranoia, but I felt it was safer than making other assumptions.

@dkulp I know you know the FPP connect section very well.  I was hoping for feedback before approving this pull request.  Feel free to blow this up and do it yourself if you have a better way.

Closes #3942

### DIFF
--- a/xLights/controllers/FPP.cpp
+++ b/xLights/controllers/FPP.cpp
@@ -63,6 +63,7 @@
 #include "../FSEQFile.h"
 #include "../Discovery.h"
 #include "../utils/CurlManager.h"
+#include "../utils/ip_utils.h"
 
 #include "Falcon.h"
 #include "Minleon.h"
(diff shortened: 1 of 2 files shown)


dkulp:
Looks fine by me.   This was on my todo to look at once I get halloween setup so thanks.  :)

dartunder:
Thanks again; now that you have relased 16, this is a good time to commit this request.

PR #4048, opened 2023-10-13
set the rotation to match the original when replacing a model

dartunder:


dartunder:
https://github.com/smeighan/xLights/issues/3924

PR #4051, opened 2023-10-13
Node Count Tooltip addition

kylegrymonprez:
Initial enhancement request for feedback: https://github.com/smeighan/xLights/issues/4050

Added the ability to count the number of nodes for display inside the tooltip of all selected models in the layout tab to the tooltip for selected models.

Additionally, there was also a use case of selecting only one group with some previous logic that didn't always fire. In the case where that tooltip was not firing, I also added the node count of the single group in another tooltip.


keithsw1111:
Really not a fan of this. Why would the model tooltip be impacted by other selected models. If you were going to display this somewhere it would be in the status bar or something.

dartunder:
Maybe I'm missing something the tooltip is already affected by "other selected models."
If you select two models, it shows the total groups, models, sub models selected.

kylegrymonprez:
@dartunder is correct. I added text to an existing tooltip text. However I'll open to displaying it in like....the bottom right corner of the frame. That was my original thought before that tooltip was pointed out. 

I can make the change to have it in both, or one or the other. Just let me know what the preference is.

keithsw1111:
Can you get rid of the excessive logging and ensure that just because it is coming from base the behaviour isnt different.

kylegrymonprez:
Removed the extra logging. All my testing couldn't find any issues from where the code is located.

(Committed and reverse committed a POC of moving the count to the lower right, but would need some extra attention to have the layout not thrash.)

PR #4065, opened 2023-10-19
add debuging for resolvedIPMap

dartunder:
It appears that some name.local addresses are not being initially resolved to IP addresses, at least on a Mac.

Feel free to kill this if you have a better idea.

std::string ResolveIP(const std::string& IP) in ip_utils.cpp is not always resolving valid names to ip4 addressses.


dartunder:
This doesn't go far enough to actually diagnose the problem only highlight it, so I will close this.

PR #4068, opened 2023-10-21
Confirm then delete selected models including locked models

dartunder:
Fixes https://github.com/smeighan/xLights/issues/4067

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -6608,13 +6608,9 @@ void LayoutPanel::DeleteSelectedModels()
             for (const auto& it : modelsToDelete) {
                 auto model = xlights->AllModels[it];
                 if (model != nullptr) {
-                    if (!model->IsLocked()) {
-                        xlights->GetDisplayElementsPanel()->RemoveModelFromLists(it);
-                        allDeleted = xlights->AllModels.Delete(it) && allDeleted;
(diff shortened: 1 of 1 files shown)


dartunder:
This change provides a consistent delete warning behavior regardless of the selected model.  It explains why a model cannot be deleted rather than just making a sound that many users won't notice.  This allows unlocked models to be deleted even if a locked model is also selected as requested.

AzGilrock:
I feel like the code already there is wrong.  It seems unacceptable to allow a partial deletion.  Does the user even know that's going to happen?  I would not want to select 50 models and have it delete 25 and tell me "some of them couldn't be deleted".

dartunder:
Thanks Gil. 

The current code already deletes the non locked models. I was simply trying fix an issue that causes the behavior to be inconsistent. 

However I agree that it would likely be best to produce the error message first and don't allow delete of any unless it can delete all items.

But that is an enhancement not a bug fix which this was intended to be. 

I can attempt to change the behavior, but this as it stands is still better than leaving it as is.

AzGilrock:
Yeah my thinking is most people are locking models so they don't accidentally drag them around the screen.  Most of the time people try to delete a model they really want it deleted.  I'd be happy if we just deleted everything selected.  If they didn't want to delete they can restore a backup.

dartunder:
This PR was updated to delete all selected models after a confirmation dialog.

PR #4071, opened 2023-10-21
Change sequence effect is present indicator to support sub-models/layers/groups

robfallone:
This changes the behavior of the recently added "effects present indicator" and tries to accommodate the requests made in https://github.com/smeighan/xLights/issues/4016.

**It changes the behavior to the following:**
1. Display a blue indicator on a line that _directly_ contains an effect on that line.
2. If line is not blue already (above), then display a yellow indicator on a line (group / model / submodel / layer / strand) that contains an _indirect_ effects within that line's submodels/layers/strands/nodes.

Yes, there was previously a decision to not put an indicator on a group that has its models already visible in the view. However, with the advent of a new color/indication I strongly believe there is good reasons to change this policy. **They are:**
1. Most importantly, I find it way easier to find buried effects (even from the ALL group). Having an indicator at the group level is really useful for this purpose.
2. Consistency goes a long ways with user experience. If things are not "seemingly" consistent then users will get confused and not use program correctly. When this feature initially came out several seasoned users in the zoom room were confused. Meaning it at least was not intuitive for them. The new policy will be "more" consistent in that blue indicators will always be visible for _direct_ effects and yellow indicators will be visible for _indirect_ effects.

**Here is a screenshot showing the output:**
image


robfallone:
Pending discussion in https://github.com/smeighan/xLights/issues/4016 , we may improve the indicator with this PR or remove it entirely with https://github.com/smeighan/xLights/pull/4076.

xLights/sequencer/RowHeading.cpp
```diff
@@ -1636,27 +1639,57 @@ void RowHeading::render( wxPaintEvent& event )
...
+                {
                     int view = mSequenceElements->GetCurrentView();
                     ModelGroup* mg = dynamic_cast<ModelGroup*>(m);
                     auto models = mg->ModelNames();
-                    for (auto it = models.begin(); !hasEffects && it != models.end(); ++it) {
+                    for (auto it = models.begin(); !containsEffectsIndicator && it != models.end(); ++it)
+                    {
                         ModelElement* mm = dynamic_cast<ModelElement*>(mSequenceElements->GetElement(*it));
 
-                        if (mm != nullptr && !ModelInView(*it, view)) {
```
robfallone, line 1651: Note, ModelInView() is a painful O(n squared) linear search, that has been removed.

robfallone:
Given @keithsw1111 's feedback in [4076](https://github.com/smeighan/xLights/pull/4076), I will close/reject this PR and keep status-quo.

PR #4076, opened 2023-10-23
Remove effect present indicator

robfallone:
This PR is to address https://github.com/smeighan/xLights/issues/4016 . There is discussion/debate whether the indicator should be improved or removed. Depending on the outcome of that discussion, this PR may be needed.

### DIFF
--- a/xLights/sequencer/EffectsGrid.cpp
+++ b/xLights/sequencer/EffectsGrid.cpp
@@ -4416,10 +4416,6 @@ void EffectsGrid::DeleteSelectedEffects()
     mResizing = false;
     mDragging = false;
     ForceRefresh();
-
-    // we need to update the row headings due to the effect indicator and this might be the last effect on the model
-    wxCommandEvent eventRowHeaderChanged(EVT_ROW_HEADINGS_CHANGED);
-    wxPostEvent(this, eventRowHeaderChanged);
 }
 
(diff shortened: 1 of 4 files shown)


dartunder:
@keithsw1111 since you created the initial code for this feature, I believe you should decide how we proceed.

I recommend closing this PR.  Then approve or review https://github.com/smeighan/xLights/pull/4071 to see if it is acceptable to you.  As he said, it has been optimized to minimize processing time and should not affect render time. I believe it still accomplishes your original intent in a slightly different way that some people think is more intuitive.

If you want to go in another direction, please guide us toward your vision or close both PRs and make your changes.

Thanks.

keithsw1111:
@dartunder I am aware of that ... I just dont agree that groups containing models with effects should be tagged if the models are shown as their own line item on the effects grid. Doing so makes the tag on model groups essentially pointless. I know some people wont find this intuitive but the reverse is not useful. 

I can see the benefit of a visual indicator showing effects exist but only if it actually helps the sequencer. While i get this may not be the most intuitive of exceptions it makes the feature actually useful.

Until I see a pull request that improves the feature and respects this I wont be approving it. When i get the time i will come back and do this if no one else does before i do.

dartunder:
Thank you for clarifying your intent. It's good to have your direction on this topic.

robfallone:
Given @keithsw1111 's feedback, I will close/reject this PR and keep status-quo.

PR #4082, opened 2023-10-26
Write FPP connect upload results to the status bar.

dartunder:
Similar to upload outputs, have FPP connect upload also write its results to the status bar.

Note if an error or warning occurs and that information is added to the messages variable then a dialog already appears.

Prompted By:
https://github.com/smeighan/xLights/issues/4081

### DIFF
--- a/xLights/controllers/FPPConnectDialog.cpp
+++ b/xLights/controllers/FPPConnectDialog.cpp
@@ -1112,10 +1112,21 @@ void FPPConnectDialog::doUpload(FPPUploadProgressDialog *prgs, std::vector<bool>
         }
         row++;
     }
+
+    xLightsFrame* xlframe = static_cast<xLightsFrame*>(GetParent());
     if (messages != "") {
         wxMessageBox(messages, "Problems Uploading", wxOK | wxCENTRE, this);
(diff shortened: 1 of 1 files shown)


xLights/controllers/FPPConnectDialog.cpp
```diff
@@ -1112,10 +1112,21 @@ void FPPConnectDialog::doUpload(FPPUploadProgressDialog *prgs, std::vector<bool>
         }
         row++;
     }
+
+    xLightsFrame* xlframe = static_cast<xLightsFrame*>(GetParent());
     if (messages != "") {
         wxMessageBox(messages, "Problems Uploading", wxOK | wxCENTRE, this);
+        xlframe->SetStatusText("FPP Connect Upload had errors or warnings", 0);
```
computergeek1507, line 1119: I would put the messages string at the end of the status text

dartunder: I get your idea, but the messages variable has new line characters and is of unknown length, which doesn't make sense for status bar text.

Instead, I ensured the messages were logged, some of which were already logged, but this puts the contents of the message box together as shown in the message box.

PR #4089, opened 2023-10-31
[feature] Adding ability to adjust the frequency of the "Auto" eye blink

kylegrymonprez:
[feature] Adding ability to adjust the frequency of the "Auto" eye blink

Came upon some discussion online where users wanted to speed up the blinking on some props, with one idea to allow a drop down on the Faces Effect Panel. I found that the current implementation is typically a random int between 4500 and 5500 ms (with some logic around Phenoms and resting which has been preserved). Created more of a scale from "Slowest" to "Fastest", with the default of "Normal" that preserves the existing numbers.

I tested new sequences, checked their save files, reloaded correctly. I also checked older sequences opening in the new format at saw that eye blinking behavior was 'upgraded'.

If there are other testing avenues needed, please advise and I will address

### DIFF
--- a/xLights/effects/FacesEffect.cpp
+++ b/xLights/effects/FacesEffect.cpp
@@ -39,7 +39,7 @@ public:
     std::map<std::string, int> nodeNameCache;
 
     FacesRenderCache() :
-        blinkEndTime(0), nextBlinkTime(intRand(0, 5000)) {
+        blinkEndTime(0) {
     }
     virtual ~FacesRenderCache() {
         for (auto it : _imageCache) {
@@ -110,6 +110,10 @@ wxString FacesEffect::GetEffectString() {
 
     ret << "E_CHOICE_Faces_Eyes=";
     ret << p->Choice_Faces_Eyes->GetStringSelection().ToStdString();
+    ret << ",";
+
+    ret << "E_CHOICE_Faces_EyeBlinkFrequency=";
+    ret << p->Choice_Faces_EyeBlinkFrequency->GetStringSelection().ToStdString();
     ret << ",";
 
     ret << "E_CHOICE_Faces_FaceDefinition=";
@@ -353,6 +357,7 @@ void FacesEffect::SetDefaultParameters() {
     SetRadioValue(fp->RadioButton1);
     SetChoiceValue(fp->Choice_Faces_Phoneme, "AI");
     SetChoiceValue(fp->Choice_Faces_Eyes, "Auto");
+    SetChoiceValue(fp->Choice_Faces_EyeBlinkFrequency, "Normal");
     SetChoiceValue(fp->Choice1, "");
 
     if (fp->Face_FaceDefinitonChoice->GetCount() > 0) {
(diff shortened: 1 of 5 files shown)


xLights/effects/FacesEffect.cpp
```diff
@@ -1084,8 +1131,17 @@ void FacesEffect::RenderFaces(RenderBuffer& buffer,
     } else if (phoneme == "rest" || phoneme == "(off)") {
             if ("Auto" == eyes) {
                 if ((buffer.curPeriod * buffer.frameTimeInMs) >= cache->nextBlinkTime) {
-                    // roughly every 5 seconds we'll blink
-                    cache->nextBlinkTime += intRand(4500, 5500);
+                    //calculate the blink time, taking into account user selection
```
MrPierreB, line 1134: So this code is repeated 5 times in this file,  would be better as a function.

xLights/effects/FacesPanel.cpp
```diff
@@ -104,7 +108,20 @@ FacesPanel::FacesPanel(wxWindow* parent) : xlEffectPanel(parent)
 	Choice_Faces_Eyes->SetSelection( Choice_Faces_Eyes->Append(_("Auto")) );
 	Choice_Faces_Eyes->Append(_("(off)"));
 	FlexGridSizer98->Add(Choice_Faces_Eyes, 1, wxALL|wxALIGN_LEFT|wxALIGN_CENTER_VERTICAL, 5);
-	FlexGridSizer98->Add(-1,-1,1, wxALL|wxALIGN_CENTER_HORIZONTAL|wxALIGN_CENTER_VERTICAL, 5);
+
```
MrPierreB, line 111: The indent spacing are different

MrPierreB (commented):
If you made changes to UI then I am not seeing the wxwidgets file in this PR.  Your changes will get lost.

AzGilrock:
Yeah like Pierre says we don't want the effect panel code to be modified manually.  You need to learn how to bring up the panel in CodeBlocks and use the wxSmith designer to make changes to the panel.  The .wxs file needs to contain the changes so that next time someone else makes a change and regenerates the panel code it doesn't blow away your manual edits.

xLights/effects/FacesEffect.cpp
kylegrymonprez: I can do this - I just copied the existing code pattern but can optimize as you say

kylegrymonprez:
> Yeah like Pierre says we don't want the effect panel code to be modified manually. You need to learn how to bring up the panel in CodeBlocks and use the wxSmith designer to make changes to the panel. The .wxs file needs to contain the changes so that next time someone else makes a change and regenerates the panel code it doesn't blow away your manual edits.

Okay - as I'm on Mac I wasn't aware that I had to go into CodeBlocks - I was wondering why variable names were so random. 

I'll figure out how to use codeblocks then - I think early on I was trying to get it working on my Ubuntu box....

AzGilrock:
Yeah I use a Mac but I Remote Desktop back into my Windows machine to do CodeBlocks work.  I'm not sure how Dan does his CodeBlocks work because he's a Mac guy also.

xLights/effects/FacesEffect.cpp
kylegrymonprez: addressed comment

xLights/effects/FacesPanel.cpp
kylegrymonprez: removing the manual edits

kylegrymonprez:
I actually had set this up previously on a ubuntu box, just didn't think about using it to update the panel - now that I see there's a visual editor, the string string variable naming conventions made a whole lot more sense.

kylegrymonprez:
Updated the code per comments. Used the visualizer to update the .wxs file and refactored the max delay logic to a function.

PR #4104, opened 2023-11-06
Enable the Controller tab's Activate etc menu items when multiple controllers are selected

degracode:
This affects:
* Activate
* Activate in xLights
* Inactivate
* Unlink From Base

Previously these menu items were only enabled when a single controller was selected. I've updated the logic to allow multi-select too. The menu items' actions already coped with handling multiple devices.

image


degracode:
Any thoughts on this @dartunder @AzGilrock ?

computergeek1507:
It would be nice to also implement a way to bulk edit controller properties like protocol, vender/model/varient

Issue #4134, opened 2023-11-17
Hang intermittently when removing Display Elements


jkriegshauser:
**Describe the bug**
The app can deadlock intermittently when removing unused elements from the Display Elements menu.

**To Reproduce**
1. Open a sequence.
2. On the timeline, select a display element group (i.e. All House) and right click > Model > Delete Effects.
3. Open the Display Elements window. In the Added: box, right click and Remove Unused.
4. ~25% of the time, the app will hang.

**Expected behavior**
Unused display elements should be removed and the app should not hang.

**Screenshots**

**Versions (please complete the following information):**
 - OS: Windows 11
 - xLights version: 2023.17

**Analysis**
Building locally and attaching a debugger, I can see the deadlock behavior.
The main thread  is attempting to take a lock in the `~ModelElement()` destructor:
```
xLights.exe!std::unique_lock<std::recursive_timed_mutex>::{ctor}(std::recursive_timed_mutex &) Line 133
xLights.exe!ModelElement::~ModelElement() Line 492
xLights.exe!SequenceElements::DeleteElement(const std::string & name) Line 439
xLights.exe!ViewsModelsPanel::RemoveSelectedModels() Line 621
xLights.exe!ViewsModelsPanel::RemoveUnusedModels() Line 1532
xLights.exe!ViewsModelsPanel::OnModelsPopup(wxCommandEvent & event) Line 1479
xLights.exe!wxAppConsoleBase::CallEventHandler(wxEvtHandler * handler, wxEventFunctor & functor, wxEvent & event) Line 639
xLights.exe!wxEvtHandler::ProcessEventIfMatchesId(const wxEventTableEntryBase & entry, wxEvtHandler * handler, wxEvent & event) Line 1401
...
```

The lock is owned by a different thread that is blocked in a condition variable waiting for a frame:
```
xLights.exe!std::condition_variable::wait_for(std::unique_lock<std::mutex> &) Line 619
xLights.exe!NextRenderer::waitForFrame(int frame) Line 125
xLights.exe!RenderJob::Process() Line 
xLights.exe!RunInAutoReleasePool(std::function<void __cdecl(void)> &&) Line 53
xLights.exe!JobPoolWorker::ProcessJob(Job * job) Line 298
xLights.exe!JobPoolWorker::Entry() Line 241
xLights.exe!startFunc(JobPoolWorker * jpw) Line 88
```

[...]

PR #4135, opened 2023-11-17
Fix for #4134

jkriegshauser:
Change the memory management model for Element and derivative classes to be ref-counted through the `shared_ptr` mechanism. This removes the need for the `Element` destructor to lock a mutex which can lead to deadlocks.

### DIFF
--- a/xLights/SeqFileUtilities.cpp
+++ b/xLights/SeqFileUtilities.cpp
@@ -1124,7 +1124,7 @@ void MapXLightsStrandEffects(EffectLayer* target, const std::string& name,
 void MapXLightsEffects(Element* target,
                        const std::string& name,
                        SequenceElements& seqEl,
-                       std::map<std::string, Element*>& elementMap,
+                       std::map<std::string, std::shared_ptr<Element>>& elementMap,
                        std::map<std::string, EffectLayer*>& layerMap,
                        std::vector<EffectLayer*>& mapped, bool eraseExisting,
                        SequencePackage& xsqPkg, bool lock, const std::map<std::string, std::string>& mapping)
@@ -1138,7 +1138,7 @@ void MapXLightsEffects(Element* target,
     }
 
     EffectLayer* src = layerMap[name];
-    Element* el = elementMap[name];
+    auto el = elementMap[name];
 
     if (src != nullptr) {
         MapXLightsEffects(target->GetEffectLayer(0), src, mapped, eraseExisting, xsqPkg, lock, mapping);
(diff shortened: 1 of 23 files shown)


dkulp:
From my standpoint, this is way too high risk to be merging at this point in the year.   Plus, it's not the right fix for #4134 anyway.   Anything that makes material changes the layouts or master views needs to call AbortRender to make sure the render threads are stopped prior to making the changes.

computergeek1507:
I always think shared pointer is a crutch for bad lifetime management. It would be better to use unique pointer instead

xLights/SeqFileUtilities.cpp
```diff
@@ -2267,7 +2267,7 @@ void xLightsFrame::ImportSuperStar(const wxFileName& filename)
     for (size_t i = 0; i < _sequenceElements.GetElementCount(); i++) {
         if (auto p = _sequenceElements.GetElement(i))
             model = p->shared_from_this();
-        if (!model || model->GetType() == ElementType::ELEMENT_TYPE_MODEL) {
```
AzGilrock: That change is not equivalent to the existing code.   Why touch it?  Do you know how to test the SuperStar import?

jkriegshauser:
Hey guys, thanks for taking a look.

@dkulp: 
> From my standpoint, this is way too high risk to be merging at this point in the year.

Definitely high risk, and definitely needs testing. I'm doing sequencing with it this season, but I totally get wanting to hold off. It definitely needs people trying it out. Multi-threaded code is non-deterministic by design and needs a lot of time and coverage to prove it out.

> Plus, it's not the right fix for https://github.com/smeighan/xLights/issues/4134 anyway. Anything that makes material changes the layouts or master views needs to call AbortRender to make sure the render threads are stopped prior to making the changes.

I'll look into this further. I tried adding an `AbortRender()` call before the elements were deleted but still got a hang.

@computergeek1507 
> I always think shared pointer is a crutch for bad lifetime management. It would be better to use unique pointer instead

`std::shared_ptr` and `std::weak_ptr` are very useful in multithreaded systems since the ref-count changes are thread-safe and can allow jobs to be dispatched and ownership to be shared with the code that is dispatching them.

xLights/SeqFileUtilities.cpp
jkriegshauser: This should already be resolved in 5e93928

jkriegshauser:
My wife has been using this build since I created this pull request with no issues related to these code changes (and without the issue mentioned in #4134 either).

dkulp:
Closing this...   shared_ptr/weak_ptr is not a direction we'd like to go with this.

PR #4147, opened 2023-11-22
Fix HinksPix LR remote assignment for remotes 3-9 in visualizer

dartunder:
The original code prevented directly assigning a controller to remotes 3-9 using visualizer


dartunder:
There still appears to be an unresolved issue with cascaded remotes

dartunder:
Sorry, this doesn't go far enough to fix the issue.  Don't recommend merging. 
If I could determine if the event came from the cascade sub-menu then I could continue, not sure today how to get that info.
Feel free to fix yourself.

PR #4151, opened 2023-11-23
Fix for deadlock

jkriegshauser:
Fix for #4134 as a safer workaround as opposed to the Element memory management refactor in #4135


jkriegshauser:
Ping :)

dkulp:
Added the abort renders.  Other changes shouldn't be necessary.

jkriegshauser:
>  Other changes shouldn't be necessary.
condition_variable needs a condition that is checked to be used properly. Adding the `rendering` flag provides that condition

Issue #4154, opened 2023-11-24
Shift palette colors


ClutchplateDude:
Sometimes when I have a color palette selected but I want to insert a color at he beginning, I have to resort to putting it on the saved slot (sucks in Windows) and then get it back.

I'd like the ability to move the colors in the palette in either direction.

No alternatives come to mind

I have already implemented this and will submit PT for this.

derwin12:
That is what this is for..
![image](https://github.com/smeighan/xLights/assets/4643499/972ccad0-b003-420b-8297-3e28229fd005)
![image](https://github.com/smeighan/xLights/assets/4643499/33ce51ff-b98a-492e-987b-87012464a3a9)

ClutchplateDude:
No, that function reverses the palette. That's not what I need since many effects are sensitive to color order. I might have a color gradient setup in the first four slots, but want to add a white before it. Can't do that.

derwin12:
ClutchplateDude - did you want to close this ticket?

ClutchplateDude:
Oh, didn't realize it was still open. Closing.

PR #4155, opened 2023-11-24
Add support for Shifting color palette

ClutchplateDude:
This PR adds the ability to shift colors in either direction (in addition to just reversing it). It also adds three new icons to support this feature.
image

Fixes Issue #4154


keithsw1111:
I see no WXS file.

ClutchplateDude:
> I see no WXS file.

Oh. I was wondering how the UI is edited, but didn't see that.... my bad. What tool is used to create/edit these? Is this something I can do in Visual Studio on Windows? Let me look into that.

ClutchplateDude:
Ok, @keithsw1111, sorry about not knowing about wxsmith. I re-did the changes and they now have a wxs file.

dkulp:
Can I ask what those icons will look like in dark mode?

ClutchplateDude:
Currently they look the same, just like the Reverse button did before this change.
image
If there is a way to change these based on dark mode, I'm happy to add ones. I don't see any reference to dark mode in `BitmapCache.cpp`.

ClutchplateDude:
@dkulp @keithsw1111  Is there are dev email list or group where you all discuss code changes? Is there a doc for new contributors beyond the CONTRIBUTING.md on the repo that goes into more detail about the code base? As a new contributor there are things I've run into that might help others that I could document. 
Also, with a change like this, am I responsible for updating the manual? If so, how?

AzGilrock:
Nope there is no list or group.  We rarely discuss code changes anymore.  Years ago we used Trello but not sure anyone is active.  I think I had trouble finding it last time I looked.  There isn't much more than the readme's for each OS.  I don't even know who updates the user manual or how its managed.

Issue #4243, opened 2024-01-01
Data Entry issue with the Waveform Filter


derwin12:
Minor bug.
Right click on the waveform, choose custom waveform filter.
Enter say 30 for the first box
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/fc50fb35-e394-43e0-936e-9ec17a1fa06b)

Now go to the next value, via tab or mouse click
Enter 40.
As soon as you enter the 4, the first box changes to 4 :-(
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/84e15685-9d10-4740-8b0e-451d6f0113b9)

Sure it is something simple bug .. maybe I will find it .. (some day)
Daryl.

PR #4249, opened 2024-01-05
Entry of Note Range High Value resetting the low value #4243

derwin12:
Allow entry of Note Range High Value without affecting the Low Note value


computergeek1507:
Think you messed this up

derwin12:
I think I did .. lol ..

PR #4250, opened 2024-01-05
#4243 Dont alter the Low Midi Note entered when entering the High Midi Note range

derwin12:
When entering the values (not the slider) don't reset the low value.

(First PR - it ain't pretty) .. if not accepted. That's fine, still learning.


computergeek1507:
probably should make it consistent and do the same for setting the low value ie lines 146-150

keithsw1111:
I dont like this as it only seems to update when you are closing the dialog which can set values other than what were on the screen when you clicked the ok button.
A better answer may be to do it on exit and then maybe invalidate the ok button until the high and low are appropriate.

PR #4251, opened 2024-01-05
Enhancement: When a ruler is defined, include real world coords in model export.

tygunn:
…he selection using "real world" units as well in the text.

Purpose: Many users who take advantage of the 3d view in xLights use CAD or photogrammetry models of their house to layout their displays.  A user may wish to lay out their display entirely in xLights and then translate the position of models to the real world for setup.  If the user adjusts their display so that the world origin in xLights corresponds to a known reference point in the real world (e.g. the corner of a lot) the real world coordinates of the props in xLights can be used to lay out their display.  Exporting the X/Y/Z coordinates of models helps when laying out the props in the real world.

Change: When the user has defined a ruler to map between the xLights coordinate system and a known distance in the real world, add 3 new columns to the model export spreadsheet containing the "real world" X/Y/Z coordinates of each model.

Here's an example of of this will look like:
Example export

### DIFF
--- a/xLights/xLightsMain.cpp
+++ b/xLights/xLightsMain.cpp
@@ -103,6 +103,7 @@
 #include "effects/StateEffect.h"
 #include "graphics/opengl/xlGLCanvas.h"
 #include "models/ModelGroup.h"
+#include "models/RulerObject.h"
 #include "models/SubModel.h"
 #include "models/ViewObject.h"
 #include "outputs/ControllerEthernet.h"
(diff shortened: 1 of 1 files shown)


AzGilrock:
Can't be that many people that are crazy enough to be out there with a tape measure and their laptop with xlights open to view the Status bar when placing props...lol.

tygunn:
In my case I just take a printout outside and use they to position things.
 So I'm definitely not crazy enough to haul a laptop around the yard, but a
paper copy?   Done it.

Perhaps the model export to CSV is a better spot for this data?

I'm fine keeping my own patch it that still seems too OCD.

AzGilrock:
It doesn't bother me either way that was kinda my point was whether having it on the status bar was useful.  Does it help you to see those measurements while working in xLights?

tygunn:
I suspect exporting and carrying a printout is likely a better way to go.
In my case I design on the computer and go to lay things out in the yard.

I'll occasionally tweak a prop position as I lay it out in the field since
seeing things on the screen is not the same as seeing them in reality.  But
that is the minority use case.

Thanks for the feedback.  I'll look at just exposing this in the export --
that would be a lot less manual work to note the locations down too.

tygunn:
Adding this to the "export models" works out quite well, much cleaner than the status text.  Let me know if this makes sense.

BTW, kudos on the overall structure of the xLights code; changes like this are trivial to implement given the current design of the code.

Issue #4247, opened 2024-01-05
Blank/Unlabeled Phoneme Results in Red Box for Graphical Faces


smtpboy:
**Describe the bug**
Phonemes should always have a label associated with them.  I recently purchased a sequence that had an unlabelled phoneme, probably from a manual edit, resulting in a "red box" during the render because the label didn't match a defined face phoneme.  Adding a label to the phoneme fixes this, but it shouldn't really be allowed to happen.  This likely wouldn't be noticeable on a non-rendered physical prop like a talking tree, but since I was running this on a virtual matrix with the rendered Blue Bulb, it became very noticeable and was quite hard to troubleshoot.

**To Reproduce**
Generate voice/word/phoneme tracks.  Edit and blank out the label on a particular phoneme, it will then render as a red box.

**Expected behavior**
xLights should generate an error or prevent you from having an unlabelled phoneme.  Would be nice if the render process or check sequence flagged the error across the board instead of a red box.

**Screenshots**
<img width="1430" alt="Screenshot 2024-01-05 at 12 13 34 PM" src="https://github.com/xLightsSequencer/xLights/assets/71394162/bfb2b953-eeed-4c2d-a017-dcfe4549a98f">
<img width="1430" alt="xlights screenshot 1" src="https://github.com/xLightsSequencer/xLights/assets/71394162/89eaccec-cab2-4bf1-a3a3-5e0f46e28786">

**Versions (please complete the following information):**
 - OS: MacOS 13.6.3
 - xLights version [e.g. 2019.37] 2023.23.1

smtpboy:
Looks like this was fixed with the commit of #4252 so I'm going to close it out.  Thanks @derwin12 !

PR #4252, opened 2024-01-06
Use rest for missing face phoneme

derwin12:
Default to rest if no phoneme is set #4247


derwin12:
Okay .. so this branch commit had one line in it, but the PR shows older commits. What am I doing wrong?

Issue #4140, opened 2023-11-19
Effect Preset - Option to turn off previews


jefft0825:
The effect preset box severely slows down any sort of use of presets when it shows the preview of the effect.    Can we add a button to disable the preview?

derwin12:
Thought .. can you rename/delete the presets directory then change the attributes on the presets preview folder to read only?
(not to say the devs could put in your request)

cybercop23:
Avail in 2024.2. @jefft0825  please close.

PR #4262, opened 2024-01-13
#4140 Add preference to skip the loading of preset gif images.

derwin12:
Requested for folks that have a large number of presets. 
 The image is still created when an effect is added - it is simply not loaded when the preset tree is read.
 Preference added to the View panel. Default to false (off) ie it will be shown.


derwin12:
I need to repush - seems my tabs are being translated to spaces?

AzGilrock:
Yep tabs in code are usually frowned upon so I believe Git is converting them to spaces.

AzGilrock:
You can usually setup your code editor to insert spaces when you hit the tab key.

computergeek1507:
If you enable clang format in visual studio it should use the clang format file that defaults to spaces

derwin12:
> If you enable clang format in visual studio it should use the clang format file that defaults to spaces

Something like this?
image

computergeek1507:
You don't need power tools, it's built into vs2022 now

computergeek1507:
image

derwin12:
So I have this .. (your pic was cut off) - is it correct?
image
Also the profiles .. should I remove these bottom 3 or are they okay to add (seems like they should stay as they are the right thing to do)
image
image

derwin12:
Now that I figured out how code::blocks gets involved here .. I updated the cb code and it injected into the commit.

Issue #4244, opened 2024-01-02
Shape Effect - Y Center is reversed when emoji is selected


derwin12:
Drop a shape effect on a matrix or whatever. Put a value curve on the Y center - I used a ramp from 0 to 100.
If you use a built in shape it moves from bottom (0) to top (100).

If you switch to an emoji, the movement is reversed  top (0??) to bottom (100??).

![image](https://github.com/xLightsSequencer/xLights/assets/4643499/9d2df1c9-5774-4523-a2a7-74c5cba20e4e)

PR #4263, opened 2024-01-13
4244 Shape Effect - Flip the Y value for Emoji 

derwin12:
The Y value was reversed for the emoji shape. ie a 0 was at the top, and 100 was at the bottom. This is the opposite to all the other objects in the shape effect. Corrects the results when using a value curve. No issue with x value.

Issue #4244

### DIFF
--- a/xLights/effects/ShapeEffect.cpp
+++ b/xLights/effects/ShapeEffect.cpp
@@ -1210,6 +1210,30 @@ void ShapeEffect::Drawpresent(RenderBuffer& buffer, int xc, int yc, double radiu
     }
 }
 
+bool ShapeEffect::needToAdjustSettings(const std::string& version)
+{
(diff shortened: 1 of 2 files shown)


computergeek1507:
You will have to write update code, this will break all sequences already using this setting

derwin12:
Was wondering about that - I dont expect that it was used that much (being a single emoji that follows a y value) but it is working the correct way now (that it has been reversed). What is an example of an update code and how would that be implemented?

AzGilrock:
Look at the PInwheel effect.  The needToAdjustSettings will specify the xlights version where the change happened and the adjustSettings function would contain the code to upgrade effects that were created in older versions of xlights.

Screenshot 2024-01-25 at 6 41 52 PM

Issue #4255, opened 2024-01-12
Expand drop downs to accommodate longer entries


kevinsaucier:
Just a minor annoyance, but still annoying.  😁  As the title says, some drop downs are way too small and cut off text that is required in order to choose the right entry.  For instance, the default view.    I could have sworn I'd seen more, but I'm not finding them now.  I'm thinking maybe on some of the Effect Settings.  🤔

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/c797af42-468c-4f86-a171-7e3eb755d88d)

AzGilrock:
So it looks like derwin implemented this one.

kevinsaucier:
I see it a bit bigger in 2024.02, but it's still pretty small (at least for a couple of my views).  Granted, most views probably aren't going to need it, but the space is there so why not use a bit more. 😁

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/9234833c-8439-4095-87df-19540114a03c)

AzGilrock:
Ok I'll add the wxExpand property.  Hopefully that will be big enough.  Doesn't look like we let you change the dialog size but the combo box will expand to the size of the surrounding panel.

<img width="858" alt="Screenshot 2024-02-01 at 3 47 54 PM" src="https://github.com/xLightsSequencer/xLights/assets/2024669/aa76974e-6395-41ad-810a-796eb7f1cb92">

kevinsaucier:
Sounds like a plan.  Thanks!

PR #4264, opened 2024-01-13
#4255 Increase size of Preference Default View Dropdown

derwin12:
Also added a couple of tooltips while I was in there.

Need to figure out how to match the prior spacing. It sure looks like tabs. Perhaps this is just old code that was different tab sizing? 

Tsk #4255


derwin12:
Going to repush using codeblocks code...

Issue #4115, opened 2023-11-10
Faces Download (from xlights repo) is failing for some faces..


derwin12:
Is it something I did :-)
From Zoom.. user is trying to download sailor and I tried Santa (04) and both failed.
Repo:
says 
<artist>Daryl Erwin</artist>
<facelink>https://nutcracker123.com/xlights/faces/Sailor.zip</facelink>

The zip is valid on here
https://nutcracker123.com/xlights/faces/

Also tried Santa
https://nutcracker123.com/xlights/faces/Santa04.png

```
2023-11-10 12:48:23,704 10172 log_base [DEBUG] File Cache downloading file https://nutcracker123.com/xlights/faces/Santa04.png.
2023-11-10 12:48:23,705 10172 log_base [DEBUG] Making request to 'https://nutcracker123.com/xlights/faces/Santa04.png' -> C:\Users\daryl\AppData\Local\Temp\xlFEFA.png.
2023-11-10 12:48:26,138 10172 log_base [DEBUG] File Cache downloading file https://nutcracker123.com/xlights/faces/Santa04.zip.
2023-11-10 12:48:26,138 10172 log_base [DEBUG] Making request to 'https://nutcracker123.com/xlights/faces/Santa04.zip' -> C:\Users\daryl\AppData\Local\Temp\xl881.zip.
2023-11-10 12:48:27,018 10172 log_base [DEBUG] Extracting C:\Users\daryl\AppData\Local\Temp\xl881.zip:Santa04\ to E:\2023\TestFolder\DownloadedFaces\Santa04\.
2023-11-10 12:48:27,018 10172 log_base [ERROR] File extract failed.
2023-11-10 12:48:27,018 10172 log_base [DEBUG] Extracting C:\Users\daryl\AppData\Local\Temp\xl881.zip:Santa04\Santa04-ai-ec.png to E:\2023\TestFolder\DownloadedFaces\Santa04\Santa04-ai-ec.png.
2023-11-10 12:48:27,018 10172 log_base [ERROR] File extract failed.
2023-11-10 12:48:27,018 10172 log_base [DEBUG] Extracting C:\Users\daryl\AppData\Local\Temp\xl881.zip:Santa04\Santa04-ai-eo.png to E:\2023\TestFolder\DownloadedFaces\Santa04\Santa04-ai-eo.png.
2023-11-10 12:48:27,018 10172 log_base [ERROR] File extract failed.
[...]

PR #4266, opened 2024-01-14
#4115 Handle the directory if found in a faces zip file

derwin12:
Some face zip files available from within xlights download, fail since they have a directory embedded into the zip file.
This handles that and creates the directory without error and completes the face creation.
A few lines added at 1209

Can Scott verify that all this clang formatting is indeed correct - perhaps I am the last person to touch this file with clang format?

Issue #4115


derwin12:
Will resubmit without the formatting changes..

Issue #3873, opened 2023-07-09
VU Meter - Type: Level Random Bar Bug


marklancer:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
When using the "Level Random Bar" type for the VU meter the last bar does not triggered.  

**To Reproduce**
See attached sequence

**Expected behavior**
All bars should be triggered at some point when using the effect.
In the attached sequence  when using VU meter types Timing Event Bar, Timing Event Random Bar or Level Bar each snowflake is triggered at some point.  When using Level Random Bar the last snowflake is never triggered.  I have confirmed the same behavior exists using different amount of props and in every case the last prop is not triggered.

**Screenshots**

**Versions (please complete the following information):**
 - OS: Windows 11
 - xLights version [2023.10]

**Attachments**

[LevelRandomBarBug.zip](https://github.com/smeighan/xLights/files/11994052/LevelRandomBarBug.zip)

PR #4268, opened 2024-01-14
3873  VU Meter Random bars missed last bar

derwin12:
Because of rounding the final bar (slice) would be skipped. Same issue on Random Note and Random Timing Track
Lines 2552-2555 and 2 other cases.
Formatting also updated via the commit.

Bug #3873


computergeek1507:
IDK why all tab/spaces are incorrect but please fix it before spamming us with more pull requests,

derwin12:
Can you pull in the branch and see if it flips back? Ie .. am I just
getting hit on older files and the commit is correct?

derwin12:
Closing this and creating new PR with no code format fixups.

Issue #3826, opened 2023-05-31
Custom default path for sequences in Preferences


DoctorWiz:
Just as it is possible to change the default folder for .fseq files and the render cache, it would be nice to have an program option for a default folder for sequences.
Currently, when doing an "open" it starts by default browsing in the main show folder and likewise when doing a save or save as it defaults to the show directory.
It would be nice to keep sequences (and any other specialized file types) in their own subfolders instead of cluttering up the show folder.
Mock up attached.
![SequenceDirectoryOption](https://github.com/smeighan/xLights/assets/20911920/23e689cf-5f7f-4565-9f64-58569d07108f)

derwin12:
Devs -- thoughts no this? Is this okay to work on?

AzGilrock:
I already save every sequence in its own subfolder so you don't need these changes to do that.  This seems like its stuffing all the sequences in another folder just not the show folder.  Now if this option created a new folder under that folder for every sequence I'd be all for it.

derwin12:
https://github.com/xLightsSequencer/xLights/issues/4302

PR #4281, opened 2024-01-19
New preference to set a default sequence folder

derwin12:
Issue #3826 Adds a preference to store your sequences. Seems many folks are putting sequences in a subfolder off their show folder. This is great but when you forget, you end up having 2 copies and not knowing which one is which.
The default is unchecked - so retaining the show folder as their default.

### DIFF
--- a/xLights/TabSequence.cpp
+++ b/xLights/TabSequence.cpp
@@ -307,8 +307,6 @@ wxString xLightsFrame::LoadEffectsFileNoCheck()
 
     //Load FSEQ and Backup directory settings
     fseqDirectory = GetXmlSetting("fseqDir", showDirectory);
-    renderCacheDirectory = GetXmlSetting("renderCacheDir", fseqDirectory); // we user fseq directory if no setting is present
-    ObtainAccessToURL(renderCacheDirectory);
     ObtainAccessToURL(fseqDirectory);
     if (!wxDir::Exists(fseqDirectory)) {
         logger_base.warn("FSEQ Directory not Found ... switching to Show Directory.");
@@ -317,6 +315,16 @@ wxString xLightsFrame::LoadEffectsFileNoCheck()
         UnsavedRgbEffectsChanges = true;
     }
     FseqDir = fseqDirectory;
+    xsqDirectory = GetXmlSetting("xsqDir", showDirectory);
+    ObtainAccessToURL(xsqDirectory);
+    if (!wxDir::Exists(xsqDirectory)) {
+        logger_base.warn("Sequence XSQ Directory not Found ... switching to Show Directory.");
+        xsqDirectory = showDirectory;
+        SetXmlSetting("xsqDir", showDirectory);
+        UnsavedRgbEffectsChanges = true;
+    }
+    renderCacheDirectory = GetXmlSetting("renderCacheDir", fseqDirectory); // we user fseq directory if no setting is present
+    ObtainAccessToURL(renderCacheDirectory);
     if (!wxDir::Exists(renderCacheDirectory)) {
         logger_base.warn("Render Cache Directory not Found ... switching to Show Directory.");
         renderCacheDirectory = showDirectory;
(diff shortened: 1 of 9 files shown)


dkulp:
This would need some validation to make sure the selected folder is a subdirectory of the show folder.    Too much stuff assumes that sequences are stored within the show folder directory tree.

computergeek1507:
Not a fan of this as now the Sequence directory is a global setting for all show folders, people could easily open wrong sequences all the time.

derwin12:
Not clear to me how this is different than someone saving a sequence anywhere on their hard disk. Nothing stopping you from doing that now. I can add a check to be under the show folder though.

derwin12:
Added check for sequence folder is within show folder.

computergeek1507:
Not a fan of it being global, people will change show folders, but then open sequence in other show folders, get model errors and be confused. I think folder settings should be only relative to the show folder if global or be shored in the show folder.

derwin12:
> Not a fan of it being global, people will change show folders, but then open sequence in other show folders, get model errors and be confused. I think folder settings should be only relative to the show folder if global or be shored in the show folder.

I am not clear what you are asking/saying here. The use of a global variable? The folder chosen is now forced to be within the show folder. What code change are you suggesting?

dkulp:
> > Not a fan of it being global, people will change show folders, but then open sequence in other show folders, get model errors and be confused. I think folder settings should be only relative to the show folder if global or be shored in the show folder.
> 
> I am not clear what you are asking/saying here. The use of a global variable? The folder chosen is now forced to be within the show folder. What code change are you suggesting?

The xsqLinkFlag is "global" and stored in the xLights global config and not part of the show configuration.    You likely don't even need that flag at all.   If the sequence directory is empty or equal to the show director, assume it's "linked" and check the box and disable the entry field when updating the preference panel.

derwin12:
Removed the global.. Did some testing with changing temp and permanent folders. If it fails to validate, it does revert back to the show folder. It also checks that you are in the show folder directory.
I will create a new issue to address fseq and xsq directory and rendercache folder to make them relative so as you switch folders it will look to see if they exist relative in the new show folder. That will be a bigger task.
Do we want a ticket to stop people from saving .xsq (or perhaps just warn?) when they save outside of the showfolder?

cybercop23:
Can this please be merged? Or are there any other ideas/suggestions?

xLights/TabSequence.cpp
```diff
@@ -317,6 +315,22 @@ wxString xLightsFrame::LoadEffectsFileNoCheck()
         UnsavedRgbEffectsChanges = true;
     }
     FseqDir = fseqDirectory;
+    xsqDirectory = GetXmlSetting("xsqDir", showDirectory);
+    ObtainAccessToURL(xsqDirectory);
+    if (!wxDir::Exists(xsqDirectory)) {
```
AzGilrock, line 320: So if I read this right if the setting the user entered for where they want their XSQ files stored cannot be found then its going to just change the registry?  Seems like some folks are going to use cloud drives which could potentially have connection issues and they wouldn't want their setting changed due to connection issues.

Also the second if statement is always true if the code goes into the first if statement so it could probably be an "else if".

cybercop23:
Been waiting for this. Any reason why this can't be merged... please....

AzGilrock:
> Been waiting for this. Any reason why this can't be merged... please....

I meant to do some testing and I was kinda waiting to see if Dan or Scott had any more feedback since they've both commented here.

cybercop23:
https://images.app.goo.gl/jo24Wqf4qpV3uCbf9

cybercop23:
[https://images.app.goo.gl/jo24Wqf4qpV3uCbf9](url)

dkulp:
Needs to be re-based.

cybercop23:
@derwin12 do you mind doing the rebase for this, please?

cybercop23:
Per Dan above, need a rebase.

Issue #3030, opened 2021-12-23
Copy/Paste Model & Row Effects Strip First Slash Of Image Paths On Network Locations


dream-encode:
**Describe the bug**
When using a Network Share, copying the Image effect using Model/Row -> Copy Effects then pasting with Model/Row -> Paste Effects is stripping the first `\` from the full path.

**To Reproduce**
1. Add an Image effect with a path to a Network Share, i.e. `\\NAS\Lights\`
2. Model -> Copy Effects
3. Model -> Paste Effect on another model.
4. First `\` is stripped.

**Expected behavior**
The first `\` shouldn't be stripped.

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version 2021.40

https://user-images.githubusercontent.com/1900750/147279619-20ea0ac7-7a09-4aaf-bed0-346528ca5de4.mp4

PR #4285, opened 2024-01-21
Handle pasting network drive files

derwin12:
Issue #3030 Devs - this works but is there any downside? 
For the paste function it uses the option to ignore escape characters so that the leading slashes don't get dropped during the paste. 
"\\DERWIN-PC\c\Users\derwin\Downloads\istockphoto-1213006169-612x612.jpg"

ie. wxArrayString efdata = wxSplit(all_efdata[1], '\t', wxT('\0'));

### DIFF
--- a/xLights/sequencer/EffectsGrid.cpp
+++ b/xLights/sequencer/EffectsGrid.cpp
@@ -4927,7 +4927,7 @@ Effect* EffectsGrid::Paste(const wxString& data, const wxString& pasteDataVersio
 
     logger_base.info("Pasting data: %s", (const char*)data.c_str());
 
-    wxArrayString all_efdata = wxSplit(data, '\n');
+    wxArrayString all_efdata = wxSplit(data, '\n', wxT('\0'));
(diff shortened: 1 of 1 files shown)


AzGilrock:
So it leads me to believe its having the effect of disabling what is considered an escape character.  Do we have any instances where we have URL's in effects that could use escape characters that could be impacted?

derwin12:
> So it leads me to believe its having the effect of disabling what is considered an escape character. Do we have any instances where we have URL's in effects that could use escape characters that could be impacted?

Yes .. that is my understanding/intention. it disables looking for escape characters other than the \t or whatever the split character is. I couldn't think of a case where we wanted it to do anything on an escape character even if it were in the string. It should treat everything as it comes in from the cut. No processing other than the split on \t. If there is a test case scenario, I am willing to try it out.

AzGilrock:
Yeah its probably fine.  I'm sure some user will let us know if it isn't. :)

Issue #4056, opened 2023-10-16
Import Mapping - # of effects shows the # including all the submodels


derwin12:
When importing a sequence the # of effects is showing the total number including all the submodels. It also shows all the submodels and their totals. This is misleading. The # for the main model should just be the number of effects on that model itself.

Eg. Showstopper Snowflake might show 10 - you go and map that and nothing appears on your sequence. Turns out all the effects are on the submodels which were did NOT mapped over :-(

derwin12:
Devs - I chased this down to **SeqFileUtilities.cpp** line 1253 during Import XLights procedure. 
It calls **int Element::GetEffectCount() const** which sums all the effects on models+subs+strands
Can this be coded with some C++ trickery, to go to this function?
**int Element::GetEffectCount() const** which only counts effects on the actual model.
I did some crude testing, and I dont see any downside to showing this effect count but I cant figure out how to modify the code.
I could create a 3rd function but that seems like overkill when it is right there..

Thoughts? Even if the change isn't acceptable, how would this be done?

![image](https://github.com/xLightsSequencer/xLights/assets/4643499/ca55a630-0bce-4005-875a-bd9e968e4d60)

PR #4294, opened 2024-01-25
#4056 Number of effects listed on Import Sequence panel

derwin12:
The number of effects shown for a model (right side) included all the submodels. If one were to map over that model, it potentially could have way less impact then you would think - as the effects are all on the submodels.
This PR only shows the number of effects directly on the model (left side).
Thoughts on the implementation? I created a new method.
2024-01-24 20_14_52-xLights (Ver 2023 23 64bit) Dec 28 2023

### DIFF
--- a/xLights/sequencer/Element.cpp
+++ b/xLights/sequencer/Element.cpp
@@ -595,6 +595,18 @@ bool ModelElement::HasEffects() const
     return false;
 }
 
+int ModelElement::GetModelEffectCount() const
+{
+    int sum = std::accumulate(
+        mEffectLayers.begin(),
+        mEffectLayers.end(), 0,
+        [](int i, EffectLayer* l) {
+            return l->GetEffectCount() + i;
+        });
+
+    return sum;
+}
+
 int ModelElement::GetEffectCount() const {
 
     int sum = std::accumulate(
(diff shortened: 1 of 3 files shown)


xLights/sequencer/Element.cpp
```diff
@@ -595,6 +595,18 @@ bool ModelElement::HasEffects() const
     return false;
 }
 
+int ModelElement::GetModelEffectCount() const
+{
+    int sum = std::accumulate(
+        mEffectLayers.begin(),
+        mEffectLayers.end(), 0,
+        [](int i, EffectLayer* l) {
```
AzGilrock, line 603: Is this a lambda function?  Wish I understood this new stuff.   I saw someone use a lambda at work and I looked it up and still don't understand it...lol.

merryoncherry: Yep just an anonymous function written right there - not that hard to understand.  std::accumulate takes a function among other things... the begin/end of the range to accumulate, etc.

That syntax looks weird at first but it is also not complicated.
[captures - none here - but it would be stuff the function needs that is not present in the parameters, but rather in the calling context](parameters/arguments){code body that does stuff on the parameters}

Issue #4230, opened 2023-12-19
Video effect sometimes glitchy - gets frames out of order causing stuttering


merryoncherry:
**Describe the bug**
The video effect sometimes stutters, seeming to switch back and forth between frames rapidly.

**To Reproduce**
Creating a 40FPS sequence and rendering any video that has a non-integral frame time in ms can be used to reproduce this, say a 24FPS or 30FPS video. 
As a specific example, in a layout with a megatree, create a 40FPS sequence, then import the tree from the following free sequence (it uses a video that is 20.01FPS):
https://eflsequences.com/?product=bluey-theme-song-remix

**Expected behavior**
Smooth video playback (within the limitation of the FPS of the source sequence).

**Versions (please complete the following information):**
 - OS: Windows 10 / 11; software video decoding
 - xLights 2023.22 / main

**Additional context**
The issue is caused by the way the code accounts for the current frame and the previous frame.  It believes the previous frame covers a narrower time interval than it actually does because it rounds down to an integral number of ms, causing it to choose a frame in the future instead of the right one.

**Attachments**
Patch:
```
index e0b518619..2291dcf98 100644
--- a/xLights/VideoReader.cpp
+++ b/xLights/VideoReader.cpp
@@ -841,7 +841,7 @@ AVFrame* VideoReader::GetNextFrame(int timestampMS, int gracetime)
         //same frame, just return
         return _dstFrame;
     }
-    if (timestampMS >= timeOfPrevFrame && timestampMS < currenttime) {
+    if (timestampMS >= timeOfPrevFrame - 1 && timestampMS < currenttime) {
         //prev frame, just return, avoids a seek
         return _dstFrame2;
     }
@@ -865,7 +865,8 @@ AVFrame* VideoReader::GetNextFrame(int timestampMS, int gracetime)
         bool seekedForward = false;
                while (!_abort && (firstframe || ((currenttime + (_frameMS / 2.0)) < timestampMS)) &&
                currenttime <= _lengthMS &&
-               (av_read_frame(_formatContext, _packet)) == 0) {
+               (av_read_frame(_formatContext, _packet)) == 0)
+        {
             // Is this a packet from the video stream?
                        if (_packet->stream_index == _streamIndex) {

@@ -927,17 +928,12 @@ AVFrame* VideoReader::GetNextFrame(int timestampMS, int gracetime)
                return nullptr;
        } else {
         int currenttime = GetPos();
[...]

PR #4296, opened 2024-01-26
Fix stuttering in video effect #4230

merryoncherry:


xLights/VideoReader.cpp
```diff
@@ -841,7 +841,7 @@ AVFrame* VideoReader::GetNextFrame(int timestampMS, int gracetime)
         //same frame, just return
         return _dstFrame;
     }
-    if (timestampMS >= timeOfPrevFrame && timestampMS < currenttime) {
+    if (timestampMS >= timeOfPrevFrame - 1 && timestampMS < currenttime) {
```
AzGilrock, line 844: Would be useful to have a comment here so we know why the "-1" is needed.

merryoncherry: So it's because of the pervasive, awful assumption that the frame rate can be captured as an integer milliseconds per frame.  With the exception of 20, 25, 40, and a few others, it can't, and those are hardly ever used in videos... lots of 24 and 30 and so on there.  So timeOfPrevFrame, in ms, is potentially off by 1... no harm in giving it a fudge factor of 1.

AzGilrock: Yeah I meant add a comment in the code.  A year from now a developer running into this won't know why its there.

merryoncherry: A comment in the xLights code?  AYS?
(I will happily put that comment in.)

AzGilrock: Yes a comment right above that line.  But it can be shorter...something like "subtracting -1 from previous frame due to integer roundoff issues for certain frame rates".  I'll merge it right after.

PR #4305, opened 2024-01-29
VU Meter - correct the bar overrun introduced in #3873

derwin12:
I incorrectly incremented the loop used to go thru the bars.


derwin12:
Correct .. the + 1, the x++ was just done for consistency with the others. Is ++x done prior to the loop vs after the loop? I wasnt positive on that.

AzGilrock:
The increment step is done at the end so for a "for" loop it doesn't really matter whether its pre increment or post increment.  But some have argued to used pre increment claiming its a couple instructions shorter.  I'm not sure that's the case these days.

Issue #3313, opened 2022-07-03
Layered arch model with non-zigzag wiring


merryoncherry:
**Is your feature request related to a problem? Please describe.**
I have a layered arch made out of strips in tubes, no data wire coming out of the far end of the tube.  The best wiring for this is to have all the connections near the controller, but the existing arch model has wiring order that "zigzags", alternating direction with each layer of the arch, so in this configuration the wire from the controller has to go to the far end of the arch for some of the layers.  

**Describe the solution you'd like**
I'd like a checkbox on the arch model property grid that provides a wiring that doesn't zigzag, so that all the layers connect from the same end.

**Describe alternatives you've considered**
Of course, as it needs multiple connections to the controller, and #2466 for multiple controller connections isn't implemented yet, the easiest way to support it is a custom model, which is what I'll be doing.  But the easiest way to generate that custom model was by exporting the layered arch with non-zigzag wiring.

**Additional context**
A more detailed description of my prop, and the arch model, is here:
https://merryoncherry.org/2022/05/22/5-layer-pex-arch/

I also have a patch that implements this feature as I've described it.  As I already exported the custom model I'm now using, my problem is already solved and I don't really need the feature in xLights, but if you think other people might want to do something similar, I can make a PR.

merryoncherry:
@cybercop23 Thanks for all the recent housecleaning.  You want the patch for this one or want it closed?

AzGilrock:
If you already had a solution I'd post the PR.

merryoncherry:
PR #4310

cybercop23:
I see waiting on MR merge. Can this be closed?

PR #4310, opened 2024-02-01
Option for arch model wiring to not zigzag (#3313)

merryoncherry:


xLights/models/ArchesModel.h
```diff
@@ -23,6 +23,7 @@ class ArchesModel : public ModelWithScreenLocation<ThreePointScreenLocation>
         virtual void InitRenderBufferNodes(const std::string &type, const std::string &camera, const std::string &transform,
                                            std::vector<NodeBaseClassPtr> &Nodes, int &BufferWi, int &BufferHi, int stagger, bool deep = false) const override;
         virtual int GetNumPhysicalStrings() const override { return 1; }
+        virtual bool StrandsZigZagOnString() const override { return zigzag; }
```
merryoncherry, line 26: This function seems not used and probably has some cases that are wrong.  I would be happy to remove it if desired.

xLights/models/ArchesModel.cpp
```diff
@@ -629,8 +644,15 @@ void ArchesModel::ImportXlightsModel(wxXmlNode* root, xLightsFrame* xlights, flo
         wxString psp = root->GetAttribute("PixelSpacing");
         wxString ls = root->GetAttribute("LayerSizes");
         wxString h = root->GetAttribute("Hollow");
+        wxString zz = root->GetAttribute("ZigZag");
 
         // Add any model version conversion logic here
+
+        if (zz == "")
```
AzGilrock, line 651: Should be able to just add "true" as a second parameter to the GetAttribute call and delete this whole if.

AzGilrock: Sounds like what happens to me at work.  I tell them "hey I don't create anything new I copy from what you guys already have in the code base and then you tell me its wrong"....lol.

For StrandsZigZagOnString yeah I'd delete it if you don't see it being used.  I did a quick search on GitHub and don't see the function actually used.  Doesn't seem to make sense for an entire model class to have it defined cause users are free to wire things however they like.

I was going to complain about the zigzag being stored as "true" instead of "1" because it takes more bytes but I searched and it seems like its been done that way elsewhere.

computergeek1507: Some developers like to store all the settings as strings to make it easier to view in the XML files. I'm not a fan of using strings as from a performance standpoint they are bad, they also are bad from a maintenance/"easy to create bug" standpoint. I prefer strongly typed variables like bool or int/enum for settings.

merryoncherry: XML files don't store anything else?  Maybe there are reasons why JSON is so popular for new projects.

AzGilrock: Yeah I thought everything gets turned into a string with the XML stuff then we convert it back when reading it in.  I mainly followed what was there when I started.  I think all the developers would like to get away from the way xLights used the "ModelXML" object carrying around XML and using it to read/write object status throughout the program.  I think we could gain a bit of speed if we read the XML or switch to JSON (which I'm not that familiar with) and then the object status is held in class object variables until we need to write back out to Save things.

merryoncherry: Is there a suggestion you want implemented for the PR?  Or a suggestion you'd like to see in a follow-up, broader PR?

FWIW, on my setup, the thing that is the slowest is building a couple of trees... I am sure I can fix those without a huge overhaul of the XML stuff.

AzGilrock: No that conversation was for the future.  This PR is good to go I was just at work yesterday and didn't look at it till now.  Thanks for reminding me.

Issue #3902, opened 2023-08-09
Add a save as option to the color palette save button


cox941:
In the current version of xlights, when you save a custom color palette it defaults to the palettes folder in the show folder. It is frustrating to have to go to that folder and rename your custom colors. 

I would like to see the addition of a save as option when you press the save icon.

With the additional colors available in AI making custom palettes is easier than ever. 

![Capture](https://github.com/smeighan/xLights/assets/111090879/b68632c2-e5ca-417f-be51-217679121cfc)

cbennett6008:
Perhaps if space is limited, an option could be added to the dropdown to save as new?

cox941:
sure  anyway to create a new named palette. thank you for looking at this.

cybercop23:
@derwin12 maybe a way to add a prompt when saving a palette file to enter/override the name it was going to use.

PR #4313, opened 2024-02-05
Added ability to save color palette with a name

derwin12:
#3902 additional box that can optionally be entered. Nothing fancy .. just a name that is stripped of any funny characters.
Thoughts on the layout?
image
image


computergeek1507:
I think it makes the GUI messy looking, I think "renaming" pallets has a minor use case and shouldn't be front and center to every user

derwin12:
> I think it makes the GUI messy looking, I think "renaming" pallets has a minor use case and shouldn't be front and center to every user

What about a popup when you click the save icon? I tossed around a couple of ideas on zoom. With this implementation .. the default window has it fall off the right and you dont even know its there :-) which is a good thing for many.

derwin12:
Change up the UI a bit..
image

derwin12:
Change up the GUI just a bit..
image

cybercop23:
@computergeek1507 are you okay with this new layout? It's stacked as to not make the top section wider and seems to "flow" more naturaly.

dkulp:
I'm not.  I don't see the point in having the palette name field there at all.

computergeek1507:
I think it has a limited use case and is not needed. Not all feature requests have to be fulfilled.

Issue #4316, opened 2024-02-07
Check Sequence Duplicate nodes are flagged when on different lines of the submodel definition


derwin12:
This comes from the prior discussions but I wanted to create a new ticket to code against.
Showstopper spinner is one example but any type of repeatative submodel that reuse pixels will get flagged.
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/47e1297c-3cef-48e1-be98-a2c3167bd4dc)

derwin12:
What are the thoughts on this "sensitive" topic .. do we remove this check? do we hide this check? do we close the ticket and ignore? :-)

cybercop23:
We should be able to hide it, just like duplicate in groups is hideable. Although with the new changes, I wonder if the previously strong objections are less of an issue now.

Issue #4292, opened 2024-01-23
Add option to Prefs/Check Sequnce to Disable Submodel duplicate node check


cybercop23:
Currently, there's an option to Disable checks for duplicate nodes in model groups. Add a similar one, or change, to also include submodels.

![image](https://github.com/xLightsSequencer/xLights/assets/23619433/70b9e0a0-ef08-46dc-ac02-fbc6cf32883c)

Toooooo muchhhhhh (and this is just the bae folder - not all props in main shows)
![image](https://github.com/xLightsSequencer/xLights/assets/23619433/78e216cf-cb57-40ae-8e86-d1654957b589)

cybercop23:
This is done with the commit for #4316 . That makes it even better than the original request.

PR #4317, opened 2024-02-07
#4316 Check duplicates at the submodel layer level

derwin12:
#4316 and #4292 - this no longer flags the intentional duplications that take place in many of the newer props.

The bottom one is not flagged, the others are.
image
Keep XY is unchanged and still works and catches the dup.

image
New results.

image


dkulp:
Needs to be re-based.

derwin12:
Going to redo this based on Dan's commit on handling the duplicates.

Issue #3876, opened 2023-07-15
3d depth issue with imported props


SniperMunyShotz:
I am noticing that when you import a prop on 3d, the prop is importing depth in the 1000's of inches like 8572, 7555 etc and that messes with alignment of props from the front etc until the depth is changed to 2 inches.  Is there a way to have xlights default the depth to 2 inches?

AzGilrock:
What prop are you importing cause I don't see that.  And does xLights say its inches?  That must be new because those values were unitless when I wrote it.

merryoncherry:
Let me try to explain it a bit better, as I've seen it also (primarily on custom models).
In a 3D layout, pick download or import.
Drag a box on the layout.
Pick the model to import.
Once complete, "Scale Z" might be set to something like 8.381; but I've also seen it set to much larger numbers.
![image](https://github.com/smeighan/xLights/assets/102841136/f7502869-203f-46e4-8665-5e9fa4fcf919)

Anyway, sure Gil, "Scale Z" is unitless, but if you have a ruler in your layout, there is a conversion done... I think that's what the reporter is referring to.
![image](https://github.com/smeighan/xLights/assets/102841136/5b3f45c3-801a-45ab-8656-fee144fcbdef)

I would also point out that it does cause inconvenience, if you don't set them consistently it messes with the depth of things and what is in front of what, so I do routinely go reset it in my layout.

derwin12:
Could it be an uninitialized value after the import?

AzGilrock:
Import positions are defined by where you draw the box for import

derwin12:
> Import positions are defined by where you draw the box for import

Even the Z?

AzGilrock:
I’ll try to double check it but yeah I thought the box you draw defines where the import lands

SniperMunyShotz:
This pretty much does it for any prop that is 2d imported into 3d with the ruler enabled.  Boscoyo cow model was the latest that imported at 7500 inches.

AzGilrock:
Yeah the problem to me is that the depth is mis-calculated for Custom models.  I can cause the same situation without importing a model because the ScaleZ parameter is what controls the depth on that model.  You can see that value change when you change the Depth number.

[...]

PR #4329, opened 2024-02-09
#3876 - set a default depth on model imports

derwin12:
#3876 - when importing or downloading xmodels the z was not set and could end up with a very large z/depth value being defined. This caused some issues with 3d. This change puts in a reasonable default.
@AzGilrock Gil -- does this seem like the right place/code to address this? Seems to work.


derwin12:
> It would probably be fine if we just changed things so that they import the same size as they export. Most displays have the same general sizes since most aren’t zooming way in or out before they start. So that code is in the layout panel I believe.

It does that if the export has the depth in it .. like the newer model exports do. I tested with some of the new xmodels. This handles the older pre 3dmodel xmodel files.

cybercop23:

Issue #4332, opened 2024-02-09
Download Models Dialog Default to Collapsed


MrPaulAR:
When opening the download models box have all the menus collapsed by default.  The end user can choose what vendor(s) they want to drill into without scrolling.

PR #4334, opened 2024-02-09
Dont initially expand the vendors on the download models screen #4332

derwin12:
Makes for cleaner and easier navigation


cybercop23:
Do I dare ask to also sort them? We do that for everything else.... ;) 😜

Issue #3125, opened 2022-02-08
Import Previews and Models - Added selection ability


derwin12:
On the layout, right click and import Previews/Models/Groups.
Can there be a couple of things added like:

Select Highlighted .. so one can hilight a bunch of props using click/shift click, to check/select them for import.
Select Siblings - not sure if its broken or what but if one selects a group or highlights a group, right click select siblings, will select all the props not just the props in that group.

Thanks 
Daryl
(from a zoomer)

bcarriveau:
Beat me to it. Was just typing one out, Yes please! Having group one click siblings would be amazing, multi select shift click would also be a plus!

cybercop23:
@derwin12 the siblings will be best.. select the model and it will then select all groups... sounds like right up your alley - toegether with select highlighted (since you have so much practice)

derwin12:
Looking at the code "siblings" is using the Tree terminology .. ie The top of the tree (Default) points to pretty much all the models,
 
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/9540fe50-658a-486f-851c-db87847b664e)

It doesnt consider groups->models as being siblings. Going to see if that is feasible.

cybercop23:
so those are acquaintances and can add true siblings

computergeek1507:
Why not have 2 opinions, all groups and all models under preview

PR #4336, opened 2024-02-10
Add select highlighted to import previews right click menu #3125

derwin12:
Added select highlighted and deselect to the right click menu. Also changed the select siblings menu choice to select the item along with recursively select all the models that the group contained, if a group was selected.  #3125


computergeek1507:
You love "select highlighted" I don't think I have ever used it

derwin12:
> You love "select highlighted" I don't think I have ever used it

Have you seen my layout :-)

Issue #4248, opened 2024-01-05
Increase Twinkle steps (or other method to slow down the effect)


schreyack:
**Is your feature request related to a problem? Please describe.**
I would like to have the twinkle effect transition more slowly.  Currently you can only go up to 200 steps which results in a fairly quick twinkle

**Describe the solution you'd like**
Increase the number of steps available from 200 to 400.

**Describe alternatives you've considered**
I don't think there are any alternatives available.

**Additional context**
N/A

Issue #4099, opened 2023-11-05
Increase maximum Movement speed for Text Effect


KinzuaKid:
**Is your feature request related to a problem? Please describe.**
When scrolling (moving) text across a large high density model, such as a panel matrix, the maximum movement speed is often too slow to achieve the desired effect. It appears the movement speed is an absolute value of pixels per second, adjusted by the speed setting, which means the larger the area in pixels, the slower the effect moves in terms of inches or cm/second. I only say "appears" to be an absolute value of PPS because that's what the practical output looks like. I actually have no idea how you're doing it. Magical incantations or genii wishes, probably.

**Describe the solution you'd like**
I'd like the "Spinal Tap" treatment for this effect. Currently, the top end value for "Speed" is 50. I'd like to see that increased to at least 51. Who am I kidding? I want that bad boy to go all the way to 255. Sometimes, I just want a word or phrase to fly past the viewer. Heck, I'd even be happy with a more sedate value like twice as fast as the current top end movement speed.

**Describe alternatives you've considered**
I've attempted to shorten the effect duration, which does precisely nothing because that's not how the speed variable works for the text effect. I've also tried entering values like 100, the imaginary number "i", or any of a dozen Mersenne primes in the speed field. Sadly, xLights is not familiar with these values and assumes I mean "50." Mashing the keyboard like homo neanderthalensis after a successful hunt doesn't appear to have any meaningful effect, either. For unrelated reasons I had to purchase a new keyboard to finish this feature request.

**Additional context**
I happen to have a rather wide matrix approaching 1000px across for which this is a particular problem. While I don't expect this to be a common issue, I reason that the proliferation of panel and virtual LCD screen matrices will make this a more significant limitation in the future.

cybercop23:
@dartunder maybe.... ?

PR #4337, opened 2024-02-10
Faster text, slower twinkle #4248 #4099

derwin12:
Increase the slider limits for text and twinkle. No change to existing sequences but now allows one to make text go even faster for larger props and for the twinkle will take longer to fade away. #4248 and #4099


derwin12:
Oh yes .. thanks .. I will get that.

derwin12:
I notice my code blocks throws lines like this in. 
Is that just some new template code that came in a newer version of code::blocks?

FlexGridSizer142->Fit(Panel16);
FlexGridSizer142->SetSizeHints(Panel16);

Issue #2444, opened 2021-01-04
Tool Icon for Select Effects


Clyde-L:
I keep forgetting there is a  Select Effects box.  Can an Icon be added to the tool bar to turn on and off the Select Effects panel?  

thank you Scott!

Clyde L

AzGilrock:
I've always wondered why we needed that panel at all.  The toolbar has all the effects plus you can drop any effect and change it with the dropdown in the Effect Settings panel.

Clyde-L:
> 
> 
> I've always wondered why we needed that panel at all. The toolbar has all the effects plus you can drop any effect and change it with the dropdown in the Effect Settings panel.

Actually it's quite helpful when helping out other people.  Especially since people don't open/render/save downloaded sequences in the originating layout.  Find Effects is utilized for this and makes quick work of bulk editing of file location for pictures, shaders, and videos.  Again, I keep forgetting about it because there is nothing to remind me it's there.

AzGilrock:
Sounds like I was confused.  I thought you meant this panel.
![image](https://user-images.githubusercontent.com/2024669/103592128-6c690f80-4eaf-11eb-8d1f-24cf62e11fd7.png)

Clyde-L:
Yeah, sorry I could have posted this originally 
![image](https://user-images.githubusercontent.com/12719816/103592212-6a5b7c80-4ec0-11eb-8460-948c1e79fe49.png)

AzGilrock:
Yeah that one is newer.  I was imagining using the other panel to "select" effects.  I always thought it was unnecessary considering its almost a duplicate of the effect toolbar.

stevegiron:
I agree it's a dupe.

What would be nice, but I'm not creating a report for is the ability to
disable effects on the toolbar. (Without doing a custom build)
I could do without Life/Tree/Piano/etc. I"m sure others would like this
too. We'd be able to make the icons bigger and have them all fit across the
screen if a few were gone.

priemd:
There is an option in preferences under the View tab to change the size....... I can go to Extra Large on a 2560x1440 display and all the effects fit fine, choosing Medium they also all fit on 1600x900 resolution

![image](https://user-images.githubusercontent.com/2279532/104080679-74021e80-51f7-11eb-8727-c2bf6df9a111.png)

cybercop23:
Is this still a thing?

cybercop23:
Implemented. Will be available in the next release or tonight's nightly build.

[...]

PR #4351, opened 2024-02-14
Add Select Effects to the UI Toolbar #2444

derwin12:
I created some icons as well .. if anyone has thoughts on if it is ok or other ideas .. feel free to comment.

My thinking was to use the effects icon and the finger since that seems to be the usual icon for select.

image

I am not sure if there are any hidden things that need to be done/included for production rollout of a change like this.


computergeek1507:
I'm Surprised that anymore uses this dialog I added

dkulp:
Kind of wish we had svg's for all the icons... :(

derwin12:
It is actually very handy for changing up stuff .. i can see power users using it quite a bit..

dkulp:
> > Kind of wish we had svg's for all the icons... :(
> 
> Would they replace the icon files or in addition to them? I've only known one way to create the icons but I'm fairly good at creating vector art.

Eventually replace.   The BitmapBundle stuff that is now used in wxWidgets supports using SVG's instead of individually sized bitmaps.   That would (supposedly) allow arbitrary scaling of toolbar sizes and such while keeping the icons looking good.   Not sure it really matters though.

kppankratz:
@dkulp - I emailed you the SVG files on 12/3/22 for the icons.   Should I resend?

cybercop23:

Issue #3817, opened 2023-05-16
Dimensions no longer working for 'Image' type props, Dimensions not showing, Rulers not working


DoctorWiz:
Dimensions are no longer working for "image" type props, displaying all zeros.  Not sure when it broke but it was several releases ago.  Definitely used to work because I relied on it the last time I modified my layout which was with in a 2022 release.
Furthermore, dimensions are shown when opening an old layout, but when I tried to recreate the problem in a new blank layout, dimensions were not even listed in the prop properties.  See attached pic.
Rulers no longer show a correct length, just always "1".
Dimensions for Candy Canes show their width as "Length" but the value appears to be correct, just mislabeled.  The height though is way off, at least 3-4X what they really are.  I believe this used to work correctly in an older version.
Single Line props show their length, but if they are at an angle they do not show how much space they take up in X, Y, Z.  Likewise, PolyLines show the total length and the individual segments lengths, but not how much space they occupy if at an angle, or curved.  Don't believe that ever worked.

<b><i>Steps to duplicate:</i></b>  <s>Create a new show folder, and a new layout.</s>  Open an old show folder and old layout.
Add a grid for reference (If there isn't one already).
Add an "image" type prop and choose a jpg, png, gif, etc., or select an existing "image" type prop.
Size it approximately using the grid.
Note that the height and width in the prop properties window are zero.
Add or select a candy cane.
Size it approximately using the grid.
Note that the listed "Length" seems to match the width, from the tip over to the vertical shaft.
Note that the "Height" (which is grayed out BTW) is 3-4 times its actual height according to the grid and/or ruler.
Also note that there is a height property near the top of the properties window, and another (apparently unrelated?) height near the bottom under "Size/Location".
Add (or select) a "Single Line" prop.
Move one end so it is at an angle.
Note that under 'Dimensions' it does not say how much space it occupies in X, Y, Z.
Add a "PolyLine" prop with several segments.
Move some of the points around to various angles.
Make one or more segments curved.
Note that under 'Dimensions' it does not say how much space it occupies in X, Y, Z.
Add or select a ruler.
Set it to an approximate known size using the grid.
Note that the "Real Length" always shows "1" no matter how long the ruler actually is or where you place its end points.

[...]

PR #4357, opened 2024-02-19
Allow dimensions for the Image type #3817

derwin12:
When editable dimensions were added, images dimensions didnt work. #3817

### DIFF
--- a/xLights/models/BoxedScreenLocation.cpp
+++ b/xLights/models/BoxedScreenLocation.cpp
@@ -1286,10 +1286,18 @@ float BoxedScreenLocation::GetBack() const {
 }
 
 float BoxedScreenLocation::GetRestorableMWidth() const {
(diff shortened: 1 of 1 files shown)


derwin12:
> I'm not sure this doesn't cause a change in behavior for the SaveDisplayDimensions and RestoreDisplayDimensions functions. If the RenderWi was 1 the GetRestorableMWidth used to return a 0 and then SaveDisplayDimensions converted the 0 to a 1. With this change it will return scalex instead of 0 so anyone with a scale other than 1 would get a different result. Looks like that function is only used for Custom models.

I did test on custom models and many others .. but I will do a side by side comparison to triple check.

derwin12:
Testing seems to show it working as coded. Created custom model, resize, export, import, download, locked, unlocked.

Issue #3752, opened 2023-03-15
Collapse/Expand Models on Import Mapping


cybercop23:
With the growth of HD props and sub-models upon sub-models, the list is getting very large especialy if you import/map at the sub-model level. It will be great to have the ability to collapse (or expand) all models, just like in the sequencer.

Additionally it will be great to have an indication when a model is collapsed that there are mappings at lower level.

cybercop23:
#4016

cybercop23:
Implemented. Will be available in the next release or tonight's nightly build.

PR #4359, opened 2024-02-19
Add expand and collapse to import mapping dialog #3752

derwin12:
On the import effects dialog, add right click option to expand all (submodels and first strand) or collapse all along with a show all mapped models #3752


cybercop23:

Issue #3821, opened 2023-05-22
Update for Sequence Preferences


Clyde-L:
Would someone consider adding a "Disable Color Update Warning" to the xLights Preferences?  It's common for us to update colors in our sequences to see the quick outcome, but to then always click the "yes update color" becomes quite monotonous.  I do see a need for this function, and it should be set to "on" for the default when a new user installs it.  However many users do not need the warning and I do not see a way has been added to disable it.

Thank you for looking.

Clyde

![image](https://github.com/smeighan/xLights/assets/12719816/d61f4a16-1549-416b-aa23-ae673e1004f7)

![image](https://github.com/smeighan/xLights/assets/12719816/ad6338c7-e62d-4ffc-9a71-62e1341e15c9)

computergeek1507:
I'm not a huge fan of the new warning either.

Clyde-L:
Scott, any way to add a checkbox to disregard this update authorization?

cybercop23:
Implemented. Will be available in the next release or tonight's nightly build.

Clyde-L:
Thank you

PR #4360, opened 2024-02-19
Add preference to hide color update warning #3821

derwin12:
If checked, when you use the update button in the color palette window, it does not issue the warning.
Perhaps in the future, it could be part of a bigger preference to something like an "advanced user" type of setting
#3821 
image


cybercop23:
People with the power.. can this please be merged?

Issue #2409, opened 2020-12-12
Warning - SubModel Does Not Exist


jimswinder:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
On first opening a sequence it does not find all instances of Sub Models that were renamed. Have to save then close sequence then reopen it several times to find more Sub Models that do not exist...had to to do this 3 times to find the 16 Sub Models that I renamed.

1. Open Sequence and find correct Sub Models from Warning
2. Save Sequence when Warning goes away
3. Reopen same sequence and answer new set of Sub Model warnings and pick correct sub model
4. Save Sequence when Warning goes away 
5. Repeat as necessary till Warning no longer pops up after opening sequence

**Expected behavior**
To correct ALL Sub Model inconsistencies after first opening sequence.

**Screenshots**

![Warning 9](https://user-images.githubusercontent.com/13740582/101996191-0361f380-3c85-11eb-8b08-799ae2593f66.jpg)
![Warning 9-1](https://user-images.githubusercontent.com/13740582/101996189-02c95d00-3c85-11eb-97f3-373d04bae7d4.jpg)

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version: 2020.54

[20201124 - Baby It's Cold Outside.zip](https://github.com/smeighan/xLights/files/5683838/20201124.-.Baby.It.s.Cold.Outside.zip)
 

**Additional context**

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

cybercop23:
Implemented. Will be available in the next release or tonight's nightly build.

PR #4361, opened 2024-02-20
Remapping of missing submodels skips over some #2409

derwin12:
When opening a sequence with missing sub-models, it will show you every other one that was missing. This is because the list of sub-models changes as you map or delete them. #2409 
If one has a cleaner way to handle this, please comment.


cybercop23:
Never knew this was a problem and always have to remap as I change things... Thanks much @derwin12 

People with the power.. can this please be merged?

computergeek1507:
I don't understand how if it's shorter by 10 only removing 1 fixes it

derwin12:
Each time you map one .. it check to see how you handled it.. it the list is shorter, it removes that 1 that you just mapped.
1.2.3.4.5.6.7 becomes
2.3.4.5.6.7
So if you were pointing at "3" it was index say 3, but now it needs to be 2.
Add rinse and repeat..

Issue #4346, opened 2024-02-13
Alias window is not scaling properly


kevinsaucier:
Yes, I know I can just drag out the window, but this is a bug tracker and this is a bug.  😁

The alias window does not accommodate longer alias names very well. 

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/e57bf9cb-646f-4a50-8d74-c11e7861d70e)

On a side note, I assume they're converted to lowercase for matching purposes, but it really kills the readability on the window.  It would be ideal to leave the entries as entered and convert to lowercase when doing the match in the code.   Just my opinion.

Thanks!

cybercop23:
Implemented. Will be available in the next release or tonight's nightly build.

PR #4362, opened 2024-02-20
Size the alias dialog #4346

derwin12:


cybercop23:

Issue #3931, opened 2023-08-22
Exporting a sphere as a 3d custom does not include all pixels


atticusm:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
Exporting a sphere as a 3d model does not export all the pixels

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)
1. Create a sphere
2. Set parameters to 360, -86, 86, 1, 40
3. Right click and export as a custom 3d model
4. Import model

**Expected behavior**
Model data should have all 40 pixels

**Screenshots**

**Versions (please complete the following information):**
 - OS: [e.g. Windows 10, Mac 10.14.5, etc...]
 - xLights version [e.g. 2019.37]

**Additional context**
After importing the 3d model, only 23 out of the 40 pixels will import. Scott mentioned there was an issue when exporting non-3d where if multiple pixels were in the same area they wouldn't export. This was fixed for the 2d model but not the 3D. 

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

I created a sphere with the dimensions above and exported it as a 3d model. I looked at this model in notepad++ and it did not have all the pixels. Same as re-importing it. https://drive.google.com/file/d/1CaEoThSSTKDO15DjQihOH49ltovAuGN0/view?usp=drive_link
If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

computergeek1507:
what is you stands per string count?

atticusm:
8. The numbers are:
Degree 360
Southern Lat -86
Northern Lat 86
# Strings 1
Node/String 40
Strands/String 8

The custom model is accessible with the google drive link above

digitydogs:
This bug is also affecting exporting of mega trees as custom models. 

Create a 1 string 950 node 19 zigzag tree and export it as a custom model.  The generated model will be over 4 megs in size and will be missing the lower right half of the tree.   

[...]

PR #4363, opened 2024-02-21
Scale the sphere to get a good 3d model export #3931

derwin12:
Using the same technique done for 2d to find the scale factor to get a complete 3d export of the sphere. Also include the dimensions if available. #3931

### DIFF
--- a/xLights/models/SphereModel.cpp
+++ b/xLights/models/SphereModel.cpp
@@ -288,7 +288,6 @@ void SphereModel::ImportXlightsModel(wxXmlNode* root, xLightsFrame* xlights, flo
     }
 }
 
-#define SCALE_FACTOR_3D (2.0)
 void SphereModel::ExportAsCustomXModel3D() const
 {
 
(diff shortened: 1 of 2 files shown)


cybercop23:

derwin12:
Just want to add after seeing Alex's comments .. that merging in means we can get some nightly builds in to test them by many folks .. no rush to put out to production at all.

computergeek1507:
I did try this at one point but it makes the custom models x^3 times bigger per iteration and then the custom dialog is super slow

computergeek1507:
I think the twinkly has the same issue, I was thinking that Gils float based multipoint model or maybe a way to allow two nodes in the same location on a custom model as a better solution then just making the models huge.

cybercop23:
> Just want to add after seeing Alex's comments .. that merging in means we can get some nightly builds in to test them by many folks .. no rush to put out to production at all.

Couldn't said it better myself. Selfishly since I'm on TestFlight, I just get the latest eveynight.. so I'm good and happy to test.

derwin12:
> I think the twinkly has the same issue, I was thinking that Gils float based multipoint model or maybe a way to allow two nodes in the same location on a custom model as a better solution then just making the models huge.

My first attempt it did .. I found going from 2 (the original code) .. most of my attempts worked at 3 or 4. So not overly big < 100x100 in the custom model. It wasnt that much of an impact. Another one of those things that I doubt is used all that often.

Issue #4364, opened 2024-02-23
Allow for changing of the View when importing a sequence


kevinsaucier:
I think this is more of a bug than a feature request but it's technically working as designed (I think) so figured I'd make it a request.

When opening an existing sequence, if you're missing a bunch of models, you're prompted to import instead of opening.  This is great.  Unfortunately, when you get to the import dialog, it appears to only contain the models/groups from the sequence and not from the current available views and you have to close the dialog, change the View, and reopen the import dialog again to finish mapping.

Prompted and click No

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/35678e4e-b64b-4267-988c-1ed561361e63)

Import dialog opens with what appears to be a random sort (ideally, this would be fixed to Alpha sort as well) so I resort it by model name into:

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/c6045171-c507-4487-a6d1-a8f33d91848a)

You can see here that Auto Map was not particularly helpful, even though I have numerous aliases set up, because the matching props/groups are not in the list.

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/b523c26d-d8fb-4ff8-b1ef-1bd56fe854c2)

Cancelled the import dialog and made my preferred view (set in Preferences) my Master, then opened the import dialog again.  Again, the list seems to be sorted in some random fashion so I re-sort it by name into:

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/fd1cf1f6-7cd8-486e-ad90-09276ea71d2c)

You can see AutoMap was far more successful this time as the matches were available.  Also, there are far more models in the 'Available' column than the first time.  I'm assuming the first time only shows 'unmatched' props and this time, of course, is showing all props.  I don't see a way to open the list of only 'unmatched' props again after changing the view.  I guess it's not a big deal, as AutoMap continues to map them again, but it's a bit confusing.

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/6ac7e145-1d1f-4e96-b260-d9e297b2e880)

What I would expect to see:  When you click No to the 'Do you want to continue opening this sequence' dialog, it should then ask which View you want to set as Master for the sequence prior to opening the Import window.  It could also just default to whatever view is set in the Preferences as the Default View.  I know that would probably result in the View of the sequence changing in the background but, since the sequence shouldn't be open yet, that shouldn't really cause any issues because no effects should exist yet.  

[...]

PR #4370, opened 2024-02-26
More tweaks to the import dialog #4364

derwin12:
Restore the sort but remove the sorting of the sub-models (the crux of the original issue). Also add indicators on the Available/Effects sort side. #4364
There is no option to pull in anything that isn't in the master view. You need to have your master view correct prior to starting import effects. All groups will continue to pull in any subgroups and/or models.


computergeek1507:
IDK if you tested this with LOR S4 and S5 stand mode to see if they are sorted correctly

cybercop23:
@derwin12 so looks like the sub-models are sorted in reverse order. SS Flake in the middle starting with **Whole Flake, but it is last on the Import map to list.

image

PR #4381, opened 2024-03-02
Suppress models with no effects and skip alias

cybercop23:
When opening a new sequnce after renaming/alias models, you get prompted with many missing models, however they will map via alias. This suppresses the message if they will later map. Also skip models missing that have no effects on them (or any of their submodels)


cybercop23:
Addresses #4382 4382

Issue #4366, opened 2024-02-23
Display Elements Windows Sizing/Maximizing


kevinsaucier:
2 part issue, mostly minor but annoying:

Part 1:  If it's done on purpose and I just don't know why, I get it.  If a size just isn't being specified, it would be great if it could be.  Every time I open the Display Elements window, it's so small as to be mostly unusable.  Not sure if this because it's expected to be docked or what.  I know I can resize it (and I do, every time), and it seems to hold the size for the session of xLights, but won't hold the size between xLights sessions and has to be done again the next time.  Could we set the default size of the window to be something a little more usable or at least save the last used sized across sessions?  

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/2ff3dbcc-d8bf-4b43-ab12-1b928488921e)

If it's not docked, seems like 30% of the monitor width and 80% of the height makes for a more usable window.

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/6727e7d1-2ecf-41c6-b820-7faa4f739e1c)

Part 2:  This has been an issue forever so not sure if it just can't be solved or has never been documented/requested.  When the Display Elements window is maximized (by double clicking the window title bar, the windows doesn't respect the Windows toolbar and goes behind it so you can't see all of the items in the list boxes.  Implementation of part 1 of this request might make part 2 less of an issue, but it's still kind of a weird annoyance.  😁

![image](https://github.com/xLightsSequencer/xLights/assets/25715814/e92535d7-8a7c-4817-876f-4dde1bf1f622)

Thanks!

cybercop23:
Argggg... this drove me bunkers... It gets set to 600x400 in init ON windows. On Mac is different. (thought that was the pourpose of using wxWidgets - crossplatform)
![image](https://github.com/xLightsSequencer/xLights/assets/23619433/c8731dcc-45ee-4684-8247-ae655671b051)
![image](https://github.com/xLightsSequencer/xLights/assets/23619433/e3529f28-2fd7-4051-be66-0a61a596dd5a)

So looks like we'll just need to set it on init and then it will use that.. in TabSequencer init to larger size and/or do a calc to get screen res and do a proportion

`    logger_base.debug("CreateSequencer: Hooking up the panes.");
    m_mgr->AddPane(displayElementsPanel,wxAuiPaneInfo().Name(wxT("DisplayElements")).Caption(wxT("Display Elements"))
                   .Float().MaximizeButton(true));
    // Hide the panel on start.
[...]

PR #4383, opened 2024-03-02
4366 displayelements windowsize

cybercop23:
Make the initial window larger for easier use. Even if too large, there's nothing at the bottom that will prevent the user to easily resize.


cybercop23:
#4366

derwin12:
Alex - I think you included your other commit in this commit.

cybercop23:
Yeah.., changed same file. Didn't knwo it until after I idid the other commit and the started to look at this one.

Issue #4349, opened 2024-02-14
Cant delete an 'inactive' model from layout


bloojoop:
xlights 2024.02

trying to delete a model that is set to inactive yields the following error message. 
![image](https://github.com/xLightsSequencer/xLights/assets/88390168/1a45eec2-0fdf-4a0f-afb2-a60530b9379e)
There are no open sequences and no effects. 

If I reconfigure the model to be active, then it can be deleted with no issues. 

I would expect that we should be able  to delete inactive models.

cybercop23:
FYI... The plan is to allow the deletion of an Inactive model.

cybercop23:
Implemented and merged to master. Please test the nightly (preferred) or in the next version.

PR #4387, opened 2024-03-05
Allow Inactive models to be deleted. #4349

cybercop23:
Remove the inactive indcatos so the list of the models to be deleted can match the available models.


derwin12:
Alex .. wasn't there a message popup that warned and said something about "or deleted models" .. perhaps update that message box.

cybercop23:
Yes, the message was showing up because the list of models to be deleted started and ended with the "inactive" characetrs <>. When that list was being matched against the allmodels list that doesn't include them, it couldn't find them, so they weren't deleted - that's why that message. Can't suppress the message since they would never be deleted b/c of a name mismatch.

Issue #3428, opened 2022-09-16
Lua PromptSequences - can it return the High Definition selection some how?


derwin12:
Seems it only returns the sequences selected. Is the high def prompt exposed so it can be used in  the lua script?

```
seqs = PromptSequences()
for i,seq in ipairs(seqs) do 
  Log(seq)

  properties = {}
  properties['highdef'] = 'true'   -- CAN I GET THE SELECTION FROM THE DIALOG?	
  RunCommand('renderAll', {})
-- etc etc
end
```

![image](https://user-images.githubusercontent.com/4643499/190700700-57632725-96c5-401b-b3f4-e0a33bb43c5e.png)

PR #4393, opened 2024-03-06
3428 lua script

derwin12:
A small tweak to allow Lua script to get the force high def checkbox value.
https://github.com/xLightsSequencer/xLights/issues/3428
Existing scripts should not need to be changed but the following is an example of the code.
```
seqs,highdefflag = PromptSequences()
for i,seq in ipairs(seqs) do
	Log(seq)
	properties = {}
	properties['seq'] = seq
	properties['promptIssues'] = 'false'
	result = RunCommand('openSequence', properties)
	properties = {}
	properties['highdef'] = highdefflag
	result = RunCommand('renderAll',properties)
	Log("Render Result: ".. result['msg'])
end
```

### DIFF
--- a/xLights/automation/LuaRunner.cpp
+++ b/xLights/automation/LuaRunner.cpp
@@ -88,7 +88,6 @@ std::pair<std::list<std::string>, std::string> LuaRunner::PromptSequences() cons
             forceHighDefinitionRender = "true";
         }
     }
-    
     return std::make_pair(sequenceList, forceHighDefinitionRender);
 }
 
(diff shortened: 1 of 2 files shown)


computergeek1507:
why not use a bool instead of a string

derwin12:
I died originally have that coded but found I would need the string type in the lua script. So I did the change in the xlights side. Is that not the case?

derwin12:
Scott . are you okay with leaving with a string? or should I use the boolean?
I assume there is a lua script to convert it back in my/our scripts?

derwin12:
Changed to a bool value being returned.

Issue #4318, opened 2024-02-07
Mirror of Main model element at the end of submodel list


jackyelliott:
I use states for a lot of actions. Using them in submodels is not possible without first placing a face effect down and using state as outline. This works for when you want to statically activate the state chosen.  However, the automation of the State effect is only feasible to be used on the main model element.  This is an issue due to the fact that submodels will over write any state effects that are active due to render order.  Having a mirror of the main element at the bottom of the list that has the same attributes as the main model would alleviate two issues. One is the face effect and the other is states. Face effect is easy enough but can get confusing to most users. 
Creating a submodel for the face effect and keeping it in the correct order is simple but it seems that some users like to reorder things and can't figure out why it's broke. 

What I think may work is giving the attributes of the main model element to the last submodel in the list. Essentially mirroring the main element and sticking it in the list where STRAND is. Or having a totally new position but at the end so that it has higher render order. This can fix the "Dark Faces" effect I use and have automation on the model by using states after the submodels. 

An alternative I've used is creating a group and placing the group down on the list and using a per model default. It works but is complicated and requires a lot of tweaking each time.

computergeek1507:
See if 2d90da687a813a2f4c77b970452f63da6187fcf6 helps with you issue

jackyelliott:
This didn't test properly. Any more chance to look at this

cybercop23:
Done. Available in the nightly and will be in the next version.

PR #4401, opened 2024-03-12
Allow states on sub-models #4318

cybercop23:
Allows adding states on submodels, so you can have both faces on main and states on sub-models

### DIFF
--- a/xLights/effects/StateEffect.cpp
+++ b/xLights/effects/StateEffect.cpp
@@ -234,7 +234,11 @@ void StateEffect::RenderState(RenderBuffer &buffer,
 
     Model* model_info = buffer.GetModel();
     if (model_info == nullptr) {
-        return;
+        buffer.cur_model = buffer.cur_model.substr(0, buffer.cur_model.find("/"));
+        model_info = buffer.GetModel();
+        if (model_info == nullptr) {
+            return;
+        }
     }
 
     std::string definition = faceDefinition;


cybercop23:
Correctamundo... that's why you'd use/create a full/virtual model that contains all nodes in the model and place it on that.

computergeek1507:
The states also have to be defined to match the Submodels nodes, not the model, which would show up incorrectly in the states dialog. Submodel of 10-20 who be 1-10 in the state dialog

cybercop23:
I think Scott and I are saying the same thing, right @computergeek1507? (just want to make sure)

cybercop23:
Well... happy to test other cases, but the one that we developed for and tested in zoom worked as expected. Granted only some advanced users will use this to "blend" faces and states, but it eliminates having groups for submodels and allows you to just place it right on the model.

cybercop23:
So that the States are defied based on the "full" model so you don't have the 10-20 => 1-10.

cybercop23:
Exactly, so that you can easily sequence on the model and only one model. Seems a heck of a lot easier than shadow model (or groups) to place face and states effects.

cybercop23:
b/c of the way states black out certain nodes on diff timings. just tried it and with layers and faces at the top then states, or vice a versa doesn't give the effect we want. only once you move the state to the bottom on the shadow/virtual sub-model

cybercop23:
Good point. Just verified... no crash.. static (as expected) since the states nodes are out of bounds.

cybercop23:
OHHH MAN... that was my initial dev path. Add one to the bottom of the list for each model, that's in essence the first one [0] that's the model itself. I can work on that ... so long it gets pre-approved since I spent quite some time and wasn't getting it... :)

cybercop23:
> I can see that being a pain though cause then every time InitModel runs its going to have to access whether the submodel has been created already.

dedup list?

xLights/effects/StateEffect.cpp
```diff
@@ -234,7 +234,11 @@ void StateEffect::RenderState(RenderBuffer &buffer,
 
     Model* model_info = buffer.GetModel();
     if (model_info == nullptr) {
-        return;
+        buffer.cur_model = buffer.cur_model.substr(0, buffer.cur_model.find("/"));
```
derwin12, line 237: (I don't know) ... but do you need to make sure buffer != nullptr?

cybercop23: not in this case. you hit this when ytou place a state effect on the sub-model (only); not hitting this if you do it at the model level; by then the buffer is alreay avail.

computergeek1507 (approved):

Issue #4350, opened 2024-02-14
FPP Connect - Select Highlighted sequnces


cybercop23:
If you want to select (or deselect) sequences in the FPP Connect you must click on one of them and shift click on the last one you want selected. However when you right click and choose Select Highlighted, the checked sequence gets unchecked but it is still highlighted.

![image](https://github.com/xLightsSequencer/xLights/assets/23619433/73dfdd70-8cba-44db-af04-14b6413ae554)
![image](https://github.com/xLightsSequencer/xLights/assets/23619433/3df3fc60-8461-4dfa-88a9-19d924cd0805)

PR #4402, opened 2024-03-12
Always sel/desel highlighted item even if previously was already selected #4350

cybercop23:
Especially on Macs, when y ou click an item (sequence) it gets selected. Shift + Click will highlight the range, but if you right click and choose Select (Deselect) Hightlighted, the first one would "flip" and not have the desired effect.


cybercop23:
@dkulp I noticed this on my Mac, seems okay on Windows, but since I still can't compile on Mac, I tested it thoroughly on Windows and did what I expected. I'll test it in Testflight once available.

cybercop23:
#4350

Issue #2138, opened 2020-07-22
Save 3D selection in layout tab.


MrPierreB:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
Currently when switching between different show folders (like google shared ones). Some are in 3D and some in 2D. It would be nice if xLights could automatically check/uncheck the 3d checkbox.

In the layout tab, when 3D checkbox is selected, to save this selection when user saves on the layout tab.

**Describe alternatives you've considered**
None

**Additional context**
None

derwin12:
Can we check to see if say if some/all/any of these objects exist in the rgbeffects file and if so .. flip to 3d?
```
  <view_objects>
    <view_object name="Ruler" DisplayAs="Ruler" LayoutGroup="Default" Active="0" Units="5" Length="390.000000" WorldPosX="-68.7355" WorldPosY="195.1492" WorldPosZ="27.7839" X2="421.553284" Y2="0.000000" Z2="0.000000" Locked="1" versionNumber="5"/>
    <view_object name="Mesh" DisplayAs="Mesh" LayoutGroup="Default" Active="1" ObjFile="E:\2023\ShowFolder3D\House3D\0e426ef3-aa21-4db1-aab2-cc9d9b46c67f.obj" Brightness="30" WorldPosX="0.0000" WorldPosY="14.0000" WorldPosZ="0.0000" ScaleX="1.0000" ScaleY="1.0000" ScaleZ="1.0000" RotateX="0.00000000" RotateY="6.93695021" RotateZ="0.00000000" Locked="1" versionNumber="5"/>
    <view_object DisplayAs="Gridlines" LayoutGroup="Default" name="Gridlines" GridLineSpacing="50" GridWidth="2000.0" GridHeight="1000.0" Active="0" WorldPosX="0.0000" WorldPosY="0.0000" WorldPosZ="0.0000" ScaleX="1.0000" ScaleY="1.0000" ScaleZ="1.0000" RotateX="-90.00000000" RotateY="-0.00000000" RotateZ="0.00000000" Locked="1" versionNumber="5"/>
[...]

PR #4442, opened 2024-03-28
Save and use 3d mode stored with layout #2138

derwin12:
This will put the 3d flag in the rgb_effects file so as you switch folders it will recognize and flip to the 2d or 3d as needed.
Note that it will assume 2d until it sees this setting in the rgb_effects file so you will likely need to click it once as you first open that 3d show folder. #2138 
image

### DIFF
--- a/xLights/TabSetup.cpp
+++ b/xLights/TabSetup.cpp
@@ -1575,6 +1575,8 @@ void xLightsFrame::InitialiseControllersTab(bool rebuildPropGrid) {
         List_Controllers->EnsureVisible(itemSelected);
     }
 
+    Panel2->SetMinSize(wxSize(400, -1));
+    Panel5->SetMinSize(wxSize(600, -1));
     List_Controllers->Thaw();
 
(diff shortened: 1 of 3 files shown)


xLights/TabSetup.cpp
```diff
@@ -1575,6 +1575,8 @@ void xLightsFrame::InitialiseControllersTab(bool rebuildPropGrid) {
         List_Controllers->EnsureVisible(itemSelected);
     }
 
+    Panel2->SetMinSize(wxSize(400, -1));
```
computergeek1507, line 1578: this is your secret Controller Panel change to prevent squished buttons....

derwin12: Yes ... not so secret. How is it working for you? Isnt it always collapsed into nothing?
image

derwin12: Enhancement request :-) https://github.com/xLightsSequencer/xLights/issues/4443

Issue #2403, opened 2020-12-07
Catch or warn users about model issues before faces/sub-models are created and time wasted


madsci1016:
**Is your feature request related to a problem? Please describe.**
A user makes a custom model that has a issue (like repeated node or missing node) and then goes and creates sub-models or faces on the defective model. As they create the sub models or faces, they don't notice the node numbers are wrong. It looks visually correct. Then, they find the defect in the model data and correct it. Now all the sub-models and faces are wrong and have to be re-created almost from scratch. Scott explained this is because node number in sub-models/faces are NOT global node numbers, but an index in an array holding the node numbers. So a duplicate or missing nodes shifts the array up or down after it's corrected. 

**Describe the solution you'd like**
A) Can sub-models / faces be made to show absolute node numbers opposed to array position? 
B) A pop up or warning that the model may be defective. (IE, the lines from the check model in check sequence that catch missing and duplicate node numbers).

Defective model fresh from the official download repository.
![oopsmouth2](https://user-images.githubusercontent.com/1082294/101381260-26069d80-387c-11eb-93c2-ef6148b5f265.PNG)
Run check sequence. 
![stillbad](https://user-images.githubusercontent.com/1082294/101381082-ed66c400-387b-11eb-9c57-79081800a28a.PNG)
When I fixed that (just renumbered the duplicate node to the missing number, my faces turned into
![oopsmouth](https://user-images.githubusercontent.com/1082294/101381116-f9eb1c80-387b-11eb-84f0-7a29105d3aaa.PNG)

PR #4445, opened 2024-03-29
Show any model warnings prior to editing Face Dialog #2403

derwin12:
Warn users prior to making changes in the face submodels of potential problems with the model itself. #2403
image

### DIFF
--- a/xLights/ModelFaceDialog.cpp
+++ b/xLights/ModelFaceDialog.cpp
@@ -198,6 +198,8 @@ ModelFaceDialog::ModelFaceDialog(wxWindow* parent, OutputManager* outputManager,
 	FlexGridSizer8->Add(SingleNodeGrid, 1, wxALL|wxEXPAND, 0);
 	FlexGridSizer2->Add(FlexGridSizer8, 1, wxALL|wxEXPAND, 0);
 	CoroPanel->SetSizer(FlexGridSizer2);
+	FlexGridSizer2->Fit(CoroPanel);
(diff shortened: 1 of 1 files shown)


xLights/ModelFaceDialog.cpp
```diff
@@ -463,6 +474,15 @@ void ModelFaceDialog::SetFaceInfo(Model *cls, std::map< std::string, std::map<st
     }
 
     UpdatePreview("", *wxWHITE);
+    std::list<std::string> warnings = cls->CheckModelSettings();
+    if (!warnings.empty()) {
+        std::string warningsStr;
```
computergeek1507, line 479: join?

derwin12: Thanks .. let me readup and fix this up..

derwin12: There is no join, there is a wxJoin but these are strings. Is there something I am missing?

computergeek1507: In the utils class Dan wrote a std::strings join method. You could also do it with std::transform and a lamda

derwin12: okay .. I am a rookie here .. it is a std::list but Joins is looking for a std::vector. I would have to scan thru it all again to switch them right? Seems like a waste of cycles. Or is it simpler?
`std::list<std::string> warnings = cls->CheckModelSettings();`

`inline std::string Join(std::vector<std::string> const &strings, std::string delim)`

Issue #4448, opened 2024-04-01
XLights Crashes when Moving Heads undergo Channel Mapping


DBASQLGuru:
Sequence had just started being built - No Sequence import had occurred. Only Moving Heads were imported, then mapped. Group select all the MH lines in the sequencer for a bulk edit of the channels - The in the next step of assigning the correct channels to the correct position/device channel - Crash occurred.

**Versions (please complete the following information):**
 - OS: [e.g. Windows 10, Mac 10.14.5, etc...]  Windows 10
 - xLights version [e.g. 2019.37]  2024.06

[xLights_dbgrpt-21364-20240331T193104.zip](https://github.com/xLightsSequencer/xLights/files/14818162/xLights_dbgrpt-21364-20240331T193104.zip)

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

DBASQLGuru:
Bulk edit  8 MH by selecting them in the sequence window, then MANUALLY updating the channels without using the import tool, no mapping file used, manual update still caused a crash..
[xLights_dbgrpt-26372-20240331T194329.zip](https://github.com/xLightsSequencer/xLights/files/14818202/xLights_dbgrpt-26372-20240331T194329.zip)

DBASQLGuru:
My experience with this failure was under both conditions.  Hitting the Remap button resulted in a crash. Editing them manually after saving also resulted in a crash ( I didn’t send the report). It was on a different sequence that the crash happened on both sequences. Then I reverted to xLights 2024.5 and the same behavior Got frustrated and just called it a night and closed the app Sent from my iPhoneOn Apr 1, 2024, at 9:55 AM, Gil Jones ***@***.***> wrote:﻿
Can you clarify are you hitting the "Remap Channels" button or are you just trying to change a channel position?  And if its the Remap Channels button I assume it works if you do it one head at a time?

cybercop23:
Available in tonight's nightly and next version.

PR #4451, opened 2024-04-02
Moving Head Crash on DMX Remap Channel Dialog #4448

derwin12:
Missed expanding this dialog to 48 channels when the dmx panel was updated #4448


computergeek1507:
Wonder who made the DMX Effect 48 channels😂

derwin12:
Actually I had it correct but had to redo it to rebase it .. and missed this file :-(

Issue #4315, opened 2024-02-06
Calculation of model amp usage in visualizer incorrect either for kulp or falcon either at model or port depending on controller


sidelights:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
A Falcon does not create virtual strings for a model that does not have a model brightness set (assumed to take port brightness) in all other cases.    A kulp however does create virtual strings for each model. 

Placing a model at 100% brightness on a kulp or a falcon with a (default port brightness of 30%) in the middle of two props at default brightness results in a total amp estimate on the port that is either wrong for a kulp (because 30% is used for both the first and last model), or Incorrect at the prop level for 3 model in the chain in the popup. 

10 pixels (no model brightness set)
10 pixels at 100% at model level
10 pixels (no model brightness set)

For a falcon this would result in a total estimated amp on the port being 1.27 amps 
.17  (30% ) + .55 (100%) + .55 (100% because chained, no virtual strings created).  3rd model in the chain however this does not show up as .55 in the popup for the 3rd model in the chain, it shows up as .17 (what it would be if virtual strings at created). 

For a kulp based the model level calculation would be correct however at the port level it would using .89 instead of 1.27 amps because of virtual strings thereby being overestimated. 

**To Reproduce**
create 3 models   2 with no model brightness set, 1 with 100% brightness, set controller default port brightness to 30%.  Place models on the same port, with the 100% brightness model in the middle.  

Compare the totals at the model and port in the popups. 

**Expected behavior**
Calculate port and model brightness based on how the controllers actually function.   Adjust estimated amp draw for a model chained for a falcon, adjust the port amp draw correctly for how a kulp operates. 
![amp_usageIncorrect - Copy](https://github.com/xLightsSequencer/xLights/assets/101691917/795fc0d1-9b35-453b-8bef-5d18363beeaf)

**Screenshots**

**Versions (please complete the following information):**
 Windows 11
Xlights 2024.2

**Additional context**
null sequence created to get an example show folder with 2 controllers and models setup for this example. 

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
[...]

Issue #3577, opened 2022-11-19
Visualizer doesnt catch max pixels when a group count is left from the prior model


derwin12:
Scenario from zoom room.
One prop (RBL) has a group count of 9 followed by a spiral tree of 100 pixels.
Uploading to Falcon F48V4 just dumps with bad string upload. (attached logs).
The fix was to set the group count back to 1 on the spiral tree.

DisplayError: Falcon Outputs Upload: Failed to set board mode.

Visualizer enhancement/bug to catch this scenario and flag it as an error in the msg box below the visualizer.
[xLightsProblem.zip](https://github.com/smeighan/xLights/files/10047789/xLightsProblem.zip)

cybercop23:
@derwin12 is this stil an issue? I don't have a Falcon to test.

derwin12:
Yes ..

cybercop23:
Will be available in the nighlty and next version.

PR #4453, opened 2024-04-03
Flag in Visualizer when Group Count puts you over the Pixel limit #3577

derwin12:
Specifically for Falcon (and coded as such) if a group count is set on the port it is used for all the remaining models on the port. This has the potential to put you over the pixel limit for the port. An upload would fail, this PR finds that issue and flags it in the visualizer. Also the right click to set the group count was altered to let you reset it back to 1 as needed. Previously it would ignore the attempt to set it to 1 - since that is normally just the default setting.
Q: Is Falcon the only one (that has virtual strings and group count) that does it this way? I didnt see anything in xcontroller that I could use either. #3577 
Note I didnt see an easy way to code #4315 (for group count) without getting it real messy.

### DIFF
--- a/xLights/controllers/ControllerUploadData.cpp
+++ b/xLights/controllers/ControllerUploadData.cpp
@@ -851,6 +851,25 @@ int32_t UDControllerPort::Channels() const {
     }
 }
 
+int32_t UDControllerPort::EffectiveChannels(std::string vendor) const {
+    int c = 0;
+    int groupCount = 1;
+    for (const auto& it : _virtualStrings) {
+        if (it->_groupCountSet) {
+            c += it->Channels() * it->_groupCount;
+            if (vendor == "Falcon")
+                groupCount = it->_groupCount; // all subsequent models unless reset
+        } else {
+            if (groupCount > 1) {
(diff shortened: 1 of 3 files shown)


dkulp:
For FPP controllers, the group count is per virtual string and setting it on one virtual string would have no impact on subsequent virtual strings on the same port.

deplanche:
Would a similar fix also be needed for null pixels on the Falcons?

derwin12:
> Would a similar fix also be needed for null pixels on the Falcons?

Does start and end nulls put a falcon over the limit as well? I hadn't seen that one.. 
The original one was folks were not aware that a group count on the first prop, would go down all the props on that port.

deplanche:
> > Would a similar fix also be needed for null pixels on the Falcons?
> 
> Does start and end nulls put a falcon over the limit as well? I hadn't seen that one.. The original one was folks were not aware that a group count on the first prop, would go down all the props on that port.

I think it could if many nulls were used.  They impact the port count just like groups, in that they count as pixels.  But not sure exactly how this is being checked here.

Issue #3231, opened 2022-05-07
Model Rotation value causes unexpected submodel definition change


cfudala:
This does not need a sequence to be seen, however it will manifest itself there.

**Describe the bug**
Short version: When adding a "90" to the rotation value on the layout tab for a 360 degree tree, the depiction of the submodel on the layout tab looks fine, however the actual submodel definitions change.  This does affect certain effects when the effects are dropped on the submodel (pictures included)
I did not notice this in 2021.38 or 2021.40

Longer version:
I have a 360 degree 32/50 strand tree with a starting pixel at bottom left.
I created 2 submodels "front" pixel 1-800 and "back" 801-1600"
When looking at the submodel definitions and the draw model view, all looks fine.
At 0 rotation on the layout tab the front submodel is depicted on the left of the screen with the starting pixel to the back.  the back submodel is on the right.
When you change the rotation value to "90" on the tree model, the submodels on the layout tab move as expected.  90 degrees ccw from the viewers perspective.  The front submodel is now in the front with the starting pixel on the bottom left and the back submodel is in the back.
HOWEVER ... the submodel definition page shows a different picture.  It appears to this untrained eye, that the submodel has rotated 90 degrees over the top and not side to side.  This of course will affect certain effects during a sequence.  For me it was noticed on large gif files. And it is affecting all submodels for the tree.

I found a work around:  If I create a group containing the submodel in question, and drop the effect on the group, it renders fine in the sequence regardless of tree model rotation value.

PS. access to the forum does not work from the xLights help pulldown or from then download page.

**To Reproduce**
Create a 360 degree tree model
create at least 2 submodels 
view the submodels and note the depiction in the submodel definition window
note the depiction of the submodels in the layout window
Change the rotation value on the tree model to other than 0 (suggested 45 or greater)
Compare the new views

**Expected behavior**
Submodels look the same between layout tab and submodel definitions

**Versions (please complete the following information):**
 - OS: Windows 10
 - xLights version 2022.09

**Additional context**

**Attachments**

[problems.zip](https://github.com/smeighan/xLights/files/8644274/problems.zip)

keithsw1111:
Without a copy of your show folder it is hard to tell but i doubt there is a problem here. The submodel dialog should not show the tree rotated. It is just like the model view in the sequencer ... models are always presented in that preview unrotated. This should not impact sequencing.

cfudala:
[...]

PR #4456, opened 2024-04-04
Use the rotation if set in Tree to show model preview correctly #3231

derwin12:
This is an old bug .. @dkulp @AzGilrock  I simply removed the comments and re-enabled the code from late 2018 - when the bug was introduced. I went back and found 2019.01 was the first time it broke.
Either of you know the history here? It seems to work just fine with this re-enabled code.
#3231 
image
@keithsw1111 adding keith here as he commented on the bug ticket.
PR that did the change.
https://github.com/xLightsSequencer/xLights/commit/3f63e274d780dffef99b55a16631d7306354be07


dkulp:
I believe this now fixed.   Was definitely a display issue, not sure how we haven't run into it before.

derwin12:
Confirmed.

Issue #4441, opened 2024-03-28
Getting a warning/error on creating a showfolder from a base (not a valid model type)


derwin12:
I have two custom models in my show folder that throw this error.

![image](https://github.com/xLightsSequencer/xLights/assets/4643499/6d21e84c-10b7-41f3-a8a2-24fd6d9bfbdf)

Looking at the code it is assuming a "DisplayAs" attribute. Not sure why these are created in this format, but they dont have that attribute.

I tried recreating these custom models but they still created it using this <custommodel tag.

```
<custommodel name="tree trunk" parm1="5" parm2="50" StringType="RGB Nodes" Transparency="0" PixelSize="2" ModelBrightness="" Antialias="1" StrandNames="" NodeNames="" CustomModel="50,,51,150,;49,,149,52,;,48,148,,53;147,,47,54,;,146,55,46,;,56,145,,45;57,,144,44,;,58,43,,143;,42,59,142,;41,,141,60,;,40,140,,61;139,,39,62,;,138,63,38,;,64,137,,37;65,,136,36,;,66,35,,135;,34,67,134,;33,,133,68,;,32,132,,69;131,,31,70,;,130,71,30,;,72,129,,29;73,,128,28,;,74,27,,127;,26,75,126,;25,,125,76,;,24,124,,77;123,,23,78,;,122,79,22,;,80,121,,21;81,,120,20,;,82,19,,119;,18,83,118,;17,,117,84,;,16,116,,85;115,,15,86,;,114,87,14,;,88,113,,13;89,,112,12,;,90,11,,111;,10,91,110,;9,,109,92,;,8,108,,93;107,,7,94,;,106,95,6,;,96,105,,5;97,,104,4,;,98,3,,103;,2,99,102,;1,,101,100," SourceVersion="2018.23"/>
[...]

PR #4460, opened 2024-04-05
Handle missing model type on Base Folder creation #4441

derwin12:
When creating a base show folder it throws a warning that it cant determine the type of the model. It appears to be a slightly different format of xml and hence needs this little change to handle that situation #4441

### DIFF
--- a/xLights/models/ModelManager.cpp
+++ b/xLights/models/ModelManager.cpp
@@ -1363,6 +1363,9 @@ Model* ModelManager::CreateModel(wxXmlNode* node, int previewW, int previewH, bo
         return grp;
     }
     std::string type = node->GetAttribute("DisplayAs").ToStdString();
+    if (type.empty()) {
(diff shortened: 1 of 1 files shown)


cybercop23:
Hallelujah... I was looking at the export for the copy/paste & automatically add to groups and noticed that. Will be great to make it just one format.

cybercop23:
Man.... controller tag for submodels was on my list too to get removed. I'm sure this will cover it. Looked a bit at the project you're using. Considering ZIP too, for some of the xtra-large shows/submodels?

cybercop23:
https://youtu.be/fTabZu3BQ0s?t=196

dkulp:
The one thing I've wanted to do for a couple years is to combine the three xlights_*.xml files into a single "xlights.xsc" (xLights Show Config) file that can be registered as a type so clicking on it would automatically open xLights to that show folder.   Also wanted to move from xml to json, but that's a separate thing.

Issue #3446, opened 2022-09-26
Color highlighting in Import Effects window.


neowolfwitch:
In the "Map Channels" dialog when importing effects, it can be hard to see what effects have been mapped and which ones haven't in the "Available" window. 

Right now-  mapped effects show up in a gray color, while un-mapped ones show up in black. Unfortunately in some cases this change is subtle and hard to see. It is especially bad in Dark Mode. I'm sure this varies by monitor, and it may just be a Mac issue.

It would be great if actual colors were used here, or if we can customize the colors used in Preferences -> Colors for Import Effects.

PR #4463, opened 2024-04-07
Add preference to override color highlighting #3446

derwin12:
The preference can be used to override the cyan/blue color in lists and another to override the grey/med grey. #3446

### DIFF
--- a/xLights/ColorManager.cpp
+++ b/xLights/ColorManager.cpp
@@ -208,7 +208,7 @@ void ColorManager::Load(wxXmlNode* colors_node)
 {
 	if (colors_node != nullptr)
 	{
-        colors.clear();
+        ResetDefaults();
         for (wxXmlNode* c = colors_node->GetChildren(); c != nullptr; c = c->GetNext())
         {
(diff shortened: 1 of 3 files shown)


derwin12:
Any thoughts on why the ::instance fails on Ubuntu?

computergeek1507:
Utils fiction needs an include to wxcolor, your auto is really a wxcolor which isn't defined in the utils class

derwin12:
I am on windows .. compiles fine -- so yea -- some lib missing on the unix side?
utils did get the colormanager.h included.

derwin12:
I committed the change from auto. 
I suppose xSchedule never used color preferences? I need to test this change in xSchedule. It might not use the color preferences at all which this would sort of mean it has to or atleast it has to access them.

derwin12:
xSchedule compile did indeed fail on windows.

derwin12:
Okay .. the cleanest answer here is to remove my wedge in the utils function and keep that clean for xschedule and such.
I will close the PR and rething/recode that.

PR #4465, opened 2024-04-07
Add first controller from ILightThat

PeteMatthews:
New controller venture in the UK.

Getting scaffolding in for new controller, running a custom ESP32 firmware. Will most likely need revisiting as we expand and add features, but there's enough here to be useful. 

*Very* open to comments/suggestions/rework as required. Keen to make this as easy as possible for code maintenance, etc, so please do let me know if you'd prefer anything revised.

Thanks,

Pete

### DIFF
--- a/controllers/ilightthat.xcontroller
+++ b/controllers/ilightthat.xcontroller
@@ -1,0 +1,26 @@
+<Vendor Name="ILightThat">
+    <AbstractVariant Name="BaseBaldrickSettings">
+		<SupportsPixelPortBrightness/>
+        <SupportsUpload/>
+        <MaxInputUniverses>9</MaxInputUniverses>
+        <MaxSerialPortChannels>0</MaxSerialPortChannels>
+        <MaxPixelPortChannels>6000</MaxPixelPortChannels>
+        <AllInputUniversesMustBeSameSize/>
+        <UniversesMustBeSequential/>
+        <SupportsAutoLayout/>
+        <PixelProtocols>
+            <Protocol>ws2811</Protocol>
+        </PixelProtocols>
+        <SerialProtocols/>
+        <InputProtocols>
+            <Protocol>ddp</Protocol>
+        </InputProtocols>
+        <ConfigDriver>ILightThat</ConfigDriver>
+    </AbstractVariant>
+    <Controller Name="Baldrick 8 Port v1">
+		<Variant Name="" Base="ILightThat:BaseBaldrickSettings">
+			<MaxPixelPort>8</MaxPixelPort>
+			<MaxSerialPort>0</MaxSerialPort>
+		</Variant>
+    </Controller>
+</Vendor>
--- a/controllers/ilightthat.xcontroller
+++ b/controllers/ilightthat.xcontroller
@@ -5,7 +5,7 @@
         <SupportsUpload/>
         <MaxInputUniverses>9</MaxInputUniverses>
         <MaxSerialPortChannels>0</MaxSerialPortChannels>
-        <MaxPixelPortChannels>6000</MaxPixelPortChannels>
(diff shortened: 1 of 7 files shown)


controllers/ilightthat.xcontroller
```diff
@@ -0,0 +1,27 @@
+<Vendor Name="ILightThat">
+    <AbstractVariant Name="BaseBaldrickSettings">
+		<SupportsPixelPortBrightness/>
+		<SupportsPixelPortColourOrder/>
+        <SupportsUpload/>
+        <MaxInputUniverses>9</MaxInputUniverses>
+        <MaxSerialPortChannels>0</MaxSerialPortChannels>
+        <MaxPixelPortChannels>6000</MaxPixelPortChannels>
```
computergeek1507, line 8: change MaxPixelPortChannels to 2250 as this would be 750 pixels per port. Then you can remove the check from the CPP code.

xLights/controllers/ILightThat.cpp
```diff
@@ -0,0 +1,130 @@
...
+                }
+                i++;
+            }
+        }
+        for (int i = cud.GetMaxPixelPort(); i < controller->GetControllerCaps()->GetMaxPixelPort() + 1; i++)
+        {
+            outputConfig["ports"][i]["num_pixels"] = 0;
+        }
+
+        std::string url = "http://" + _ip + "/settings";
```
computergeek1507, line 114: if you plan to support FPP Proxy, this URL wont work. If you use the BaseController::PutURL method it automatedly adds the proxy path for you.

xLights/controllers/ILightThat.cpp
```diff
@@ -0,0 +1,130 @@
...
+        wxJSONWriter writer;
+        wxString Data;
+        writer.Write(outputConfig, Data);
+        logger_base.debug(std::string("SetHttpConfig: Data: '") + Data + "'");
+
+        std::string contentType = "application/json";
+        int ReturnCode = -1;
+        std::vector<unsigned char> value(Data.begin(), Data.end());
+        CurlManager::INSTANCE.doPost(url, contentType, value, ReturnCode);
```
computergeek1507, line 123: PutURL(/settings, Data, "", "", "application/json");

controllers/ilightthat.xcontroller
PeteMatthews: Stupidly misinterpreted that one - Thanks :)

Revised to 2250, removed CPP check.

xLights/controllers/ILightThat.cpp
PeteMatthews: That _is_ handy, thanks :)

I'll freely admit I borrowed this code from another driver. If I can remember which one, I'll fix there and PR that too later.

xLights/controllers/ILightThat.cpp
PeteMatthews: Edited as recommended, thanks :)

xLights/controllers/ILightThat.cpp
```diff
@@ -0,0 +1,130 @@
...
+    static log4cpp::Category& logger_base = log4cpp::Category::getInstance(std::string("log_base"));
+    std::string const json = GetURL("/settings");
+    if (!json.empty()) {
+        wxJSONValue jsonVal;
+        wxJSONReader reader;
+
+        reader.Parse(json, &jsonVal);
+        _connected = true;
+        
+        if (_connected) {
```
derwin12, line 43: .. you just set it to true the line above. No need for the if {}

PeteMatthews: A fair point - Revised. 

Thanks :)

derwin12 (approved):

Issue #2412, opened 2020-12-14
Some pop up windows do not respect the Windows taskbar


kevinsaucier:
When opening the Display Elements window to manage the props in the sequencer, if you maximize the window, you can no longer see the bottom of the window as it goes below the taskbar.  Same issue with the House Preview window.  Also could have sworn I'd seen the issue on a 3rd window but I'm not finding it now....  

![image](https://user-images.githubusercontent.com/25715814/102099263-1d6f1380-3df6-11eb-9e22-835e0cd2c34a.png)

![image](https://user-images.githubusercontent.com/25715814/102099467-50190c00-3df6-11eb-9df0-e22a0c257fa6.png)

derwin12:
Pretty sure we have addressed these. If not please open a new ticket.

PR #4466, opened 2024-04-07
Set Max Size of Window Based on Display Size #2412

derwin12:
Remove the arbitrary hard coded largest window size configs with actual values for the hardware. This impacts the Display Elements, House Preview and Model Preview windows. #2412

### DIFF
--- a/xLights/sequencer/tabSequencer.cpp
+++ b/xLights/sequencer/tabSequencer.cpp
@@ -13,6 +13,7 @@
 #include <wx/utils.h>
 #include <wx/tokenzr.h>
 #include <wx/clipbrd.h>
+#include <wx/display.h>
 #include <wx/filename.h>
 #include <wx/filepicker.h>
 #include <wx/fontpicker.h>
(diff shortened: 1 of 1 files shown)


xLights/sequencer/tabSequencer.cpp
```diff
@@ -89,15 +90,21 @@ void xLightsFrame::CreateSequencer()
     logger_base.debug("                Set dock size constraints.");
     m_mgr->SetDockSizeConstraint(0.25, 0.15);
 
+    const auto maxHeight = std::max(1, wxDisplay().GetGeometry().GetHeight() - 100);
+    const auto maxWidth = std::max(1, wxDisplay().GetGeometry().GetWidth());
```
dkulp, line 94: How would this work for those of us that are using multiple displays, possibly/likely at different resolutions?    I'd honestly prefer NOT setting any sort of MaxHeight on anything so that the windows can easily be moved to different displays and still be maximized.

derwin12: If it goes off the screen you cant get to the bottom. The double click still maximizes it. The original code was setting the max off your screen. GetHeight should return the largest your monitor can handle.  Not sure how it handles resolutions and such though. This code is similar to code used in Audacity.

dkulp: But that's part of the problem....   My "main" monitor has a height of 1620 (logical size), but if the window appears on notebook screen, it has a max of 1329 and it gets cut off.

derwin12: The current code is hard coded to 1050. This is better than hard coded arbitrary value. Is there an even better way? If you dont have it set, you end up having to resize the box every time.

derwin12:
Replaced by other changes

PR #4470, opened 2024-04-08
Handle duplicate model/sub-models names in groups when replacing a model

cybercop23:
When replacing a model with a model, the original model remains in groups but the new model gets renamed and also stays in hte group, thus resulting in M/S-M being in groups twice.


cybercop23:
This was introduced back in 2022.16, revised in 2022.17 and broken since. 2022.15 was working fine.

Issue #4471, opened 2024-04-08
Remove Start/End Channels from groups and sub-models


cybercop23:
Since Start/End Channels mean nothing at the group and sub-model level, remove them from showing up. Users believe they are actually all (re-)assigned at each of these levels. Should also speed up walking the tree.

PR #4472, opened 2024-04-08
Remove Start/End channels from submodels in Groups and Models #4471

cybercop23:
Keep list simple since it channels mean nothing at group level ir sub-model level.


cybercop23:
image

cybercop23:
Cuz IDK what I'm doing when it comes to this stuff... I think it putit back and added the change. LMK if that works or I need to do something else.

PR #4474, opened 2024-04-08
Remove Start/End channels from groups and sub-models

cybercop23:
To keep lists clean.

image


cybercop23:
Awesome. Thanks and sorry for the ttroubles.

Issue #4473, opened 2024-04-08
Add active Alias list to sub-model info


cybercop23:
When expanding a Model and click on a sub-model, you see info about the sub-model. Add the newly available alias at the submodel list in the below panel

![image](https://github.com/xLightsSequencer/xLights/assets/23619433/9defe240-c6fe-48c9-a68f-8745cad80be1)

computergeek1507:
I sense another pull request about to come in

PR #4476, opened 2024-04-08
Add sub-model alias to info panel #4473

cybercop23:
Including the Aliases in the list....  for @computergeek1507 ... :)
image


computergeek1507:
You could fix the the bug with a comma at the front of the node list while you're in there

cybercop23:
> You could fix the the bug with a comma at the front of the node list while you're in there

Where's that issue? Not sure I know about it. Was looking for a until ListToCsl, but we don't have one. Maybe I'll create one for easier processing.

cybercop23:
> You could fix the the bug with a comma at the front of the node list while you're in there

looking into fixing it too

cybercop23:
@computergeek1507 all good?

cybercop23:
@computergeek1507 you were right, as usual. That wasn't needed.
image

PR #4493, opened 2024-04-11
Switch FaceInfo and StateInfo to virtual getter/setters

computergeek1507:
@cybercop23 @dkulp I think this is what Dan wants you to do to add polymorphism for SubModels. I might have broke some stuff with all the reverse inserts with std::map


computergeek1507:
GCC might be too old on ubuntu for map contains

cybercop23:
@computergeek1507  WOW!!!! This is really awesome adn way way above my abilitry. I'll take a look and start testing once I get the notification. Thank you so much.

cybercop23:
BTW.. is this check/run on all platforms new? Haven't seen this.

PR #4519, opened 2024-04-22
Bars effect - Add ability to chose the highlight color.

MrPierreB:
Before Highlight would always highlight the bars with a white pixel.  Added a colorpicker similar to how we have sparkles color.
This gives much more options to set a softer highlight.

BarsHighlightColor

Let me know if you spot anything of concern.

### DIFF
--- a/xLights/effects/BarsEffect.cpp
+++ b/xLights/effects/BarsEffect.cpp
@@ -37,6 +37,11 @@ xlEffectPanel* BarsEffect::CreatePanel(wxWindow* parent)
     return new BarsPanel(parent);
 }
 
+static inline int GetHighlightColour(const std::string& HighlightColourString)
+{
+    return 0;
+}
+
 static inline int GetDirection(const std::string& DirectionString)
 {
     if ("up" == DirectionString) {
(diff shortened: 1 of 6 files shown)


dkulp:
By adding a new control type to the effect panel logic, this opens up a ton of additional work that would need to be done.   Things like "Bulk Edit", this color should have a value curve, etc....      This also then starts making it so colors are controlled in multiple spots.   

My "gut feeling" is that this should be implemented as a simple check box of "First Color is Highlight" or similar that would make the first selected color the highlight color which would match the Morph effect's "head" behavior.

MrPierreB:
> By adding a new control type to the effect panel logic, this opens up a ton of additional work that would need to be done. Things like "Bulk Edit", this color should have a value curve, etc.... This also then starts making it so colors are controlled in multiple spots.
> 
> My "gut feeling" is that this should be implemented as a simple check box of "First Color is Highlight" or similar that would make the first selected color the highlight color which would match the Morph effect's "head" behavior.

Was thinking about the colors in multiple spots.  I went this way as I used the Sparkle color selector (which cant do value curves).   
Ok cool - I am happy to make the changes so that it is rather selected as the first color selected from colors if a checkbox is selected.  
Guess we leave sparkle alone and not do the same change there (For now :P )  
 
I will do the changes

MrPierreB:
Closing this one as I have opened a new PR for this here [https://github.com/xLightsSequencer/xLights/pull/4531](url)

Issue #4514, opened 2024-04-21
Rendering Process stops around 80 percent completed - approx


DBASQLGuru:
Hello Developers of XLights -- A very unusual behavior today from XLights. I've been using the software all day rendering many sequences without issues.  Then, this 1 sequence, with the same layout, no changes, refuses to finish the render.  It re-starts the render after clicking on a model in the Sequence tab.  Have to kill the program using Task Manager, restart the application, try again.. same behavior. No zip package created to submit to Github/bugs.  Please advise how you'd like to proceed in obtaining the logs, model and sequence info.  Thanks..  UPDATE:  After forcing the app to close again. I decided to try a different song, building a new sequence.. same behavior it looks like 80% completed and then it just dies and can spin for 30 minutes or more without any progress...

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)
1. Import a song,  save then name the sequence
2. Import Boscoyo Bulby & Tune to sign - save, render then save again
3. Import sequence package - map then save, 
4. Click on render - watch many groups pass by in the render process - all stars or snowflake rings = progress stops
5. Task Manager shows CPU ranging from 5% up to 70% / Memory usage bouncing between 5 GB and 120 MB
6. Application hangs....
7. See error

**Expected behavior**
Rendering should be completed in less than 5 minutes as experienced on previous rendering experience of this sequence.

**Versions (please complete the following information):**
 - OS: [e.g. Windows 11
 - xLights version 2024.09

[xLightsProblem.zip](https://github.com/xLightsSequencer/xLights/files/15050517/xLightsProblem.zip)

derwin12:
There is no crash or anything in this log that I can see. You might need to package sequence and send to one of us. It appears to be in the render of the video - so perhaps this has been addressed by Scott already. Can you try the nightly or try without the hardware encoding enabled in the preferences.

derwin12:
I did see this in the file..

`2024-04-20 21:50:56,122 31368 log_base [WARN] Model::InitRenderBufferNodes firstNode + Nodes.size() = 0. ModelGroup::'Windows and door'. This commonly happens on a polyline segment with zero pixels or a custom model with no nodes but with effects dropped on it.`

Can you check that group .. perhaps delete it or remove the effects from it .. or add some prop to the group?

derwin12:
[...]

PR #4528, opened 2024-04-24
Crash with empty groups with effects #4514

derwin12:
This code is quite old - so surprising that this is needed. Perhaps something was changed upstream that triggers this condition. #4514

### DIFF
--- a/xLights/RenderBuffer.cpp
+++ b/xLights/RenderBuffer.cpp
@@ -782,7 +782,7 @@ void RenderBuffer::InitBuffer(int newBufferHt, int newBufferWi, const std::strin
     // This is an absurdly high number but there are circumstances right now when creating a buffer based on a zoomed in camera when these can be hit.
     //wxASSERT(NumPixels < 500000);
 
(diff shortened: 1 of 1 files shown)


dkulp (commented):
I don't believe this is correct as it shouldn't happen.   The BufferWi and BufferHt should be at least 1 (so NumPixels is at least 1).     For me, an empty group is giving both as 1.   In any case, if there is a 0 for either, my gut feeling is that we'll have other crashes and hangs elsewhere so I'd also like to get an understanding as to how the 0 is happening.

derwin12:
Ok .. I will dig further. The showfolder is in the original ticket #4514

derwin12:
Here is the layout that is crashing .. I reduced it to one empty group and a couple of effects. Debugger shows it coming in as 0x0.
[ZeroBufferCrash.zip](https://github.com/xLightsSequencer/xLights/files/15102322/ZeroBufferCrash.zip)
@dkulp

dkulp:
Fixed in ModelGroup

PR #4529, opened 2024-04-25
Arches import/export via new XML method

cybercop23:
@AzGilrock
Added LayerSizes (and helper function) + tons of other attributes (WIP)

Top arches are the new XML ones and the bottom are the original.
image

The OG has ModelBrightness that I can't find anywhere, including the rgb_effects, so I think we can kill it.
It also has the SourceVersion vs the new & rgb_effects having model versionNumber - I think that's fine too.

image


cybercop23:
> I think we will need to add the stuff that gets done by the ImportModelChildren call. It would be nice to get that done inside the XmlSerializer but to just get something quick working I just had it make the function call.

Yup.. that's what i'm working on now once i fix the ubuntu issue i created with the helper function. Want to get the submodels and groups going.

cybercop23:
Hmmm.. I'm a bit (a  lot) stuck. How do I get & pass (or use) the actual model from the visitor to a new funtion? Added a new submodel function to add children, but i need to get and pass the model so i can get to the submodels and enumerate them.

cybercop23:
@AzGilrock I got all of the sections added to the xmldoc for submodels, groups and aliases at all 3 levels. If you can please point me in the right direction to get the actual model object passed into/by the serializer I should be able to complete this and get the actual data.

image

cybercop23:
> Guess I should have looked first. Looks like Model.h already has GetSubModels() that returns what you should need.

right, but no matter what i try, i can't get it to "find" it. may be my inexperience with cpp, but usually after some reading and  trial/error i get it. this time, not so much... keeps saying GetSubModels() is unidentified 
even trying something simple to get the size... (arch->*GetSubModels().size())

cybercop23:
> void Visit(const ArchesModel &arch) override { const std::vector<Model*>& submodels = arch.GetSubModels();
> 
> That compiles and then you have a reference to the submodel vector.

I swear I had something like that.. very close at least.. bakc on track.... thank you

cybercop23:
I can get to the Aliases for the model and submodels and can output them. To get to the groups, I'm trying to use GetGroupsContainingModel or GetGroupsContainingModelOrSubmodel but they take in a Model where as wht we have is the unique model ArchesModel, CandyCaneModel etc... can't find any helper function or anything to convert from the "native"model to generic model so I can get to the groups. I don't want to use the ModelXML to get to them to only change them later. Any ideas or hints where I should go next?

image

cybercop23:
Thanks Gil for the help. I was able to get to the Groups, and creating groups tags now. Still need to get to the actual groups themselves, needs some of the aattributes added, so we can export them. Same for submodels. I have the tags, but missing some attributes. 
Need to dig into the Circle model since the 3ptslocation crashes when added to the Circle model.

cybercop23:
I f'ing just give up. I made a mistake and excluded the model{}.h since I had to add the visitor code. Was trying to just remove them so the other net-new files could be merged. I clearly didn't know what that meant. I then (I think) reverted the .gitingore but somehow my local branch isn't updating and my .h files are missing and I can't get them back to compile. ARGGGGGGGGG

PR #4536, opened 2024-04-30
XMLSerializer with visitor WIP

cybercop23:


cybercop23:
Started a new PR since I really screwed up the last one.

xLights/models/XmlSerializer.h
```diff
@@ -280,22 +512,180 @@ struct XmlSerializingVisitor : BaseObjectVisitor
...
+        wxXmlNode* archNode = new wxXmlNode(wxXML_ELEMENT_NODE, XmlNodeKeys::ModelNodeName);
+
+        const std::vector<Model*>& submodels = arch.GetSubModels();
+        const std::list<std::string>& aliases = arch.GetAliases();
+        const std::string name = arch.GetName();
+
+        const Model* m = dynamic_cast<const Model*>(&arch);
+        const ModelManager& mgr = m->GetModelManager();
+        std::vector<std::string> groups = mgr.GetGroupsContainingModel(m);
+        std::vector<Model*> modelGroups = mgr.GetModelGroups(m);
```
cybercop23, line 589: @AzGilrock so I can get to the list of groups per your suggestion before, so I can create the group with the proper name and model (Arch in this case), but when I create the group itself, I need to pull in more info from the group itself, center, buffer, preview, tag color, etc... but I can't get it since that's a ModelGroup object vs teh ArchesModel object that I'm in. Dynamic casting doesn't work (at least couldn't figure out a way to do it).

Same for SubModel, but for that I created some more functions to return what I need, but still same issue... can't get from a type ArchesModel to SubModel or ModelGroups.

Any ideas/suggestions?

cybercop23:
> I'm sure you can create a ModelGroup object if you have the list of model groups but I'm out of town at the Minnesota mini so can't give specific help till Sunday or Monday.

No worries. I'll keep trying. Enjoy the Min.

cybercop23:
okay.. i'm back on this.. so how can i convert from an arch model to model or submodel to get the rest of the relevant  info?

cybercop23:
Okay.. I was passing the results of that into another function and that's what got me screwed up. I changed it to instead pass the model so that I can re-use the AddGroups for all models and I'm a bit further.  But I'm sort of back where I was... that I can't (don't know how) get to the [ModelGroup] that's part of the "g" variable.

I call AddGroups from the Visit/Arches and pass the casted model (m). I then use that again within the AddGroups to get the list and create the groups. You can see in the output (top right) that I get the groups, and layout_attribute (from baseobj) but how can I get to the rest of the [ModelGroup] specific items?

image

cybercop23:
Yeah.. that's what tried and the GetModelGroups returns a Model not ModelGroup... This is what I keep getting stucked on.
image

cybercop23:
Arggg... why am i always one character off.... got it...... haven't looked into those values.. but that's what the func return...

`<modelGroup name="All" models="Arche1" LayoutGroup="Groups" selected="0" layout="Groups" GridSize="400" TagColour="#000000" centreMinx="-842150451" centreMiny="-842150451" centreMaxx="-842150451" centreMaxy="-842150451" centrex="-224.015060" centrey="950.592346" centreDefined="1" DefaultCamera="2D" XCentreOffset="0" YCentreOffset="0"/>
    <modelGroup name="Arches" models="Arche1" LayoutGroup="Default" selected="0" layout="Default" GridSize="400" TagColour="#FF0000" centreMinx="-842150451" centreMiny="-842150451" centreMaxx="-842150451" centreMaxy="-842150451" centrex="-469.535522" centrey="1084.479492" centreDefined="1" DefaultCamera="2D" XCentreOffset="0" YCentreOffset="0"/>`

cybercop23:
Here's the Arch model exported with model, submodels, groups and aliases at all levels.
Two things I just noticed - 1. You don't get prompted to choose and export groups - it just does it & 2. On Import although I just deleted the original model that was exported, the import picked up a new name (-2) and thus groups weren't updated as they should.

Will update the xmlserializer.h in a few, but if you can take a look and make sure things look good, I'll start at the other models.
[Arch.txt](https://github.com/xLightsSequencer/xLights/files/15243535/Arch.txt)

cybercop23:
> Kinda hard to just look at an export and know if anything is missing or wrong. I would export with the latest version of code then compare against an export with your code. I don't even know about a prompt to ask about groups when exporting but less prompts the better.

Yeah.. that's what I've been doing.. exporting and importing to make sure they match. Once I  have all models done (soon) I'll export my hole show and reimport it and diff the rgbeffects file.

I noticed, I'm getting this when casting.. should I be concered or any ways to address it? I don't get it for the ModelGroup cast:
image

cybercop23:
Assuming no issues, all models with all parms changes export and match the existing rgb_effects file. (We already had submodels and aliases exported at all levels)

Added placeholder export fuunctions to all models, so I can export and check (whynot) and lots of functions to get the attributes needed.
We still need to look at 3layerscreen location since that's crapping out and disabled on most models - but still rgbeffects match.

Next are Faces and States - since they apply to all models.

cybercop23:
I've added Faces and States and a custom sort function (needs to be added to all -I will do that) that sorts the attributes, but keeps name in the first position to make it easy to see/locate.

cybercop23:
@AzGilrock Made some structural changes and tested all exports copied into one file as xlights_rgbeffects.xml and it loaded as expected... well almost....
The `AddThreePointScreenLocationAttributes` func fails on dynamic_cast for some models (CandyCane, ChannelBlock & SingleLine) and those are the ones that are diff on the new rgbeffects.xml. It is either the x2 or angle that's needed.
I tried to "fix" ChannelBlock, but my initial thought failed. You said you had some ideas.. Can you please review that part.
Once we correct that, we should have a clean way to use this new format for import/export and save rgbeffects.

cybercop23:
Okay.. chaged to TwepPointScreenLocation and I got the x2/y2/z2 and shear. Just need the Angle that comes from screenlocation.

cybercop23:
Hey... okay.. I just added the 2Point that I needed for CB (channel block) & SL (SingleLine) and also added the DimmingCurve part.
I was just going to ask... and you already answered, so I'll combine 2&3 and just pass a parameter (I wanted to do that, but didn't know how you felt about combining functions like that).

CC works fine now with 3Point.  Yes, I have an issue with Matrix, althought Tree work, so I'll dig more into it.

What you can take a look at for me is importing with submodels. If the import has submodels, althought all 10000% correct, xlights freezes. Kept trying to pinpoint but nand. Again, if I combine all xmodel -> rgbeffects, ALL loads correctly, even SM.

cybercop23:
> The x2, y2, and z2 need to be deleted from the ThreePoint one cause they come from the TwoPoint class. If I had added that it was a mistake. You only get Shear, Angle, and Height from the ThreePoint class I believe

you get 'em too from the 3, but when using 3 it was failing on the angle/shear

cybercop23:
IDK if this is tghe right medium for this.. maybe truly stick to a discussion or get a fb char or slack going...
BUT... how can I test/see the other changes? right now I export and look at them, but all those other object can't be exported....

cybercop23:
Ahhh.. yes.. I haven't tried download, only imports, since downloads are old fortmat, we're adding more parms, so I've only been doing imports/exports. LMK when you fix that and I'll add it to my code. Not too verse in Git so IDK how we can both  work on the same one, especially with mine being a draft. Hope I won't have to rebase..
 
 Okay... can we (you) :) change the save model.. they are all actually done now, to save with the visitor? We can do all models, and hold off on the others.. already started working on that. I may need some guidance on what models to get, but i'll fumble throught first.

cybercop23:
> OMG our CustomModel XML is horrible. Kinda wish we had done X/Y pairs for the spots where pixels exist instead of a damn comma for every cell.

Hmmm... interesting.. yeah.. especially with the HD props that they make the w/h so big to get better model resolution.... Shouldn't be that hard to move it over, should it?

cybercop23:
Like this.... if Depth=1;step 3 else step 4
image

cybercop23:
> I took a quick look. There is still a ton of work to do. For each model you need to look at the class definition to figure out which screen location class it uses then you'll know which class attributes need to be added. For instance a TreeModel inherits from MatrixModel and that class is "class MatrixModel : public ModelWithScreenLocation(BoxedScreenLocation)" So that means it needs the BoxedScreenLocation attributes. That function does not exist yet. Also the ThreePointScreenLocation models will all need the TwoPointScreenLocation attributes. You want the functions to only add the attributes that are in each class. Every model gets the AddModelScreenLocationAttributes because they are common to all.
> 
> So we need all the following new functions: AddBoxedScreenLocationAttributes AddTwoPointScreenLocationAttributes AddPolyPointScreenLocationAttributes and I'll need an AddMultiPointScreenLocationAttributes but I can add that when I make that model available.
> 
> Then we get into the Objects like (RulerObject, TerrianObject, MeshObject, ImageObject, GridlinesObject, etc.) Example: TerrianObject is class TerrianObject : public ObjectWithScreenLocation(Terrian
[GE Rosa Grande.txt](https://github.com/xLightsSequencer/xLights/files/15287475/GE.Rosa.Grande.txt)
ScreenLocation) TerrianScreenLocation inherits from BoxedScreenLocation so we would need to call functions for both those classes to get all the screen location attributes as well as the common screen location attributes function.
> 
> Candy Cane is a ThreePointScreenLocation so that one should work. I'll see if I can try it out since you said that was failing.

I added another parm to the 3pt and if it gets 3, it will then add the 3pt specific attribures, angle & shear. That also takes care of poly.
Not sure where to start on the 3d obj. IDK if that's easier for you to take care, or guide me. I haven't figured out a way to export them like the models so I can check.

Also, I have the "compressed" custom model (CustomModel2.) format and for GR alone, that's a reduction of 94% in character size; 289,616 -> 16,618. If you want to make this change, I can start taking a look and add import/ini logic.

cybercop23:
| -> Type 1: (use this type for custom models that have continuous numbering with no repeats and no gaps)
[compression type], [node count], [reserved1], [reserved2], row1, col1, row2, col2, row3, col3, etc.

Hmmm... did you mean r1,c1,r2,**c1**,r3,**c1** What'a an example for this? With Matrix, I've added individual start nodes on Circle and Star (PR done) and will Do it on Matrix also, thus no need to  turn a matrix into a CM.

cybercop23:
Okay.. done... How do we choose between type 1 and type 2? Look at Node count and HxW and if all filled we go with 1? Or since I have both.. very small precessing... I can always pick the smallest string length.

RG CM=289,616, CM2.0.1=10,776, CM2.0.2=16,629

image

cybercop23:
> There's going to be some models that use the same node number multiple times like 8 channel singing face models so they would need to be type 2. So maybe a function that checks for duplicate numbers or gaps

KK.. CM has IsAllNodesUnique() so I can use that to check

cybercop23:
image

cybercop23:
> By changing the CustomModel field, you are essentially changing the public API for the models that get posted on prop vendor websites. If you move forward with this, then I would request that you make the type field negative values (-1, -2). That way, if the CustomModel field starts with a "-", it could be interpreted with the new logic, and if not, the old logic could be used.

There will be support to automatically convert from the current format to the new one, once the layout is saved. Import will still take the existing format, but once saved in the current show folder, it will become the new format.

cybercop23:
Well.... well... well.... did not expect this.... mindblown
image

I have a version that can now import the new "compressed" format. I'm working on supporting depth, but anything with type 1 or 2 & single depth works. Those are the results above.
For depth... since who know what they heck they are doing, I'll treat it as type 2.

cybercop23:
> So a huge speed increase on the loading times? I'm also expecting a nice increase in overall program speed once we can eliminate all the internal ModelXML stuff cause its got to take a lot of time to constantly be processing XML strings.

Not only load times, but also reinit on any layout tab change.... been spending lots of time to optimize whatever code I'm adding, but this is just crazy...

cybercop23:
@AzGilrock any suggestions/ideas if/how to change the data that's showing up in the Model Data grid? Looks like it wants the csv format to populate the notebook page, so I'd have to expand it for that function, or maybe/somehow use a different process...

`std::string CustomModel::GetCustomData() const
{
    return ModelXml->GetAttribute("CustomModel").ToStdString();
}`
...
`std::string data = m->GetCustomData();
 _saveModelData = data;`
....
` UpdatePreview(_saveWidth, _saveHeight, _saveDepth, _saveModelData);`

AzGilrock:
This work was merged into the XmlSerializer branch which has now morphed into the removeModelXml branch so I'm going to close this PR.

Issue #4408, opened 2024-03-15
Nested Group Display in Sequencer


CatalinaLights:
**Is your feature request related to a problem? Please describe.**
I love it that when Groups are expanded in the Sequencer View the items included in the group indent to the right as an indication that they are part of the group above. Expanding the sub-group contents does not further indent its collection of items, making it difficult to determine if they are directly in the top-level group or the sub-groups. Several times I have accidently added an effect to what I thought was a model within a sub-group, but was actually adding it to a model in the top-level group and vice-versa.

**Describe the solution you'd like**
I think it would be very useful if when groups are included in other groups, expanding the top group and then the sub-groups would indent the sub-group items even more to the right, making it clearer that they are part of the group above them. 
On the attached sketches, the red outline is the top level EVERTHING group, green are the sub-groups, and blue are the individual models within the sub-groups. I would like to see the blue-circled models indented to the right, somewhat like is shown on the Nested Groups-Enhanced image. 

![Group Nesting-Current](https://github.com/xLightsSequencer/xLights/assets/110269429/6d5c5f2f-bde4-432d-b8e9-52923f8da98d)
![Group Nesting-Enhanced](https://github.com/xLightsSequencer/xLights/assets/110269429/b0daf9aa-f55e-4133-8a3e-e26e087244ac)
[Nested Group Display.zip](https://github.com/xLightsSequencer/xLights/files/14612692/Nested.Group.Display.zip)

**Describe alternatives you've considered**

**Additional context**

PR #4549, opened 2024-05-07
Clean up the indentation  of the sequence tab #4408

derwin12:
#4408 Attempt to clean up the display and properly indent the groups, models, strands and nodes.

### DIFF
--- a/xLights/sequencer/RowHeading.cpp
+++ b/xLights/sequencer/RowHeading.cpp
@@ -1510,13 +1510,17 @@ void RowHeading::render( wxPaintEvent& event )
     }
     wxPen effectNoticePen(effectNoticeColor);
     wxBrush effectNoticeBrush(effectNoticeColor);
-    
+
+    wxString groupPrefix = "";
+    wxString modelPrefix = "";
+    wxString submodelPrefix = "";
+    wxString strandPrefix = "";
+    bool isExpanded = false;
     for (int i = 0; i < mSequenceElements->GetVisibleRowInformationSize(); i++) {
-        Row_Information_Struct* rowInfo = mSequenceElements->GetVisibleRowInformation(i);
-        wxString prefix;
         wxString layers;
-        if (rowInfo->submodel) {
-            prefix = "  ";
+        Row_Information_Struct* rowInfo = mSequenceElements->GetVisibleRowInformation(i);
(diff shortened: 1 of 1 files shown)


derwin12:
> I've always thought the layer number looked weird when expanded since the first line is a total and my eyes always think it should look like 1, 2, 3

Agreed but the first line is [3] total layers .. perhaps flip that to [1] when expanded? would that change confuse people??

derwin12:
> I can't really tell what's different. And for instance is the "Single Line" model part of the "Group with One Item"? Maybe the name threw me cause its looks like a lot more than one item. The Single Line is indented making it look like its under that group but the group doesn't have the minus box so I'm confused.

[-] is only to collapse layers.. BUT yes this is a bug .. going to fix that up
The first single line should be indented from the group but not the models sspinner and the model itself single line.
Expanded..
image
Collapsed..
image

cybercop23:
> Hmm....so not sure why we don't use the +/- to expand/collapse groups. Not sure sure about the layer numbering. We should number from the bottom up and say its the render order. :)

Ohhhh.. I like that.... it will be the truth.... and the 3 on the group/model can stay as it... NICE

deplanche:
But Layer Blending is based on the current layer being Layer 1 and the layer below that being Layer 2.  If they are numbered bottom to top, won't that confuse people even more?

derwin12:
Thoughts on this? 
The numbering ... which way .. this may be something we just don't touch? - as it is built in for canvas mode and such..
Do I bother coding the Empty group to not show the [+] - like who creates empty groups besides me during testing..
What if the closed [3] changes to [1] then  [2] then [3] when expanded??
image

derwin12:
Here is a new look .. shows 1.2.3 when expanded
image

cybercop23:
I do really like the idea of showing them in the way they render... from the buttom up. Layer Belnding is done at the effect level, so it should not interfeer with how we display and manage the view itself.

derwin12:
Here is another rendition .. Not perfect - groups in groups are not (easily) determined for indenting. Better than before.
Notice I moved the group open and close indicator over to the right. The left side is/was required to open/close the layers.
You still just double click on it to open, but the triangle gives some clues as to what state you are in.
image

Subtle differences from current .10 release..
image

PR #4554, opened 2024-05-07
Allow Individual Start Nodes to Circle and Star

cybercop23:
There will be part 2 for the Matrix

There's a new read only Attribute, TotalNodes, that's the sum of all layers. Node counts comes from that sum
Added logic (best it can do for abckwars support) to set parm1/parm2 based on any changes to layers and ports
Allow user to enter nodes for a single layer -> since that's where the total nodes come from

### DIFF
--- a/xLights/controllers/ControllerUploadData.cpp
+++ b/xLights/controllers/ControllerUploadData.cpp
@@ -76,7 +76,11 @@ UDControllerPortModel::UDControllerPortModel(Model* m, Controller* controller, O
     }
     else {
         _startChannel = _model->GetStringStartChan(string) + 1;
-        _endChannel = _startChannel + _model->NodesPerString(string) * _model->GetChanCountPerNode() - 1;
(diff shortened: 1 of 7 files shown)


dkulp:
I kind of think this needs a little work, but possibly just in the order of the properties in the panel.    It seems like the definitions of the strings is done before the number of nodes is set which then doesn't really work.  I'm wondering if the "Layers" setting should be moved up to the top (and Total Nodes then just below it) so you define the "Star" first, then set the string properties for how it's connected.

cybercop23:
Does this work? Or move string even further down, maybe even after Controller?
image

cybercop23:
Okay.. so something broke with switching those around (unlikely) or I screwed up my original commit and things don't work as they should. Digging into it.

cybercop23:
Figured out the issue, it was due to the redesign and moving string definition around, plus fixed custom model and make resetting to original format easier.

Issue #2785, opened 2021-08-31
Props not showing up in display elements when there is a timing track with an identical name.


MDickson00:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

When there is already a timing track in the sequencer with a name (suach as Santa) if there is a prop with an identical name, it will not show up in edit display elements. Alternatively, if there is a already a prop with a name (sucha as Santa) and you create a new timing track and name it the same name (such as Santa) the timing track name will save as Santa_1.

1. Go to sequencer
2. Start a new animation
3. Create a timing track and name it Santa
4. Go to your layout and create a prop named Santa
5. Go back to sequencer, click edit display elements
6. Notice the Santa prop is missing from display elements
7. Alternaitvely, create a prop named Santa
8. Go to sequencer and creat a new timing track and name it Santa
9. Notice it will change the timing track name to Santa_1

Including a video below
https://youtu.be/klJwRk7xYR0

Also attaching an empty sequence with a prop named Santa
[xlights bug.zip](https://github.com/smeighan/xLights/files/7080570/xlights.bug.zip)

cybercop23:
Should be available in the nightly and next version.

PR #4567, opened 2024-05-16
Check that timing track name doesnt conflict with models #2785

derwin12:
If a timing track is named the same as a model, it is hidden. Checks were in place to catch a rename but were not done on the initial Add Timing track option. #2785

### DIFF
--- a/xLights/sequencer/RowHeading.cpp
+++ b/xLights/sequencer/RowHeading.cpp
@@ -693,14 +693,19 @@ void RowHeading::OnLayerPopup(wxCommandEvent& event)
                 } else if (selected_timing == "Empty") {
                     bool first = true;
                     wxTextEntryDialog te(this, "Enter a name for the timing track", wxGetTextFromUserPromptStr, selected_timing);
+                    selected_timing = RemoveUnsafeXmlChars(selected_timing);
                     OptimiseDialogPosition(&te);
-                    while (first || xml_file->TimingAlreadyExists(selected_timing, mSequenceElements->GetXLightsFrame()) || selected_timing == "") {
+                    while (first || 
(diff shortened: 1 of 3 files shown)


derwin12:
oh for sure .. cant get them all but trying to close the one door...

cybercop23:
SequenceElements/AddElement is called with "model" type that then calls ElementExists. We can pass _type_ to ElementExists and check that types also matches, thus allowing same names on different types, timingtracks, groups and models.
2 other functions will need to have the type added since they call ElementExists, but IMHo I don't think those 2 need to check at all - will be okay if we add the type, but maybe reduce a call.

cybercop23:
> SequenceElements/AddElement is called with "model" type that then calls ElementExists. We can pass _type_ to ElementExists and check that types also matches, thus allowing same names on different types, timingtracks, groups and models. 2 other functions will need to have the type added since they call ElementExists, but IMHo I don't think those 2 need to check at all - will be okay if we add the type, but maybe reduce a call.

I was wrong... as usual.. at least ViewsModelsPanel need to have the check, else all modes will still show up on the left... small change.. but adding type.. seems to work...
image

derwin12:
I believe there are a more than a few places where it assumes it is getting a sequence or a timing track when it does the lookup.  As Gil mentions .. we search by name. Would take going thru with a fine tooth comb.. Allowing both is a no-going-back change. Anybody using older versions could fall victim to duplicate names. I dont think many people actually hit this scenario .. ie naming a timing track the same as a model, so not sure it is worth the efforts.

cybercop23:
Please tell me what else I should look at... new seq, views changes, adding/renaming TT and adding same name models works... I can try groups.. what else should I try?
YES... there may be crap out there not using this proper function, but what you found and what I looked at, they all use the ElementExists function, and that has been changed to account for type.

PR #4570, opened 2024-05-19
Add sACN/ArtNet support for ILightThat controller

PeteMatthews:
Also cache and restore test colour for models, so they're not lost whenever we push config from xLights.


xLights/controllers/ILightThat.cpp
```diff
@@ -87,6 +108,14 @@ bool ILightThat::SetOutputs(ModelManager* allmodels, OutputManager* outputManage
                     brightness = 100;
                 }
                 
+                if (model_test_cols.find(model->GetName()) == model_test_cols.end()) {
```
computergeek1507, line 111: this is real nitpicky(and xLights is not consistent), You have inline braces at the beginning of the if but then use newline braces at the else. I would inline it all to be consistent.

PeteMatthews: Entirely fair :) Reworked this and a few others to keep the consistency

PR #4577, opened 2024-05-23
Added ability to load secondary audio files and use them with VU meter

computergeek1507:
Need to integrate with Audio Value Curves and other effects that can use audio.


derwin12:
Would this be just one or multiple? Looking for perhaps the ability to add say drums, piano, guitar etc that we can get from various AI platforms.

cybercop23:
Multiple, and you can point the effect to either Main, or choose any other from a drop-down menu. I think you can install that version from the assets link.

derwin12:
I think we have this now ..

Issue #4568, opened 2024-05-18
Submodel dialog - buttons can get squished..


derwin12:
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/905e9216-fdee-4af8-b017-aa3af38a6f2a)

cybercop23:
I know the ledt side is so wide because of the buttons... they all need work...
We have the icons now but there may be an issue with using them and how we install them on the different OSs.

PR #4580, opened 2024-05-25
Keep buttons visible in submodel dialog #4568

derwin12:
Trigger the resize when the splitter is moved. #4568


computergeek1507:
Looks like you neutered the stretchy grid class so it probably isn't needed anymore

derwin12:
> Looks like you neutered the stretchy grid class so it probably isn't needed anymore

I did but couldnt really figure out how to remove it.. Feel free to grab the branch or maybe provide some howto?

derwin12:
Removed the StretchGrid class .. the grid itself stretches cleanly with the splitter window event.

computergeek1507:
In codeblocks change the class name to wxGrid

derwin12:
Done .. thanks.

computergeek1507 (approved):

PR #4584, opened 2024-05-29
Fix issue with miss-identification of Falcon boards in FPP connect.

cybercop23:
Falcon.cpp has ModelDecode as below, which is correct based on what ip/status.xml returns (ie: <p>130</p> for F16V5.
However FPPConnect was looking at the 88-91 range vs 80-93 to set up a Falcon V4or V5 board.
case 128:
    model = 16;
    version = 4;
    break;
case 129:
    model = 48;
    version = 4;
    break;
case 130:
    model = 16;
    version = 5;
    break;
case 131:
    model = 48;
    version = 5;
    break;

### DIFF
--- a/xLights/controllers/FPP.cpp
+++ b/xLights/controllers/FPP.cpp
@@ -3608,8 +3608,7 @@ static bool supportedForFPPConnect(DiscoveredData* res, OutputManager* outputMan
         return res->majorVersion >= 4 && res->mode == "remote";
     }
 
-    if (res->typeId == 0x88 || res->typeId == 0x89 ||
-        res->typeId == 0x90 || res->typeId == 0x91) {
+    if ((res->typeId >= 0x80) && (res->typeId <= 0x83)) {
         // F16V4 / F48V4 / F16V5 / F48V5
         return res->mode != "bridge";
     }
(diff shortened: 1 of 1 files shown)


cybercop23:
But those are not the right IDs based on Falcon.cpp and what gets returned by the discovery process with IP/status.xml

An F16/V5 returns 130 which is 0x82.

computergeek1507:
I think xLights needs to map the id of the falcon status page to the FPP id

dkulp:
On my F16v5, the discovery is setting the type to 0x90 which is correct:

Screenshot 2024-05-29 at 3 18 03 PM

dkulp:
And when in Remote mode, it properly appears in the list:

Screenshot 2024-05-29 at 3 21 39 PM

computergeek1507:
Is that the results from a FPPs Multisync results or directly from the controller. And is it over Ethernet or WiFi. We have seen different results over different interfaces, all the XML from the controller is different

cybercop23:
In John's case, we can get it to show up on FPP but never in FPP Connect. Ctrl is defined in xlights, Active, has all right info, and it is wired.

@dkulp what do you get from the ctrl status.xml 192.168.3.220/status.xml

dkulp:
Screenshot 2024-05-29 at 3 31 38 PM

Most likely, somewhere around line 577 of Discovery.cpp, if the cd->typeId == 0 then it should map the Falcon model/version  combo to the appropriate typeId.   I don't see it setting the typeId there so if it's not discovered via the normal multicast Ping packets, it's likely not being set and is defaulting to 0.

dkulp:
Also, is it NOT in DDP/e1.31/ZCPP modes?   That's important as well.

cybercop23:
> Also, is it NOT in DDP/e1.31/ZCPP modes? That's important as well.

No, they are in Remote mode with an SD card inserted, so we can upload sparse fseq data.

Issue #4587, opened 2024-05-30
Useless Feature Request In / Out Transition drop down sort to alphabetical


JoelDearing:
**Is your feature request related to a problem? Please describe.**
No

**Describe the solution you'd like**
within the layer blending window, for the in transition and out transition tabs, can the drop down selection that shows fade, wipe, clock, and so on be put in alphabetical order?

**Describe alternatives you've considered**

**Additional context**

cybercop23:
Should be available in the nightly and next version.

PR #4588, opened 2024-05-30
Sort dropdowns for In/Out Transitions and Layer Blending #4587

cybercop23:


xLights/TimingPanel.cpp
```diff
@@ -327,59 +327,59 @@ TimingPanel::TimingPanel(wxWindow* parent, wxWindowID id, const wxPoint& pos, co
     BitmapButton_Out_Transition_Adjust->SetLimits(OUT_TRANSITION_MIN, OUT_TRANSITION_MAX);
 
     // If i pull this from the map it sorts them alphabetically and I dont think that is what we want
-    Choice_LayerMethod->Append(_("Normal"));
```
dkulp, line 330: There is a comment here that specifically says that this is not what we want....  Transitions maybe, but not for the blending.

cybercop23: But it just says "I dont think that is what we want".....  dont think... so not firm...
Why won't we want to make it easier to follow the list? It's not breaking anything, as far as I can tell. Had several users in the Zoom room when we discussed this and consensus was it makes sense to be sorted. Again, if it breaks something, I haven't yet found it.

derwin12:
Easy to add a sort here? Timing tracks on the face effect.
image

cybercop23:
Should be available in the nightly and next version.

brett-foy:
Just downloaded last nights build.   The sort order on the transitions has changed my entire sequence.  Every effect that had Fade (was first on list) is now set to Blend.  I don't think this is the desired outcome?

MrPierreB:
Good spot @brett-foy 
Although, it is still using fade even though it is showing Blend.  Bigger issue is Blend actually doesn't work anymore.
Seems T_CHOICE_In_Transition_Type wasn't set in the xml when the very first option is selected from the dropdown, so the fix will mostly likely need be to default set the type to Fade if it is blank in the file, and then going forward also making sure the T_CHOICE_In_Transition_Type is always set. ... or revert the ordering - as fade is the most used transition, going to suck selecting it on almost every effect. :)

Definitely a blocker in the current state. @dkulp

cybercop23:
Fade should still be the default for any new sequence or when you drop the effect in the sequencer. It's just an option lower on the list. I'll have to take a look this morning, but I remember checking that was the case. 
I'll look at that TIn/Out too.

Thanks much for testing the nightly and finding and reporting.

cybercop23:
Fixed.

MrPierreB:
@cybercop23  Not fully fixed yet.  The "Blend" transition is not working.   When selected blend is selected it is fading.

MrPierreB:
Also, if you select Blend as the transition, then unselect the effect and reselect it, it will change the fade again.

cybercop23:
How have you tested the latest? Using TestFlight? Or downloaded the build, or built your own?

cybercop23:
So the issue id with the lack of default TransitionIn/Out values not being available. I can either move the Fade back to the top and have the rest of the list sorted (not my favor) or get all items/regardless of position/default to have values. Will try for this.

MrPierreB:
> How have you tested the latest? Using TestFlight? Or downloaded the build, or built your own?

I got the latest code and built it locally.

cybercop23:
OMG!!!!!!! WHY!!!!!!

`if (Choice_In_Transition_Type->GetSelection() > 0) {
    s += "T_CHOICE_In_Transition_Type=";
    s += Choice_In_Transition_Type->GetStringSelection();
    s += ",";
}`
So I removed the if and the **T_CHOICE_In_Transition_Type** now always get written. Also added code for existing ones to convert to Fade and on first save it will have the TT attribute.

MrPierreB:
> OMG!!!!!!! WHY!!!!!!
> 
> `if (Choice_In_Transition_Type->GetSelection() > 0) { s += "T_CHOICE_In_Transition_Type="; s += Choice_In_Transition_Type->GetStringSelection(); s += ","; }` So I removed the if and the **T_CHOICE_In_Transition_Type** now always get written. Also added code for existing ones to convert to Fade and on first save it will have the TT attribute.

loooooool :P
Im guessing the thought was fade will be used so much, it will make the size of the xml (xsq now) more  .. thats the only thing i can think of at the moment before the morning coffee kicks in....
anyways, glad you got it sorted :)

dkulp:
> All of our effects default values are not written to the XML that’s how we kept them smaller. Look how many characters are being added to every effect all because you guys get anal about sorting that everyone was fine with for 9 years

Yea.. I'm beginning to lean toward reverting the entire thing.....

cybercop23:
Well.. the only way is to leave Fade at the top and sort the rest. I think we should still sort the list, but frankly, the XSQ is so small that I don't think 35 characters/fade would be a problem.  Nowadays disk is cheap and plentyfull so that shouldn't be a concern.

computergeek1507:
The change ruined my muscle memory of selecting fade in and out

Issue #4593, opened 2024-06-06
Middle Mouse Button Functionality for Sequencer Navigation.


ak463010:
Subject: Feature Request - Middle Mouse Button Functionality for Sequencer Navigation

Dear xLight Team,

I hope this message finds you well. I am writing to suggest a new feature that I believe would significantly improve the user experience in our sequencer.

**Feature Description:**
I propose the implementation of a feature that allows users to navigate around the sequencer using the middle mouse button. This would involve clicking and holding the middle mouse button and then moving the mouse to pan around the sequencer.

**Benefit:**
This feature would provide a more intuitive and efficient way for users to navigate large and complex sequences. It would align with common practices in many other software applications, making our sequencer more user-friendly, especially for new users.

**Implementation:**
While I do not have a detailed understanding of our codebase, I believe this feature could be implemented by capturing the middle mouse button click event and translating the subsequent mouse movements into sequencer navigation commands.

I look forward to hearing your thoughts on this proposed feature. I truly believe that it would be a valuable addition to our sequencer.

Best regards,
Avinash Kumar

cybercop23:
Should be available in the nightly and next version.

PR #4597, opened 2024-06-09
Add middle mouse to pan sequencer #4593

derwin12:
With a sequence open, on the sequence panel add the ability to middle click, hold and pan to move the sequence window. #4593

### DIFF
--- a/xLights/sequencer/EffectsGrid.cpp
+++ b/xLights/sequencer/EffectsGrid.cpp
@@ -64,6 +64,8 @@ EVT_LEFT_UP(EffectsGrid::mouseReleased)
 EVT_MOUSE_CAPTURE_LOST(EffectsGrid::OnLostMouseCapture)
 EVT_RIGHT_DOWN(EffectsGrid::rightClick)
 EVT_LEFT_DCLICK(EffectsGrid::mouseLeftDClick)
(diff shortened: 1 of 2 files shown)


derwin12:
Wait .. perhaps the action built version only included the 2nd commit .. so it only did the refactor in the .exe.
@computergeek1507 Scott - would that be the case? would it be the combined commits or just the last one that was sent in?

cybercop23:
They'd had to be stacked.. else your 2 new functions, mouseMiddle[Down|Up] would have failed with no function definition, no?

computergeek1507:
I think Alex found out that the pull request build is actually using master not the pull request id. I haven't fixed it.

cybercop23:
> I think Alex found out that the pull request build is actually using master not the pull request id. I haven't fixed it.

ohh yeah.. forgot about that.... :)

dkulp:
I merged this, but the two finger pan on the touchpad is WAY WAY smoother.   Likely should look closer at that as the base.

derwin12:
> I merged this, but the two finger pan on the touchpad is WAY WAY smoother. Likely should look closer at that as the base.

I wonder if that is a wxwidget builtin shortcut to moving the bottom scrollbar. I didnt see any code there that did the two finger pan. I will investigate further.

dkulp:
It's PROBABLY part of MainSequencer::mouseWheelMoved.

PR #4602, opened 2024-06-12
4587 sort transitions and belnding

cybercop23:
Fixed issue with Default values for FadeIn/Out


xLights/TimingPanel.cpp
```diff
@@ -427,8 +427,8 @@ void TimingPanel::SetDefaultControls(const Model* model, bool optionbased)
         TextCtrl_Fadeout->SetValue("0.00");
         Slider_In_Adjust->SetValue(50);
         Slider_Out_Adjust->SetValue(50);
-        Choice_In_Transition_Type->SetSelection(0);
-        Choice_Out_Transition_Type->SetSelection(0);
+        Choice_In_Transition_Type->SetSelection(10);
+        Choice_Out_Transition_Type->SetSelection(10);
```
dkulp, line 431: I would think SetStringSelection("Fade") would be a better option here.   If additional transitions are added, the "10" may need adjusting again.

Issue #4603, opened 2024-06-12
add the ability to reverse the 3d fade instead of from color to black, but from black to color for single strand effect 3d fade


sidelights:
**Is your feature request related to a problem? Please describe.**
Not tied to a problem.  

**Describe the solution you'd like**
The single strand always starts is fade at the head of the string. like a comet.  I had an idea for a sequence where I wanted to have more of a blend up to bright.  I found there was not an easy way to do this consistently for a specific number of cycles

**Describe alternatives you've considered**
Use of bars effect, morph, colorwash etc.    I did find that I can get the results I wanted using the static left-right and adjusting the offset to effectively scroll the single strand backwards.   This however is very time consuming to align the start and end to beats or existing effects that follow cycles. 

**Additional context**
I was able successfully adjust the code in a local repositiory to accomplish reversing the starting with dark going to color, within standard code, and will be available for review.

PR #4604, opened 2024-06-12
#4603 Add Reversed 3D fade on Single Strand

sidelights:
Allow for 3D fade to fade from black to the selected color instead of from the color to black within the single strand effect.  While keeping the number of cycles built in to the single strand effect already.    Also added tooltips for the single strand panel. #4603

### DIFF
--- a/xLights/effects/SingleStrandEffect.cpp
+++ b/xLights/effects/SingleStrandEffect.cpp
@@ -179,6 +179,7 @@ void SingleStrandEffect::Render(Effect* effect, const SettingsMap& SettingsMap, 
                                 GetValueCurveInt("Color_Mix1", 10, SettingsMap, eff_pos, SINGLESTRAND_COLOURMIX_MIN, SINGLESTRAND_COLOURMIX_MAX, buffer.GetStartTimeMS(), buffer.GetEndTimeMS()),
                                 SettingsMap.Get("CHOICE_Chase_Type1", "Left-Right"),
                                 SettingsMap.GetBool("CHECKBOX_Chase_3dFade1", false),
+                                SettingsMap.GetBool("CHECKBOX_Chase_3dFade1_Rev", false),
                                 SettingsMap.GetBool("CHECKBOX_Chase_Group_All", false),
                                 GetValueCurveDouble("Chase_Rotations", 1.0, SettingsMap, eff_pos, SINGLESTRAND_ROTATIONS_MIN, SINGLESTRAND_ROTATIONS_MAX, buffer.GetStartTimeMS(), buffer.GetEndTimeMS(), SINGLESTRAND_ROTATIONS_DIVISOR),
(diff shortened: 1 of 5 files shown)


computergeek1507:
Gross more checkboxes

derwin12:
Brian .. seems you need a little help from the admins to allow your code to compile... @computergeek1507

sidelights:
that would be great what I can do to help make that happen?

dkulp:
I kind of agree that another checkbox for this doesn't make sense.  I think I'd prefer that the "Fade" turns into a choice with a few options of {"None", "Tail", "Head", "Both"}.   That would require a bit more work as the older sequences would need to be updated.

sidelights:
With additional options it would be better as a choice.  I was only considering adding the reverse fade originally, and that would just be something new not requiring any conversion to existing sequences.   It made more sense as an additional box because it only would exist when you had 3d fade enabled, then flip it.  Adding additional options like both or even reverse both (logically) would be better in a drop down.    I was also pondering the bars effect and the spiral which also has similar logic for 3d fade and if that would be 'nice' there or not.     I figured instead of just saying to a developer hey I think this would be neat, could you make it for me, I would try to contribute help as well instead of just ideas.  

With the options described then it turns into the following I believe:
new 'FADE From Choice box (default none) -- existing 3d fade box unchecked. 
Choices: 
New name   -   OLD name  -   visual for communication here  (from left to right) maybe in a tool tip? 
------------ 
"None"              -  n/a unchecked -     n/a
"From Head"            -   3d fade            -    ++---------
"From Tail"               -   3d reversed      -    ---------++
"Head and Tail"        -   n/a aka both    -    ++ --- ++
"Middle"                 -    n/a                   -    ---+++---


Also I could not trigger a way to get the non- alpha buffer blend (HSV) to trigger in the code.  Does anyone know what causes that logic path to be available.  If I'm going to rewrite to allow additional options I would like to know how to trigger hsv so it can be validated well.

sidelights:
I can be in the zoom room tonight after 7pm central if you want to demo what you are talking about Gil.  I attempted to do what I think you are attempting to describe and I could not get it to do what I can do with the additional fade options proposed with increasing options under the existing 3d fade functionality in the single strand.  But I'm happy to learn something new if I missing something. 
Thanks!

dkulp:
> It will depend on how the effect is using the feature. Search for IsSpatial and GetSpatialColor to find the effects that support it. It can be added to effects that don't utilize it.

.... with the caveat that as soon as spacial colors are used, rendering on the GPU is not possible at this point so it's stuck rendering on the CPU.

sidelights:
single strand does have spacial, and using the -> or <- against the color does fade the direction with a gradient over the chase length.  However the HEAD of the single strand when in reverse (faded out) overlays the tail causing a black 3rd pattern over the single strand.  see video comparing the reversed 3d fade (head dark to tail bright), and 3d fade (orig - head bright tail dark), and the reversed 3d fade , to the color fade spatial as you mentioned
[rervese 3d fade vs spatial - Copy.zip](https://github.com/user-attachments/files/15828899/rervese.3d.fade.vs.spatial.-.Copy.zip)
.   Basically it looks like it keeps the 'head' as the most prominent pixel.

Issue #4605, opened 2024-06-12
Add option to duplcate effect to retain duration


derwin12:
When using duplicate effect, if you have it set to cell add the ability to retain the effect duration rather then the default of expand for the duration of the timing mark.
Enhancement to simply copy that effect and retain the sizing to each of the timing marks upto the # set in the Count parameter.

From:
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/91d79340-241f-4332-8b42-55fb537338e6)
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/e1e290f0-09b4-4481-bc7e-58738bf98ad1)

Currently:
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/44f9732a-46ec-46aa-a5e3-dc11a7f1a0af)

PR #4606, opened 2024-06-13
#4605 Optionally retain the effect duration when duplicating an effect

derwin12:
In the right click duplicate effect menu, a new checkbox (@computergeek1507) will optionally make the duplicated effects retain the same effect length rather then expand the effect to the max between the timing marks. This is good, for example, to create a shockwave of .5s at each of the various timing marks.


derwin12:
> Seems like it could use the Paste by Cell and Paste by Time options already available
Slightly different but significant...
By Time:
image

By cell:
image

Results with the new option:
image

derwin12:
@computergeek1507  Scott the xLIghts CI merge build fails .. doesnt get the branch? It assumes the branch name is simply the PR#? (just a guess on my part)

PR #4607, opened 2024-06-13
 Always save T_CHOICE_In_Transition_Type as attribute and upconvert old xsq to Fade type

cybercop23:
This time for real


xLights/TimingPanel.cpp
```diff
@@ -481,11 +481,9 @@ wxString TimingPanel::GetTimingString()
     // Fade in
     if ("" != TextCtrl_Fadein->GetValue() && "0" != TextCtrl_Fadein->GetValue() && "0.0" != TextCtrl_Fadein->GetValue() && "0.00" != TextCtrl_Fadein->GetValue()) {
         s += "T_TEXTCTRL_Fadein=" + TextCtrl_Fadein->GetValue() + ",";
-        if (Choice_In_Transition_Type->GetSelection() > 0) {
```
dkulp, line 484: Why not just a GetStringSelection() != "Fade" ?    We shouldn't be putting default values into the xsq.

cybercop23: Okay... I see.. that may work.. so now we don't put Fade.. we'll only write it for nonFade and then can remove the upconversion... nice, thank you

PR #4608, opened 2024-06-13
Move Delete to lower in the list to reduce accidental click

cybercop23:


derwin12:
Well there is a confirmation prior to deleting it .. but is it alphabetical? /s

cybercop23:
Yeah, apparently double clicking too fast, gets past that. There's also the undo, but IDT delete should be at the top of the list. Now is nicely nested in the *it will never be sorted* menu list.

derwin12:
Perhaps we can add a captcha to ensure you really want to delete the model :-)

Issue #4601, opened 2024-06-12
Upload Outputs does not work for Advatek E4-S Mk3 Controller.


mjunek:
So far I have encountered two issues with attempting to upload to this controller.

Issue#1 - A wireshark of the traffic is showing that the initial request to ```/ver``` is being sent with a MIME type of ```application/x-form-urlencoded``` rather than ```application/json```; despite the actual data being transmitted being in JSON format.
I have tested by making changes on lines 1132, 1195, and 1519 of ```PixLite16.cpp``` to specify the JSON mime type; which proves successful in negotiating the /ver request and getting the Hardware and API version details.

The next part is the new configuration endpoint ```/v1.5```. This appears to treat JSON messages differently to previous versions.

For example, the first complaint the E4S has about the upload is the DMX structure:
```
"auxPort": {
    "dataSrc": [
        "sACN"
    ],
    "mode": [
        "Off"
    ],
    "uni": [
        1
   ]
}
```
When MODE is set to OFF, the ```dataSrc``` and ```uni``` fields should not be present for the upload to succeed.
It also complained about the ```zigZag``` parameter passed for the pixel outputs being zero (not in range of 1-50).
I've tested both of these changes in the Boomerang REST client and the API call succeeded.

Unfortunately I'm out of my depth with coding in CPP and submitting a PR to get this resolved further (and not break other Pixlite users), but I'll continue to investigate the changes in API calls for the v1.5 endpoint.

**Versions**
 - OS: Tested on both Windows 10 and Windows 11
 - xLights version 2024.10
 - Advatek Firmware version 3.4.0

keithsw1111:
There is not enough information here. Many of these settings are arrays ... what if there is zero pixels on one port but pixels on another ... what should zigzag look like then?

Auxport is ok as there is only ever one port.

keithsw1111:
Also the /ver is a get not a post so it can't be json

mjunek:
Agreed @keithsw1111 - Working on getting more detail around the formats / where it's not quite right. Just posting this to start with as I figure the bits out in regards to it in case others find this controller doesn't work with upload.

My initial testing showed the default zigZag count (from xLights) was set to 1 on the unused port, and 0 on the used ports. Setting it to 1 on all ports resolved that particular issue. 

And yes, you're right, I got confused on the /ver - it was the first post to /v1.5 that was going with the wrong MIME type. I've been playing with it on and off this last week and that was one I fixed early in the piece.

[...]

PR #4613, opened 2024-06-14
#4601 Fix JSON & ZigZag parameters for PixLite E4-S v1.5 firmware

mjunek:
Fixes issue in #4601 - ZigZag must be specified in the range of 1-50 in the new firmware versions; and requires JSON format. 
Untested changes put in by Keith didn't work on the MIME type.


keithsw1111:
This is not the right fix. This will impact the old pixlites. The change needs to be made where I made it.

derwin12:
Keith implemented fix.

Issue #4625, opened 2024-06-19
Error when merging models, no group merging in 2024.11.1


Clyde-L:
We've found a bug in 2024.11.1 when replacing a model with a model.  What should happen is, if a model in a layout is attached to groups, and you import the same model with different group; then the groups from the imported model should merge with existing groups.  We found this today while trying to update a layout with a new model with different groups than the user already had.  The imported model groups were not merged when replacing the model.  Please see video link below for full demo.

I'm not sure when this broke.  I was able to revert to 2024.2, and process of replacing a model with different groups, and applying it to the existing mode, easily merges the groups from different sequence vendors.  

We also found that reverting the layout from 2024.11.1 back to 2024.2, the moving heads would freeze layout opening process and cause xLights to hang-up.  We had to create a copy of the layout, delete the moving heads, and then revert to the older version of xLights in order to test this.

Any help to get this fixed will be greatly appreciated.

Thank you for looking,
Clyde

https://www.dropbox.com/scl/fi/te4x54vkuiaa6xqrocwgg/Model-Import-Replace.mp4?rlkey=wl8rt4mbd23z4mvtwinznkgl7&dl=0

Clyde-L:
Gil that's reasonable.  Any idea what happened with group merging?

derwin12:
Same in .10. 
I will go back a few versions and see if it was any different.

derwin12:
Narrowed down to these .. seems like that 3rd last one is the likely culprit
![image](https://github.com/xLightsSequencer/xLights/assets/4643499/9afb6559-2679-4f5e-af59-bd6dcb8a8a4d)

PR #4627, opened 2024-06-21
Fix issue with replacing models/groups and duplicates #4625

cybercop23:
RemovedDuplicated updated modelNames but wasn't used. Old vewVal was still being passed. Now we're updating newVal after Dedeup detected


derwin12:
Is "<<" valid/safe for a wxString? or should it be +=

PR #4630, opened 2024-06-23
ILightThat controllers: Set universe size in config

PeteMatthews:
Add ability to set universe size for ArtNet and E1.31 
Add default model colours for new models


PeteMatthews:
Not clear why the Windows check failed - looks like something git-based, and it passed on my form pre-PR. If someone could retry it, that would be lovely :)

PeteMatthews:
Ah, hadn't realised that all Win build checks are currently failing on PRs... Ignore me :)

PR #4640, opened 2024-07-01
ILightThat Controllers - Minor internal rebrand

PeteMatthews:
Moving the branding on current board - Have to add the new name here first, or discovery will break as we go. Will remove the effective duplicate when we're convinced most people have the new firmware.


PeteMatthews:
Added additional commit, adding new board project.

Issue #4650, opened 2024-07-14
MHA - Reset to defaults


derwin12:
Resetting on the Status page.. should it clear out the other values? Suggest it should..
![image](https://github.com/user-attachments/assets/30d77cd1-0f39-4d12-84f7-49eb4c53855d)
![image](https://github.com/user-attachments/assets/4b0d25e9-695d-4564-ac90-7d311e2e7fcb)
![image](https://github.com/user-attachments/assets/2f620e14-faac-49b9-950e-cc9220a532e3)
![image](https://github.com/user-attachments/assets/3570a974-e1fa-42ea-9045-256aa98d1d9d)

cybercop23:
Done in 13

PR #4682, opened 2024-07-28
Reset color panels when you Reset to Default #4650

derwin12:
Clear out the color rgb and wheel choices when you Reset to Defaults #4650


computergeek1507:
You spelled color wrong

derwin12:
Funny .. I just followed the trend .. wonder how u got in there!

PR #4690, opened 2024-08-01
Removed auto MH selector when clicking on an effect until a better so…

cybercop23:
…lution found.


computergeek1507:
I dont understand what you are fixing with this? seems like the fixture checkboxes wouldnt be set now

cybercop23:
right. its not a total fix per say. i added that func, among others and notices under a cerrtain cenario is broken. so removing the auto check for now, since Dan is planning on a new ver very soon for the weekend mini. don't want a broek version out there with the new effect... but do want the other changes that were included... :)

derwin12:
Just wanted to add .. that the better commit is just to remove the code.

cybercop23:
Yes, but I'm woking on fixing it but wanted to make sure there's a band-aid before a new version comes out and renders (hehe) the effect unusable.

PR #4699, opened 2024-08-05
Automagically reduce MH effect to match number of configured fixtures.

cybercop23:
This allowes for easy import from 8->6 or 8->4 without the need to adjust each effect.
When the new effect is used right, after an import to less number of heads you'd have to do a dummy/fake change to get them to recalc and readjust to the lesses number of heads. Now,. you don't need to do that.

### DIFF
--- a/xLights/SeqFileUtilities.cpp
+++ b/xLights/SeqFileUtilities.cpp
@@ -40,9 +40,9 @@
 #include "xLightsMain.h"
 #include "xLightsVersion.h"
 #include "models/DMX/DmxModel.h"
+#include "models/DMX/DmxMovingHeadAdv.h"
 #include "models/ModelGroup.h"
 #include "sequencer/MainSequencer.h"
-
 #include "SequencePackage.h"
 #include "Vixen3.h"
 #include "effects/BarsEffect.h"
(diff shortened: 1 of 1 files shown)


MrPierreB:
I think rather than implement half or temporary solution we need to think of a more complete solution to mapping from the source sequence to our layouts. Not only going from 8 to less, but also from 4 to more - and also fixing existing issues on the import.
Looking at the code in this PR, there is a lot of string manipulation happening and also some magic numbers like substr 13, which usually isn't the greatest as when anything changes somewhere else it will start causing this to fail. Also there is no error handling so if it does fail, we don't want it to break imports completely.
I haven't run the code, but from looking at it I think there might actually be a bug here as well, when you write Heads: back - the values i suspect might not be correct if you map for example a single moving head effect placed directly on moving head 7, to your layout which only has 4 heads, and you mapped their moving 
head 7 to your moving head 3.

There is also the existing bug that if you for example map over moving head effects that are directly on MovingHead4 to your Moving Head 3 and you save - in the xml you will see the effect still has Heads: 4 so when you render and play, the moving head will not move as if the effect is not there. Only when you click on the actual effect in the sequencer does it autocorrect, and then when you save it has Heads: 3. Have to click on each effect one by one.

@AzGilrock  @dkulp 
Open to discussion on ideas here -  One way I think could be on the mapping screen, once you did your mapping selection and click ok, it checks if any of the source effects will be moving heads, if so it shows a new 2nd window that asks how you would like to map over x number of moving heads to your y number of heads.
This way you will have the x to y mappings so you can correctly set the Heads: property and also you can copy the effects on E_TEXTCTRL_MH from the correct option to the correct destination..   Thoughts?

MrPaulAR:
This has been open since Aug 20 with no other proposed solutions.  Since we continue to see this in the zoom room I vote to approve the PR and move forward.

AGFazio:
Also agree to move ahead. We are getting hammered with issues in the zoom room. Spending way too much time trying to do a work around.

computergeek1507:
I am working on a new feature to make this unnecessary

cybercop23:
Do tell please

Issue #4624, opened 2024-06-19
.11 Crash on Update All Aliases Using Mapping during the Sequence Import


derwin12:
I will try and track it down but here are the logs.
I was importing the BFL sequence from the Xlights Gems, mapped them over, saved the mapping then clicked the update alias mapping.

```
Walking from an exception.
[01]	ADDR:0x00007ff7eb4a2881	HMODULE:0x00007ff7eac30000	0000000000872840	Model.obj	?IsAlias@Model@@QEBA_NAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@_N@Z@61E971 
[02]	ADDR:0x00007ff7eb4a2e3e	HMODULE:0x00007ff7eac30000	0000000000872df0	Model.obj	?AddAlias@Model@@QEAAXAEBV?$basic_string@DU?$char_traits@D@std@@V?$allocator@D@2@@std@@@Z@61EF2E 
[03]	ADDR:0x00007ff7eb8d3ba0	HMODULE:0x00007ff7eac30000	0000000000ca38b0	xLightsImportChannelMapDialog.obj	?OnButton_UpdateAliasesClick@xLightsImportChannelMapDialog@@AEAAXAEAVwxCommandEvent@@@Z@A4FC90 
[04]	ADDR:0x00007ff7eb985ef4	HMODULE:0x00007ff7eac30000	0000000000d55ea0	wxbase33u:appbase.obj	?CallEventHandler@wxAppConsoleBase@@UEBAXPEAVwxEvtHandler@@AEAVwxEventFunctor@@AEAVwxEvent@@@Z@B01FE4 
[05]	ADDR:0x00007ff7eb97af4f	HMODULE:0x00007ff7eac30000	0000000000d4af00	wxbase33u:event.obj	?ProcessEventIfMatchesId@wxEvtHandler@@SA_NAEBUwxEventTableEntryBase@@PEAV1@AEAVwxEvent@@@Z@AF703F 
[06]	ADDR:0x00007ff7eb97ba9b	HMODULE:0x00007ff7eac30000	0000000000d4ba20	wxbase33u:event.obj	?SearchDynamicEventTable@wxEvtHandler@@QEAA_NAEAVwxEvent@@@Z@AF7B8B 
[...]

PR #4707, opened 2024-08-07
Fix crash on create aliases from mapping #4624

cybercop23:


xLights/xLightsImportChannelMapDialog.cpp
```diff
@@ -2706,12 +2706,17 @@ void xLightsImportChannelMapDialog::OnButton_UpdateAliasesClick(wxCommandEvent&
     _dataModel->GetChildren(wxDataViewItem(0), models);
     for (size_t i = 0; i < models.size(); ++i) {
         xLightsImportModelNode* m = _dataModel->GetNthChild(i);
-        if (m->HasMapping()) xlights->GetModel(m->_model)->AddAlias(m->_mapping);
+        if (m->HasMapping() && m->_mapping != "") {
```
computergeek1507, line 2709: ! string.empty() is preferred

Issue #4701, opened 2024-08-06
Copy/Paste Error Where Unrelated Prop Data is Erased


Bgoldie20:
**Describe the bug**
At random times, when using the AC lights toolbar, with the "Select" tool chosen, I will copy/paste two lines from two props and paste them to two other props. When pasting, it will erase a fifth line of sequencing that is seemingly unrelated to the copy or paste source or destination props.

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)
1. Click on AC lights tool bar
2. Use "Select" mode for mouse
3. Copy 2 or more lines from 2 or more props
4. Paste to new props at same timestamp
5. Observe another line's data get deleted

**Expected behavior**
Expect to paste data and not interfere with other lines/props.

**Screenshots**
Video link to the problem in attachments section.

**Versions (please complete the following information):**
 - OS: Windows 10, Version 22H2, OS build: 19045.4651
 - xLights version: 2024.13 64bit (no updates available as of 8/5/2024)

**Additional context**
None

**Attachments**
Video link: https://www.dropbox.com/scl/fi/r7ktr4kmbegsv9x3l2u4h/xlights-copy-paste-error.mp4?rlkey=vm7ff47gton1y8ykz5vtwjnz3&st=slcli660&dl=0

Log files: Please PM and I can provide the dropbox link with the sequence file.

keithsw1111:
This could be (in fact most likely is) happening due to model channel overlap. Without access to your layout it would be difficult to validate it.

Bgoldie20:
Thanks for the quick comment! Very possible, I don't have hardware in hand yet so I skipped the configuration steps for models/channels and likely made the error there. Once I start setting up hardware and correct the overlap, I'll report back if still seeing this issue.

Bgoldie20:
Here's the .fseq in case you're interested: https://www.dropbox.com/scl/fi/vomh8b14l02zb1t9y8rax/xLightsProblem_Copy_paste_error.zip?rlkey=ntozpcyb4ui60mwyarfj9n7ka&st=awu0n8ir&dl=0

derwin12:
I cant duplicate this..
Used your seq .. I did a render .. nothing on controller. Check sequence shows no issue.
How did you do this?
"Copy 2 or more lines from 2 or more props"
I did ctrl-x, ctrl-v.

[...]

PR #4714, opened 2024-08-11
AC copy and paste was clearing out the wrong rows #4701

derwin12:
Code is set to clear out the area where the paste is to take place. The row calc seemed to be incorrect. 
Also removed legacy comments.
@keithsw1111 please review. #4701


keithsw1111:
I dont know ... so long since i looked at that ... I assume you tested it with rows open, collapsed etc ... scrolled down ... not scrolled down.

derwin12:
Yes .. covered off those various  tests. Seems low risk and does appear to work.

Issue #4717, opened 2024-08-13
v2024.14 Virtual Matrix not rendering properly with simultaneous Picture effect and group effect


continja:
**Describe the bug**
In ver 2024.14.1, adding an image to a virtual matrix that is part of a group with the Shockwave effect applied reduces the size of the virtual matrix and cuts off the group effect. Works fine in ver 2024.12.1 and 2024.13.1. 

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)
1. Create a virtual matrix with # Strings = 50 and Nodes/String = 200.
2. Create a group (eg, 'all house') that includes the virtual matrix.
3. Create a new sequence. Add a Shockwave effect to the 'all house ' group on the timeline.
4. At the same position on the timeline, add a Picture effect to the virtual matrix.

This works as expected.

5. Change the size of the virtual matrix to # Strings = 51 (or higher) and Nodes/String = 200.

One or more rows are cut off at the bottom of the virtual matrix where the Shockwave effect displays, though the Picture is visible on the whole matrix. Continuing to increase the # Strings removes more rows from the bottom. 

**Expected behavior**
As occurs in v2024.13.1 and prior, the Shockwave group effect should be visible over the entire virtual matrix behind/blended with the Picture effect on the matrix.

**Videos**
1. Test_ver2024.12.1.mp4 - Virtual Matrix size set to my usual 270x480 using ver2024.12.1. Shockwave looks the same on the matrix, whether the Picture is on or not.
2. Test_ver2024.14.1mmp4 - Matrix still set to 270x480 using ver2024.14.1. Shockwave is fine when Picture is off. When Picture is on, most of the bottom of the matrix is cut off. At this size, it looks like the Picture is blocking the Shockwave, but it's not. See 4 below.
3. Test_ver2024.14.1_50x200.mp4 - Matrix size set to 50x200 using ver2024.14.1. Shockwave works as expected and looks the same whether the Picture is on or not.
4. Test_ver2024.14.1_70x200.mp4 - Matrix size set to 70x200 using ver2024.14.1. Shockwave looks as expected when Picture is off. When Picture is on, Shockwave is cut off several rows from the bottom.

**Screenshots**
1. Model Preview_50x200.png - Shows the Shockwave effect moving across the 50x200 virtual matrix behind a black box with a green border.
2. Model Preview_70x200.png - Shows the Shockwave effect moving across the 70x200 matrix behind the same Picture, with rows cut off the bottom.

[...]

PR #4720, opened 2024-08-13
Increase the max buffer wi and ht #4717

derwin12:
The group incl VM, is being processed as a single line and hence the buffer width is overly large. Any ideas why this isn't processed as a per preview?
```
    } else if (type == SINGLE_LINE) {
        bufferHt = 1;
        bufferWi = newNodes.size();
        int cnt = 0;
        for (int x = firstNode; x < newNodes.size(); ++x) {
             ... etc ...
        }
```

### DIFF
--- a/xLights/models/Model.cpp
+++ b/xLights/models/Model.cpp
@@ -4225,13 +4225,13 @@ void Model::InitRenderBufferNodes(const std::string& tp, const std::string& came
         logger_base.warn("Model::InitRenderBufferNodes BufferWi was 0 ... overridden to be 1.");
         bufferWi = 1;
     }
-    if (bufferHt > 10000) {
-        logger_base.warn("Model::InitRenderBufferNodes BufferHt was overly large ... overridden to be 10000.");
-        bufferHt = 10000;
+    if (bufferHt > 100000) {
+        logger_base.warn("Model::InitRenderBufferNodes BufferHt was overly large ... overridden to be 100000.");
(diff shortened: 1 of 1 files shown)


xLights/models/Model.cpp
```diff
@@ -4225,13 +4225,13 @@ void Model::InitRenderBufferNodes(const std::string& tp, const std::string& came
...
+    if (bufferHt > 100000) {
+        logger_base.warn("Model::InitRenderBufferNodes BufferHt was overly large ... overridden to be 100000.");
+        bufferHt = 100000;
     }
-    if (bufferWi > 10000) {
-        logger_base.warn("Model::InitRenderBufferNodes BufferWi was overly large ... overridden to be 10000.");
-        bufferWi = 10000;
+    if (bufferWi > 100000) {
+        logger_base.warn("Model::InitRenderBufferNodes BufferWi was overly large ... overridden to be 100000.");
+        bufferWi = 100000;
```
dkulp, line 4234: Kind of wondering if these checks should only be applied if (bufferWi*bufferHi) > 2.1M or so.  Basically, a really long single line is fine, but kept to some level of reasonableness.  (~2.1M would allow a 1080p hd virtual matriix which would be kind of a lot for rendering.).

derwin12: Like a wrap the above in another if -- seems like a good idea. Just trying to stop that crash from wacky values.

PR #4721, opened 2024-08-13
ILightThat controllers: Add new board

PeteMatthews:


computergeek1507:
should it be a dmx port not a pixel port?

PeteMatthews:
> should it be a dmx port not a pixel port?

You're absolutely right - committed from wrong repo clone. Apologies.

The controller plugin doesn't need amending - It already supplies all the config this board needs.

Thanks :)

patdelaney:
@PeteMatthews What are  ILightThat controllers?

computergeek1507:
The Baldrick Controllers buildalightshow.com, Gilbert Engineering, Wired Watts are selling.

Issue #4675, opened 2024-07-27
Turn off Alias Prompting Option


Clyde-L:
Can a new Preference be created to stop the auto-prompt for saving an Alias?  

Thank you for looking,
Clyde L

![image](https://github.com/user-attachments/assets/d399ab64-f9ea-47a0-9e0e-0b6be2d8a4f6)

kylegrymonprez:
A commit has been merged for alleviating this issue. Check it out and see how it works for you @Clyde-L

Clyde-L:
> A commit has been merged for alleviating this issue. Check it out and see how it works for you @Clyde-L

Thank you!

Issue #4657, opened 2024-07-21
Relocate directory configuration for FSEQ, RenderCache and Backup out of RGBEffects


mjunek:
## Problem ## 
xLights saves the location of FSEQ, RenderCache and Backup to the rgbeffects file.

This is a real pain for anyone who switches and changes machines, like I do, as when they are not found, they get defaulted back into the show folder.
This is definitely pronounced when using shared storage, such as a NAS, Cloud, or Git to move machines.

As an example of my my configruation:
* On one machine, the RenderCache directory might be D:\Temp\RenderCache
* On a second it might be /Users/mjunek/rendercache
* On a third it might be /tmp/rendercache
* and the fourth might be C:\xLights\RenderCache

I switch between Windows, Mac and Linux installations of xLights, across a few machines depending on my physical location.

As these directories are not found on the "other" machines when loading xLights (after the sync of show folder), they are automatically recreated in the show folder.

This was a big problem whilst storing the show directory in cloud storage as this caused large amounts of data to be resync'd to the cloud. I've mitigated it by switching to git (The defaults are in .gitignore); however disk space arrangements on some machines is a reason to move it elsewhere.

## Proposed Solution ##
The configuration for the FSEQ, RenderCache and Backup directories are set in the xLights settings dialog. These should be saved with the machine, such Windows Registry, a local configuration file or somewhere machine specific. 

This would then allow me to have different configurations depending on the machine I'm sequencing on, without it being overwritten every time the machine is changed.

keithsw1111:
I dont think this is as simple as this. When you work with multiple show folders on a machine you do not want to be mixing these folders. So then you end up with needing in the registry to distinguish show folders ... so create a subfolder under the folder you might say ... but this assumes all show folders are uniquely named ... which they are not ... unless you include the whole path ... and now you have path length issues.

mjunek:
Thanks Keith - that makes perfect sense when running across multiple show directories. Not something we considered when we discussed it in Melbourne; but obviously a glaring issue :)

A potential alternate proposal then:
Within each show folder have a separate file ```xlights_prefs.xml``` or something along those lines; sitting next to the rgbeffects/networks files which store the UI specific settings. This way it would be per-show-folder; and then the user could choose whether to synchronise the settings across machines or not.

AGFazio:
@mjunek  Use a service like Dropbox. I keep my showfolder there and all my sequences. I can then access from any machine I have xLights on.

[...]

PR #4735, opened 2024-08-20
Issue #4675 - Support the ability to set the alias behavior for renaming models

kylegrymonprez:
For Issue [4675](https://github.com/xLightsSequencer/xLights/issues/4675)

Added the ability to set the behavior for alias setting on renaming models. The default is the current behavior. The user can set "Always Yes" or "Always No" per their preference, or leave it to always prompt.


cybercop23:
@kylegrymonprez  seems like you are mixing PRs here. The original one was for #4657 (I would not have added that, but that's just my 2 cents) and then you are adding to the same PR.. the FPP Connect change. They should be 2 separate PRs.

kylegrymonprez:
Oh shoot - I forgot to switch branches. Let me address and separate the branches. 

My bad....

kylegrymonprez:
Let me close this and reopen a new one.....

Issue #4732, opened 2024-08-19
Moving Head Models (old type) Crash xLights / Upgrade process


davidlsl123546:
It's known, and documented in issue 4712 that older (non MovingHeadAdvanced) moving head models crash xlights on import.

To produce the error:
1) Grab a model like this one: https://www.dropbox.com/scl/fi/a40riy2ksranh9nukgexp/Dominar-Lempa-1-v12-15-2023.xmodel?rlkey=xku2tmchr6ghmjhm3fm1ah7n4&dl=0

2) Import to xLights layout via "import".
3) Error pane appears - "DMXMovingHead3D" is not a valid model type for model "DMXMovingHead3D"

Version: 2024.14, but I believe also any version since the new moving head model appeared.

Crash Log: https://www.dropbox.com/scl/fi/9opbv14yqlr9facx7fmfw/xLights_dbgrpt-29800-20240819T072433.zip?rlkey=blt0bs389np1kfjs9uy45p925&dl=0

Thanks!,
-David

cybercop23:
We will look to fix the crash that's happening trying to import a "non-existing" model. The issue with importing the new MH model has been fixed and will be available in the next version. I suggest you provide both versions as there is no current plan to add support to upgrade from the old model to the new model on import.

PR #4744, opened 2024-08-22
[4732] fix crash on old DMX model

kylegrymonprez:
Fix crash on importing an older DMX MH model. xLights no longer crashes after the error prompt


derwin12:
#4732

Issue #4753, opened 2024-08-24
Corrupted MP4 uncompressed matrix model export - matrix 216/384 (testing shows failed exports dating to 2024.09 release)


mcenerykwm:
Use case:  export xLights effects  from a 192/384 Matrix model as MP4 uncompressed video

Discovered that since 2024.09,  the generated uncompressed video fails to display on local machine (Windows media player) nor via video effect in xLights sequence

Expected behavior in 2023.23 - 2024.08.   Broken behavior 2024.09 - 2024.15

Platform: Windows
Layout - single matrix - 192/384
Sequence - 30 second butterfly effect 

Layout, simple butterfly test sequences and several MP4 non-compressed video export examples

When place video effect in sequence, link to video does not display (blue box).  Of interest. the exported MP4 file is shown by other application including uploading to GITHUB (below) 

Thanks

https://github.com/user-attachments/assets/8b93f573-8ac3-4117-a347-603388b9609c

https://github.com/user-attachments/assets/b9819933-a1c3-43c6-8db8-6aa756eb589c

[ButtetflyTest.zip](https://github.com/user-attachments/files/16738060/ButtetflyTest.zip)

derwin12:
I think this was the move from AVI to mp4.
From Google:
"You can't do it. MP4 specification only accepts H264 or H265 codecs for video (based on the MPEG4 file specifications) inside an MP4 container. If you need an uncompressed file probably you need to create an .AVI or .MOV file"

-- do we just remove the option of uncompressed MP4?

derwin12:
Can you try changing the name of the file to an .avi when you export it (not just renaming the resulting file). There is code that handles an .avi file differently.
I cant duplicate the issue using .mp4 but that can me different codecs installed on our PC's

mcenerykwm:
Daryl,

If can't duplicate the MP4 display issue then agree could be local  codec issue - although primary computer recently purchased. 

Agree IF the uncompressed option is effectively not available  then removing "uncompressed" option should be considered to be removed....

Kevin

derwin12:
Use of uncompressed AVI still appears to work.

computergeek1507:
The "Compressed AVI" is created a MPEG4 in a AVI container. Not really a common format(it does load in VLC) Dan added the "Compressed MP4" which uses H.264 and works with the hardware decoder on the Raspberry Pi. "Compressed AVI" probably is using different settings with newer ffmpeg which maybe causing the issue.

[...]

PR #4760, opened 2024-08-26
Remove uncompressed MP4 as an option - not supported #4753

derwin12:
This still leaves uncompressed AVI and compressed MP4 for exporting of models. It seems things broke when ffmpeg was upgraded and google says uncompressed MP4 really isn't really a thing. #4753


computergeek1507:
https://github.com/xLightsSequencer/xLights/commit/6be9fab6629834583c0fbbf008f2a9a60f20b130

Issue #4749, opened 2024-08-24
State Dialog has UI issues compared to the Face Dialog


derwin12:
When first opened it is too small (and the min is too small).
![image](https://github.com/user-attachments/assets/40b0f431-5f20-4b7a-96db-d150517621f9)

Face Dialog
![image](https://github.com/user-attachments/assets/53fbdda8-75c6-4bfa-b285-160848dd5fb9)

With 100% windows res, the top buttons are too thick.
![image](https://github.com/user-attachments/assets/a60cc370-82c8-47ce-8430-c839f6b9ee28)

PR #4763, opened 2024-08-27
Tweak the State Dialog to avoid squished buttons #4749

derwin12:
Set min size for the slider panel so it doesnt start of squished #4749


computergeek1507:
400, my 640x480 CRT won't like that

derwin12:
I assume you are joking .. what does it look like now .. it must be unusable then.

derwin12:
Min size for these buttons to show poperly
image

computergeek1507:
mine only does this small with the current master code
image

derwin12:
Confirmed .. I will close this out.. not sure why mine was messed up... perhaps I didnt get the latest commits.

PR #4772, opened 2024-08-29
[enh] Making the alias prompting better by adding "Skip New Models"

kylegrymonprez:
The prompt is still useful, but mostly when adding new models it can be annoying. Adding a choice to skip asking only if it was the last generated model.


keithsw1111:
I dont think this should be an option ... it should just be the behaviour.

kylegrymonprez:
Fine by me :) I'll update the PR later today.

kylegrymonprez:
Updated default behavior per comments.

Issue #4629, opened 2024-06-22
Change Pop up from Abandon Changes to Save Changes


Jaa1i:
When editing model data, and you press cancel, a prompt asks "Abandon Changes" with the option of No or Yes.

This is vastly different to standard practice with pretty much all computer applications. When attempting to close something without saving, it will ask "Do you want to save changes?" not "Abandon Changes".

I can see someone inadvertently pressing yes, thinking they are saving changes.

<img width="293" alt="Screenshot 2024-06-22 at 23 03 21" src="https://github.com/xLightsSequencer/xLights/assets/50488459/e02f400f-d748-4b07-b66e-1a8c08d64790">

Jaa1i:
Yes, if you press OK, it saves. However if you press Cancel, you receive the Abandon Changes pop up. This is not the industry norm. In almost all other applications, if you close a document that you have not saved changes, it doesn't ask you if you wish to Abandon Changes, it asks if you want to save your changes.

computergeek1507:
The import and Submodel dialog have the same reverse logic questions on close, I thinks it is because everything is upside down and backwards in Australia

PR #4790, opened 2024-09-06
Make the abandon messages consistent #4629

derwin12:
Consistent wording of the abandon dialogs along with catching the close window on the import dialog. #4629


computergeek1507:
What about the Submodel dialog

derwin12:
This seems okay - or is there another place?
image

PR #4791, opened 2024-09-06
[enh] Custom Model - improvements to Node starts for Individual Strings

kylegrymonprez:
A common troubleshooting issue in the zoom room is when splitting custom models into individual strings, the user often has to update all of the numbers manually.
The intention behind this addition is to make this process easier for the user to implement.

### Current Behavior
When the user checks "Indiv Start Nodes", it will evenly break up the amount of nodes evenly per string. This is pretty uncommon when people are connecting nodes to lights, especially on models with node counts that aren't easily divisible by 50 or 100.

### Changed Behavior

- With this change, nothing is auto calculated by default. When a user checks "Indiv Start Nodes", the strings are auto populated to 0 counts. Then the user enters the starting pixel for the 2nd String. After that happens, the code auto calculated the node count, and propagates that count as far down the string counts as it can.
- Strings with 0 counts are highlighted in red to indicate to the user a setup issue.
- An error dialog was added if the user attempts to add string counts out of order.
- As before, if a user unchecks the box, and then rechecks the box, it will reset the starting node count per string. In this case, it resets all the node start counts to 0 instead of evenly dividing it again, allowing the user to start over if they make a mistake.

Screenshot 2024-09-06 at 10 08 53 AM
Screenshot 2024-09-06 at 10 09 36 AM
Screenshot 2024-09-06 at 10 09 46 AM
Screenshot 2024-09-06 at 10 09 07 AM

### DIFF
--- a/xLights/models/CustomModel.cpp
+++ b/xLights/models/CustomModel.cpp
@@ -143,13 +143,15 @@ void CustomModel::AddTypeProperties(wxPropertyGridInterface* grid, OutputManager
                     ModelXml->AddAttribute(nm, val);
                 }
                 int v = wxAtoi(val);
-                if (v < 1) v = 1;
                 if (v > NodesPerString()) v = NodesPerString();
                 if (x == 0) {
                     psn->SetValue(v);
                 }
                 else {
-                    grid->AppendIn(p, new wxUIntProperty(nm, nm, v));
+                    auto localProp = new wxUIntProperty(nm, nm, v);
+                    grid->AppendIn(p, localProp);
+                    if( v == 0)
+                        grid->SetPropertyBackgroundColour(localProp, *wxRED);
                 }
             }
         }
@@ -215,7 +217,7 @@ int CustomModel::OnPropertyGridChange(wxPropertyGridInterface* grid, wxPropertyG
         if (hasIndiv) {
             for (int x = 0; x < _strings; x++) {
                 wxString nm = StartNodeAttrName(x);
-                ModelXml->AddAttribute(nm, ComputeStringStartNode(x));
+                ModelXml->AddAttribute(nm, x == 0 ? "1" : "0");
             }
         }
         IncrementChangeCount();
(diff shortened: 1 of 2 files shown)


keithsw1111:
I dont see how this is a better behaviour. It is less convenient and makes it less clear to the new user what then need to do.

cybercop23:
say they have strings of 150 or something that's not 100... and we see users having a hard time adding the set number to each string. having something like this to help and pre-fill would be a huge help. how do you suggest to make it better?

keithsw1111:
Default to multiples of 100 would likely be better but defaulting to zeroes is a bad idea.

cybercop23:
We also talked about when checking the invChan there will be another entry that will give "steps".. similar to what you are saying.. then they can enter that value and they will be updated. Make sense? Will that work? Or you feel regardless of what we do, we need to have some "default" values?

keithsw1111:
We have to have default values ... i would take the number of strings and work out the nearest multiple of 100 that makes it work. The only challenge then is deciding if the last string is > or < than the others and that should be based on if the rounding is < or > 50.

kylegrymonprez:
@keithsw1111 so something like this could work:
- round(totalNodes/#strings, <nearest 50>)
- take that result and multiply by strings -1
- leftover nodes go on the last string

Example:
Rosa Grande is 1392 nodes.
User puts 6 strings.
round(1392/6, 50) = round (232, 50) = 250
250*5 = 1250. Strings 1-5 are 250 each
String 6 is the leftovers: 142 nodes

Then user can update individual strings as needed, or the whole thing (like now).

kylegrymonprez:
The alternative example is the last string gets more nodes:

Model X is 1300 nodes
User puts 6 strings
round( 216, 50) = 200
Strings 1->5 = 100 nodes. 300 leftover
String 6 gets 300, and user makes a choice to keep or redo (more strings or update numbers)

keithsw1111:
I would not base it on strings of 50. 100s are so much more common for those using 12V and most people use 12V.

computergeek1507:
I personally only have 50s I'd prefer rounding to 50

kylegrymonprez:
(/me stops the urge to put string size preference selector)

kylegrymonprez:
i toyed with the idea of a "..." button after the checkbox that would pop up a dialog box that could be used to enter a string size calculator/wizard

cybercop23:
> I personally only have 50s I'd prefer rounding to 50

that's the only thing that makes sense. port pack too.

keithsw1111:
If we are doing it based on what we do then i would leave it exactly as is. If we want to cater to the most common situation then 100 makes most sense.

PR #4819, opened 2024-09-20
Fix Fade on 'Head and Tail' fade type for certain layer blendings

MrPierreB:
There is currently a bug on the Head and Tail fade type on Single Strand when you use another layer blending besides normal.
Here you can see I selected 1 color only from the orange palette, I set the fade type to Head and Tail and then I set the Layer blending to Effect 1 (you can use many others like additive etc)
BeforeFix

Here is the same after the fix.
AfterFix

### DIFF
--- a/xLights/effects/SingleStrandEffect.cpp
+++ b/xLights/effects/SingleStrandEffect.cpp
@@ -722,19 +722,25 @@ void SingleStrandEffect::draw_chase(RenderBuffer& buffer,
                             } else {
                                 color.alpha = 255.0 * ((2 * i +1 - max_chase_width)) / max_chase_width;
                             }
-                            
                         }
+
                     } else {   // head tail not alpha 
                         HSVValue hsv1 = color.asHSV();
-                        if (i > middle_chase_index) {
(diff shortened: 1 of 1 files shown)


xLights/effects/SingleStrandEffect.cpp
```diff
@@ -714,33 +714,33 @@ void SingleStrandEffect::draw_chase(RenderBuffer& buffer,
...
-                                    color.alpha = 255.0;
+                                color.alpha = 255.0;
                             } else {
                                 color.alpha = 255.0 * ((2 * i +1 - max_chase_width)) / max_chase_width;
                             }
-                            
-                        }
-                        if (i == i ) {
-                            color.alpha = color.alpha;
                         }
```
MrPierreB, line 725: not sure why this was here, seems redundant.

MrPierreB:
@dkulp or @computergeek1507   Any reason why this didn't make it into master.  Has been up for 4 weeks and didn't have merge conflicts until 2 weeks ago after 900accf1c73be08bd3fcc98c2e67a1eb919a19c9

Issue #4704, opened 2024-08-06
Enhancement: Better keybindings for media control


DoctorWiz:
**Is your feature request related to a problem? Please describe.**
I have some [USB foot pedals for transcription](https://a.co/d/gKbOAfP).  When used with transcription or editing software, pressing the big middle pedal with your foot plays, and letting it up pauses.  The smaller pedals on the right and left can be configured to fast forward or rewind, or jump fwd/back to a previous marked spot, or change speed.  I use these in Audacity and in DaVince Resolve and they make work soooo much easier!  I think they would be even more useful in xLights when you are constantly starting and stopping and replaying little segments over and over.

I will send a **FREE!** set of pedals, with shipping (U.S.) to the dev who will add the needed key bindings.

**Describe the solution you'd like**
Keybindings to:

1. Play from where-ever it was last paused or stopped.  But if already playing, is ignored.
2. Pause if playing, ignore if already paused/stopped.
3. Jump back to previous mark/chapter.  If already at a mark, jump to the prior one.  (See Keith’s answer to issue 2639)
4. Jump to the next mark/chapter.  If already at a mark, jump to the following one.
5. Decrease the playback speed (as opposed to setting to a particular speed) (Like the **Audio_S_1_4_speed**, **S_1_2**, etc. but goes in steps)
6. Increase the playback speed.
7. Fast-Forward  (**Audio_F_3_Speed** or **F_4** already works good for this)
8. Rewind _with audio_ so you can tell where you're at (optional, this might be tricky to implement).
9. Resume previous playback from fast-forward or rewind.  (**Audio_Full_Speed** works for this, but would be better if it could go back to previously set speed which might be ½, ¼...)

#1 and #2 would be assigned to the big middle pedal, down and release.  #3 and #4 would be assigned to the small left/right pedals in one mode, #5 & #6 in another mode, and #7 & #8 to the left and right pedal down in a 3rd mode with #9 to left/right pedal up.

**Describe alternatives you've considered**
Also note that the **PLAY** keybinding starts from where-ever last clicked in the waveform, not where it left off, so that does not apply well to the foot pedals.
The **PAUSE** keybinding won’t restart the audio after doing anything that causes the waveform portion of the window to lose focus, which is almost anything.
[...]

PR #4840, opened 2024-09-28
Keybindings to increase and decrease speed along with next and prior

derwin12:
#4704 key bindings to increase / decrease speeds rather then just the current set speeds.
Also key bindings to jump to prior or next tags while doing the playback.


keithsw1111:
You should remove the testing debug statements

PR #4841, opened 2024-09-28
First conTrollers from GalloTech

Gallosauro:
New controller from GalloTech

Getting started with new controllers compatible with WLED software with ESP32, ESP32-C3, ESP8266 chip.
 
Very open to comments/suggestions/rework as required. Please do let me know if you'd prefer anything revised.

Thanks,
GalloTech


Gallosauro:
Please disregard the Workflow.

PR #4842, opened 2024-09-28
New Controllers from GalloTech

Gallosauro:
Added new Controllers from GalloTech.
WLED compatible with ESP32, ESP32-C3, and Esp8266 chip.
First Pull request is very open to comments, and suggestions, of any kind so please let me know if any revision is required. 

Thanks,
GalloTech

### DIFF
--- a/.github/workflows/Ubuntu_Window_Release.yml
+++ b/.github/workflows/Ubuntu_Window_Release.yml
@@ -7,30 +7,13 @@ on:
     
   workflow_dispatch:
 jobs:
-  build-linux:
-
-    runs-on: ubuntu-22.04
-    steps:
-      - uses: actions/checkout@v2
-        with:
-          submodules: 'true'
-      - run: docker pull ghcr.io/xlightssequencer/xlights-build-docker:master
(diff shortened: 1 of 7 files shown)


MrPaulAR:
Why does the WLED controller not work for this if it's "WLED compatible"?

derwin12 (commented):
How is this different than the current WLED controller?
Is there a website for this controller?

cybercop23:
At best, this should be a Model under the WLED Vendor and not a vendor. You copied all of the WLED code and just changed the name. NO. There's also no difference in pinouts so it can/should be just streight WLED Vendor & Model.
image

computergeek1507:
Please provide information about your controller and the company behind it. We do not add controllers from unknown brands until there is some of a user base in our community. If it is a WLED based controller, no new upload class is needed, it should use the existing WLED code I wrote.

Gallosauro:
Thanks for the answers, there is not company behind yet, I made 3 controllers that are in the testing phase. 
the controllers works with the WLED software, one is a micro with only one port, one with 6 ports and one with 8 port and LCD. the pinout is slightly different than the one pointed out from Cybercop23. I did not know I could add them under WLED, my bad, it is the first time I add something in github.

PR #4845, opened 2024-09-29
Add support for default port gamma with Genius controllers

thekev:
Pretty simple change. Was implemented in FPP-based controllers, but found it missing in Experience controllers. Tested on MacOS 15.0 arm64 with Genius Pro 16.


computergeek1507:
you need to add the same code at line 348 to use default gamma for the receiver ports

thekev:
> you need to add the same code at line 348 to use default gamma for the receiver ports

thanks for the catch. done.

Issue #4839, opened 2024-09-27
When creating tags - a wxWidget error floods the debug logs


derwin12:
Create a music sequence, in the top of the timeline, right click and add a tag (0-9). Get this failed ExtFloodFill error.
(Current and prior versions I assume)
Seen in Windows.
![image](https://github.com/user-attachments/assets/2aa2a6a1-c152-418c-9489-0c4974e10b5b)

PR #4846, opened 2024-09-29
Bug fix to fill the tag correctly #4839

derwin12:
Remove the use of the floodfill that was throwing an error and not working. Now using a polygon #4839
image
(please verify on mac)


dkulp:
Not readable in dark mode.

derwin12:
> Not readable in dark mode.

Added code to correct that.

Issue #4829, opened 2024-09-24
Add another face option - Wings


Littleaton:
**Is your feature request related to a problem? Please describe.**
I have messed with states and so on for bat wings or penguin wings. but I think adding it as a face option might be easier. 
there are many props out there that have wings or other "body " parts that move, this could be a Beter solution to make movement out of them that is auto like eyes. Id love to have bat wings or other flying things randomly flap through sequences like eyes do.

**Describe the solution you'd like**
Id like to see a face option added for wings flapping
Wings 1
Wings 2

**Describe alternatives you've considered**
I have used this as a state for bat wings, but using it as a face option might be easier. 

**Additional context**

computergeek1507:
Use the eyes open/close 2 or 3 option

Littleaton:
> Use the eyes open/close 2 or 3 option

This is a great idea, but what if i don't want the wings to flap when eyes blink? like be more random not both at once?

SmAsH-jl:
It would be a good addition for props with arms or wings separate from the eyes , bonus if you can set a different cadence for the arms vs the eyes, the props could look more organic, I would use it for my penguins and bats for sure, then the wings can flap faster than the eyes blinking. I feel this would make a ton of different prop options be more dynamic, and open to some creative ideas with other props and animations that could utilize the randomness easily without having to program effects at the strand levels. Bubbles in a cauldron also come to mind,

Clyde-L:
How about adding an option to States, that allows for the "auto advance" to create a movement between the listed states... Say you have "wings" and the wings state has 2 positions, one and two.  Then on the States effect, you would add a random random blink of the states listed (one and two).  This adds the "random blink" to the states.  But also makes it not necessary to attach to a specific timing grid.

kylegrymonprez:
Talked about this a bit with @derwin12 - we think the best case here is to use a timing track, but augment some of the "metronome with Tags" functionality to do this. Coding it up is way more trivial, and there are so many more use cases with the timing track that could be interesting.

I'm going to take a look at this addition to the timing track generator and then you can attach that timing track to the states effect to accomplish this goal.

derwin12:
Kyle's change is in. So using states is more flexible then adding more elements to the face effect.

PR #4858, opened 2024-10-02
[Enh] Randomization options for new "Metronome w/Tags" Dialog

kylegrymonprez:
Something requested that seems powerful that I implemented:

- Added the ability to add randomized timing intervals between min and max. This will create random intervals in the new timing track.
- Added the ability to randomize the labels between 1 and X. This is useful for various state effects to give the appearance of randomization.

This gives various value curves implications on various effects as well.
Screenshot 2024-10-02 at 1 21 02 PM


kylegrymonprez:
Related to #4829 as a solution to solve that request.

PR #4881, opened 2024-10-11
Add automap of submodels and models to submodels

derwin12:


derwin12:
Going to tweak this a bit more ..

PR #4938, opened 2024-11-05
Hightlight To/From models to only map that subset

cybercop23:
Added the ability to select models in own layout and/or available effects to only map a subset of models->effects.


computergeek1507:
Huh?

cybercop23:
> Huh?

you know you like it... Say you added a Showstopper Spinner and want to map that. The source sequence has effects on 17 submodels... you can't use auto map, since it will map anything else in your display, that you may not want to. if you multi-seelct all effects on the right for the SSS and then (optionally) select your SSS model, only those two will be AutoMapped.
In essence I check if anything is selected in either list and when mapping, if there are selections in effect, only those selected would be "available" to match.

dkulp:
I specifically did not merge this as there is  (IMO) a big problem.  Once you click on anything, then the auto mapping no longer works.

PR #4969, opened 2024-11-18
Hightlight To/From models to only map that subset

cybercop23:
Added the ability to select models in own layout and/or available effects to only map a subset of models->effects.

Redo on top of Daryl's change for sub-model map.


computergeek1507:
Why not add a right click instead of adding more buttons? Button overload is not ideal

cybercop23:
That can be an option but too much gets hidden behind right clicks.

computergeek1507:
I don't think from an economics and mouse movement a button makes sense, the user is suppose to highlight a row then move their mouse to the bottom to click the button, then move back up to click the next model and repeat. If it's a right click they wouldn't have to move the mouse off the list.

cybercop23:
Ohhh.. you mean on the list itself... I see... right not they highlight on left one or more models, highlight on right if they only want a subset and then click the button.. you're saying to right click on the list itself/multiselect...
 hmmmmm

PR #4982, opened 2024-11-21
Re-resolve host names on ctrl click or Test if ctrl wasn't on at boot

cybercop23:
Currently if using hostnames for the ctrls and if they weren't on when xlights started, the ctrl status LedPing won't update and you also get an error when doing Tools -> Test since it can't ping the ctrls. We'll now force a re-resolve if _resolvedIp has characters ie: was never actually resolved.
There may be a bit more to do for FPPConnect, but in my testing that also works now.

### DIFF
--- a/xLights/utils/string_utils.h
+++ b/xLights/utils/string_utils.h
@@ -93,10 +93,13 @@ extern const wxString xlEMPTY_WXSTRING;
         return in.find(contains) != std::wstring::npos;
     }
 
-    inline bool hasAlpha(const std::string& str) noexcept 
-    {
-        std::regex pattern("[a-zA-Z]");
-        return std::regex_search(str, pattern);
+    inline bool hasAlpha(const std::string& s) {
+        for (auto c : s) {
+            if (std::isalpha(c)) {
+                return true;
+            }
+        }
+        return false;
     }
 
     inline std::string StripAllBut(const std::string& in, const std::string& but)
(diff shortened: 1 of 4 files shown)


xLights/utils/string_utils.h
```diff
@@ -92,6 +93,12 @@ extern const wxString xlEMPTY_WXSTRING;
         return in.find(contains) != std::wstring::npos;
     }
 
+    inline bool hasAlpha(const std::string& str) noexcept 
+    {
+        std::regex pattern("[a-zA-Z]");
+        return std::regex_search(str, pattern);
+    }
+
```
dkulp, line 101: Compiling a regex and applying a search for this case is way overkill.    A simple loop would likely perform better:
`
inline bool hasAlpha(const std::string &s) {
    for (auto c : s) {
        if ((c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z')) {
            return true;
        }
    }
    return false;
}
`

computergeek1507: (https://en.cppreference.com/w/cpp/string/byte/isalpha

cybercop23: ha... that was my original implementation, but figured it was ugly....

cybercop23: and that too... stsd::isalpha requires the for loop & if...

i can combine both ideas into one..  EVERYONE on board?

PR #4986, opened 2024-11-22
Moved SM remap and skip if no alias

cybercop23:
Moved SM map out of main loop (and back tp its original position) and added logic to skip map functionc all if no alias exists.
Original PR was totally broken, mapping took like 22 min.. with the move was down to 3.22 and adding the skip is back down to 27 seconds (as it should be).


blun23:
Has this been pushed over to TestFlight so I can try it out.  I can't get anything going as it's taking literally forever

Issue #4921, opened 2024-10-24
Ability to mark Ports as do not use in the Visualizer


derwin12:
Just a UI enhancement to mark ports as "Do Not Use" - perhaps a color or hashing. For those of us with controllers where you have ports you don't want to place pixels on. Right click on the port, mark as Do Not Use ?

computergeek1507:
Why

cybercop23:
Yes please. I ahve several DR ports that were blownup on a few ctrl when my yard was struck by lightning and it will be  nice to have a way to clock them.
Even today, on zoom we helped a guy that had some issues with a port in the F48 that he couldn't use. So it will be nice to be able to block them.

derwin12:
Personally I have a bad jack on my F48 and some of other controllers I dont have all the ports wired - ie no pigtails on them to use. Or I have used them for power injection.

cbennett6008:
What about something like color tags, and let the user determine what their colors mean? maybe make it more one code fits all?

cybercop23:
I have a working version of this using the Description field (that's where i keep my info now) ie: Bad Ports:16,17,18
There's a ContainsBetweenCommas function in string_utils but it sort of lies (maybe) since it doesn't catch first and last unless they are trully encapsulated in commas. 

@dkulp @computergeek1507 since the function only used in one other place, any issues with making it handle leading and trailing values? Or make another? Heck.. I can manually add them when in the  function call, but it seems silly.

PR #4988, opened 2024-11-22
#4921 added Bad Port markers

cybercop23:
Using the controller description fileld.. when adding the <tag> Bad Ports: 5,7,9 it will mark those ports in the visualizer as "bad" red


computergeek1507:
Think this is a bad idea

cybercop23:
I know several other in the same boat.... look at mine... I think this will be very handy to have...
image

PR #5026, opened 2024-12-05
Reduced render time for bars by 60-80%, by caching IsSpatial

degracode:
In my profiling, `GetSpatialColor` represents 33% of the time spent inside `BarsEffect::Render`, just checking `buffer.palette.IsSpatial(colorIdx)` for every pixel. If you're not using spatial colors, this time is pure wastage. The optimiser can't see deep enough to know that it can cache `IsSpatial`, so I've done it manually. Despite expecting a saving of only 33%, it's actually more like 60%, because skipping the `GetSpatialColor` function call also skips all of the maths done to build its parameters, such as `(float)x / (float)buffer.BufferWi`, and `(float)(n % barHt) / (float)barHt`.

5.878s -> 2.184s for my test sequence, that is one bars effect with 3 colours, across all channels.

More savings could undoubtedly be had by performing this same optimisation on other effects.

### DIFF
--- a/xLights/effects/BarsEffect.cpp
+++ b/xLights/effects/BarsEffect.cpp
@@ -214,11 +214,15 @@ void BarsEffect::Render(Effect* effect, const SettingsMap& SettingsMap, RenderBu
                 }
                 color = hsv;
             }
+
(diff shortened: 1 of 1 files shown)


degracode:
Build failure is unrelated to PR; looks like 602444ad91696205e813573e594f09b4df6b73a8 is the cause.

PR #5079, opened 2024-12-27
Write Effects to Temp for NAS

cybercop23:
When rgbeffects is on a NAS and we save a new version, because of the line by line save, it takes a very long time to save. Saving it to the local OS temp folder and then renaming it to the show folder, seems to address this slowwwwwnesssss

### DIFF
--- a/xLights/TabSequence.cpp
+++ b/xLights/TabSequence.cpp
@@ -706,7 +706,7 @@ bool xLightsFrame::SaveEffectsFile(bool backup)
     viewpoint_mgr.Save(&EffectsXml);
 
     wxFileName effectsFile;
-    effectsFile.AssignDir(CurrentDir);
+    effectsFile.AssignDir(wxFileName::GetTempDir());
     if (backup) {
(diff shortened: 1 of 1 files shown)


tgrinnall:
this fixed the 4 min save time for my nas now it only take 2 sec. thanks

computergeek1507:
Not a huge fan of slowing down saving for 99% of users to help 1%

cybercop23:
For local disk savings is <1 sec.

computergeek1507:
With 2000+ models?

cybercop23:
Than you are the other 3% with a commercial show.

computergeek1507:
I don't know the wx xml library very well but I assume it could be generated in memory and nod then saved to disk

cybercop23:
I was thinking the same, but since it was never used, figured it didn't exist. I know we talked about several perofrmance enhancements with the move to QT... maybe start now

dkulp:
Updated the save routine to use a wxBufferedOutputStream which keeps it from writing temp files and should fix this.

PR #5080, opened 2024-12-27
same enh for saving xsq

cybercop23:
Same Nas change for saving a sequence


dkulp:
Updated the save routine to use a wxBufferedOutputStream which keeps it from writing temp files and should fix this.

PR #5092, opened 2025-01-02
Split HostName and IP plus others

cybercop23:
1. Split Hostname and IP
2. Make inital sort by IP and make it IP range aware so .100 will be after .99 and not between .1 and .2
3. Ability to sort on either Hostname or IP column
4. Last column sorted will be saved and next time used as the initial sort
5. Added Deselect All for Media
6. Only show Playlist for FPPs that are in playermode (clean up interface)
7. Added some more help hints in the column help pop-ups
8. Initial work for better match of FPPs - created struct with ip, hostname and uuid to be used later for deduping the list

### DIFF
--- a/xLights/controllers/FPP.cpp
+++ b/xLights/controllers/FPP.cpp
@@ -89,6 +89,18 @@ static std::set<std::string> FPP_VIDEO_EXT = {
     "mpg", "MPG", "mpeg", "MPEG"
 };
 
+struct FPPDInfo {
+    std::string hostname;
+    std::string ip;
+    std::string uuid;
+
+    bool operator<(const FPPDInfo& other) const {
+        // Custom comparison operator for uniqueness and sorting
+        return std::tie(hostname, ip, uuid) <
+               std::tie(other.hostname, other.ip, other.uuid);
+    }
+};
+std::set<FPPDInfo> fppDiscInfo;
 
 FPP::FPP(const std::string& ad) :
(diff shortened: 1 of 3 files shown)


computergeek1507:
Added .local to end of a hostname doesn't work on all networks, Dan told me it can cause issues

cybercop23:
> Added .local to end of a hostname doesn't work on all networks, Dan told me it can cause issues

Right.. so it used to be just one link, now there are two. So if hostname.local doesn't work, you can just use the IP one; now that they are separated.

dkulp:
I'm 100% against adding .local to the hostnames.   Don't do that.

cybercop23:
Can we have an option/preference or something else?  How else can we do it?
Lots of folks configure and use hostname.local as the ctrl IP and if it works for them great. How can we allow the same in the FPP connect window? Why do we have to keep it "stuck" on IP? There has to be a way to open up hostname also vs just IP

cybercop23:
What if we first check the controller definitions and if they have any defined with hostname and .local then we know it works for them and we can make the link use that, else if they are all IPs, then just don't make the hostname link-able?

dkulp:
If there network is properly configure, just "http://fpp" works.    If there network is not, then they'll need to use IP.   I do not use ".local" when going to ANY of my FPP instances.

dkulp:
Fixed up and merged...

Issue #5039, opened 2024-12-09
Some suggestions around FPP Connect


kenmcc:
Using FPP connect to upload several sequences, have come across the following niggles which might be nice to consider (am open to contributing, but always seems to crash on linux on me, I don't have a windows dev pc)

- Select/deselect all option in the list of sequences, save having to unclick previously uploaded sequences or click on all sequences.
- A way to upload sequence(s) to more than one playlist - i.e. i have one which is "lights only", which has no sound for antisocial times, and that runs as default, not in remotefalcon. But some of the sequences I like to have in remote falcon playlists also, so it would be nice to have a way to load to multiple playlists when a sequence is updated or added. Possibly an "Upload and retain dialog" button, or a tickbox "remain open after upload" or similar.
- More granularity on the folders shown in the FPP connect window - from memory it only shows the 'show folder' and can recurse into subfolders but it would be nice to be able to drill down to a specific folder so as not to have lots of sequences (perhaps old/testing ones) in the list and accidentally have them uploaded with an errant mouse click.

Thanks

kenmcc:
Thanks - never thought to try right clicking :(

cybercop23:
when in doubt, right click.. even areas you won't think they'll have any info.. like the colum headers on the top part.. you'll be surprized in what you'll find hidden there.

cybercop23:
Will be available in the next version
![image](https://github.com/user-attachments/assets/99d94018-2510-4d7d-8b60-fab4af429232)

PR #5094, opened 2025-01-03
FPP Rediscover and KeepOpen (plus a few more FPPConnect enhancements)

cybercop23:
Few more FPPConnect changes:

- #5039 Keep window open to upload to multiple playlists
- Rediscover button if you forgot to turn on controllers
- Additional sort on Upload column, not saved, but will be based on previously sorted order (for PatD)
- Right click Upload Outputs on controller for faster/easier uploads when you screw things up
- New/never seen controllers will not be checked for Upload or have Media checked - let the user decide
- Made Hostname case insensitive for those with less OCD that mix their ctrl names (Pat D and Ron A)
- Use UUID as the key for saved settings to reduce the false lookup when DHCP ctrls change IP - if no uuid, falls back to IP
- Reporpose BatchRenderSelect to be more generic and added logic to save and select failed sequences
- Added extra logging for FPPConnect sequence uploading to each instance - include bytes sent


computergeek1507:
Unsure why you would need a right click upload and upload button

cybercop23:
> Unsure why you would need a right click upload and upload button

too much mouse travel in large screens when you need to do one or two ctrl cuz you scrwed up

Issue #5096, opened 2025-01-05
Make "BatchRenderItemList" comma safe


derwin12:
If you batch render a sequence that contains commas, subsequent batch renders wont remember this value.
This is also an issue with the new fpp connect remembering the sequence.
config->Read("BatchRenderItemList", & itcsv, "");

Perhaps just clean the suggested file name that is currently based on the audio mp3.
Or perhaps it needs a better separator choice?

![image](https://github.com/user-attachments/assets/5ac33d01-bf0e-49e7-a190-de029550a5e6)

PR #5097, opened 2025-01-06
#5096 change batchrender separator

cybercop23:
Make Daryl happier

### DIFF
--- a/xLights/BatchRenderDialog.cpp
+++ b/xLights/BatchRenderDialog.cpp
@@ -269,7 +269,7 @@ bool BatchRenderDialog::Prepare(const wxString &showDir)
         config->Read("BatchRendererItemList", &itcsv, "");
 
         if (!itcsv.IsEmpty()) {
-            wxArrayString savedUploadItems = wxSplit(itcsv, ',');
+            wxArrayString savedUploadItems = wxSplit(itcsv, '|');
 
             wxTreeListItem item = CheckListBox_Sequences->GetFirstItem();
             while (item.IsOk()) {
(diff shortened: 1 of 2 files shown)


dkulp:
Can I ask "why?"     What's the point of this other than it will break everyone that currently uses batch render extensively and have their current lists setup the way they expect/want them to be?

computergeek1507:
A comma is valid in a file name so comma separated list breaks if a file has a comma in it.

derwin12:
and it is an internal temp list so it should not break anything..

computergeek1507:
It's in the settings so they are not temporary

cybercop23:
Right, its in the registry, but I forgot to check/adjust how it is written. I'll look at that

dkulp:
I should point out that | is a valid filename character as well, at least on real OS's

derwin12:
Scott - correct me here .. one does a batch render, it builds this list (stores in registry) .. then in fpp connect one can optionally right click and pull that list to pre-select the files in the fpp connect.
So yes there is a risk that if they dont do a NEW batch render in this version, it will have an old list. Ie .. it wont pull anything. Much like it would be if you batch rendered prior to this feature being added.
The issue is .. if your sequence has a , in it .. it wont read the list, or it loads up till it hits this "bad" file name. 
As for the | character .. yikes if files have that in there. Isnt that the output pipe in linux AND mac?

cybercop23:
Both saved and retreive are using |. I think that's safe as a delimiter, but happy to change to soemthing else, if there's stong feeling.

Issue #5087, opened 2024-12-31
Import from zip needs to remove the commas from the ImportMedia directory.


derwin12:
When importing from zip or xsqz it will try to use the song track to create the importmedia folder. This means it creates a folder with potentially harmful characters. For sure it created one for me with a comma in it. It looks like it is working but upon reopening the sequence it complains about missing assets.

![image](https://github.com/user-attachments/assets/bab3cf72-7143-48ed-bf83-b2d84d2ef8f9)

```
Checking problems with file paths containing repeated use of show folder name.
    WARN: Unable to check file E:\2023\ShowFolder3D\ImportedMedia\David Guetta because it was not found. If this location is on another computer please run check sequence there to check this condition properly.
```

`ERR: Shader effect cant find file 'E:\2023\ShowFolder3D\ImportedMedia\David Guetta'. Model '01 Everything', Start 1:50.150`

PR #5100, opened 2025-01-07
Dont create media dir that is invalid in xsq #5087

derwin12:
Since an Import, import media by default is based on sequence name it allows it to create a directory that contains commas. Commas in the path of shader, images etc will break in the xsq file. #5087


computergeek1507:
There is a clean name until function, other characters should be removed too

derwin12:
I looked for it. Can you point it out to me?

derwin12:
Note that if it is in the file name it is valid to use 'cept for the import with the comma as it breaks the effect. So I shouldnt have to recheck for invalid characters like pipe for example since it cant be in the file name I am importing from.

cybercop23:
AKA... Create SAFE Media Dir...

Issue #5018, opened 2024-12-02
ESpixelstick controller settings missing in layout and with upload not overwriting disabled/not used channels


dutch-anykey:
**Describe the bug**

1 - Xlights is not handling the ESPixelstick correctly in the layout tab. I'm missing option in the controller settings section like color order and brightness

2 -  controller upload is not updating/deleting unused ports correctly

ESpixelstck v4.X (tried different ESPixelstick models and all react the same)

Auto discovery works and Xlights finds the ESpixelstick:

![3](https://github.com/user-attachments/assets/e432ea25-7c09-43a6-88c5-96ebeb1a7913)

Created 3 objects in Layout

![2](https://github.com/user-attachments/assets/b409bc18-97d0-43f9-9125-1e4f65980f3e)

assigned objects to the controller:

![4](https://github.com/user-attachments/assets/3f5833c8-fe9e-4ea8-bef2-582ddf7290ba)

Now when looking in the layout I'm missing the options like "color order" and "Brightness"
Only port, protocol, set end null pixels and set Zig Zag are there.

![5](https://github.com/user-attachments/assets/58e38653-310e-4db9-942b-8a91aa3a12a6)

Second part of the problem:

Uploading to the controller. This works and settings are visible in the controller settings:

![4](https://github.com/user-attachments/assets/5efc0cd2-db04-4f3e-8dc0-1ec285f309e1)

![7](https://github.com/user-attachments/assets/248c9f67-120f-468b-923d-d8c6d47f49ca)

But when I change some settings in Xlights i.e. moving objects to other ports and/or make one port empty, the update to controller only updates changed or new port settings, but port that has become unused in Xlights is not overwritten with empty value.

![8](https://github.com/user-attachments/assets/dd9661bc-e2db-4205-8a04-f0497ef79dd8)

So it looks messed up like below:
![9](https://github.com/user-attachments/assets/3c0eea7b-2b34-4e21-bd63-2bebc10ff88c)

**Expected behavior**

1)  Would like to see all details in controller connection in the layout tab (specifically brightness and  color order is what I'm missing now)

2) Upload mechanism to the controller should correctly update outputs that are not used anymore.

**Versions (please complete the following information):**
 - OS:  Windows 11 pro
 - xLights version 2024.19

kind regards,
Maurice

cybercop23:
Done. Will be in the next version or try the nightly.

dutch-anykey:
Thanks !!!

Looking good now :)

PR #5105, opened 2025-01-09
#5018 Add/enh ESPixelStick v4 support

cybercop23:
Added Default and Port Brightness and Gama
Added PreNull pixels and ZigZag
Disable unused ports since it used to keep all previously configured ports active in perpetuity


cybercop23:
@MartinMueller2003, please take note.. Nothing really for you to do, but just be aware of the changes.

Next we need to talk about the upload process. Killing me...

Issue #5116, opened 2025-01-15
Add an option to remove all aliases


derwin12:
Seems we need some new feature..  tools->delete all aliases ..

AGFazio:
YES please !!

cybercop23:
NotePad++
![image](https://github.com/user-attachments/assets/75b874e6-9d96-4048-be94-c9a1d2409fcd)

AGFazio:
But I want a button to do it.....

cybercop23:
It is... Replace All :)

blun23:
> It is... Replace All :)

You really think it's smart for someone like me to be messing with the code?

cybercop23:
Yup.

cybercop23:
What about a Lua script to remove all?

computergeek1507:
Burry it in a right click menu

derwin12:
Here perhaps if not tools...
![image](https://github.com/user-attachments/assets/5c2def55-8838-452d-a26a-e55e46460be0)

cybercop23:
Availbe in next version both via Lua script and Right click options.

blun23:
> Availbe in next version both via Lua script and Right click options.

Awesome

PR #5125, opened 2025-01-16
DeleteAllAliases = Lua script + supporting functions

cybercop23:
Plus some Lua helper function/changes


cybercop23:
#5116

computergeek1507:
Seems like a right click would be easier

cybercop23:
Now available in Right click menu - for Scott

PR #5142, opened 2025-01-22
Add FPP support for xLights Only state

cybercop23:
One of the bigest issues we see on Zoom, people not knowing how to set up FPP absed ctrl to xLights only.

### DIFF
--- a/controllers/fpp.xcontroller
+++ b/controllers/fpp.xcontroller
@@ -44,6 +44,7 @@
         <SupportsDefaultBrightness/>
         <SupportsDefaultGamma/>
         <PreferredInputProtocol>DDP</PreferredInputProtocol>
+		<PreferredState>xLights Only</PreferredState>
         <ConfigDriver>FPP</ConfigDriver>
 		<FPS40Pixels>800</FPS40Pixels>
 		<FPS40Pixels_SR>720</FPS40Pixels_SR>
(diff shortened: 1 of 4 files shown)


computergeek1507:
This is probably not a good assumption, many people don't use master remote mode and put their FPP controllers in bridge mode

cybercop23:
I really wish we can talk reality vs feel. We're sitting on Zoom hours on end and helping folks with freezing and double data because they have them set as Active. This last season, I only rememebre ONE guy that had Kulps as bridge.
Again, we need to do this and support majority and based on what we see day in/out on zoom 99.9% of users, need to be xLights Only and a large number have them set wrong. We've had people that ran like this for 2+ years not knowing there was an issue.

computergeek1507:
It is naive to think what you see in the zoom room is "all xLights users" if you look at the FPP statistics page you will see approx 9000 device with multisync disabled and 10000 with it enabled. Other user statistics done by some vendors and mini/conferences show approximately 20-30 percent of Users use xSchedule and 70-80% FPP. Lots of xSchedule and FPP user don't use master remote. Making changes because "we see a lot in zoom" fails to recognize the bias that people only go to zoom if they need help.

cybercop23:
That is fair and I would agree with it. If folks know what they are doing, then no isssue. If we're trying to make it easier for new folks to get started, then we need to make things like this easier/default. Experience users would a) not be adding new controllers or even running Discover and b) would know to change it appropriately.
Maybe we can start having some of those features tied to a user level. If (default hehehe) Beginner, then we'll do as much as we can for you, if  Intermediate, some.. Advance.. you're on your own.

Issue #5155, opened 2025-01-25
Batch Render - dont bother with the gif background images or yellow effect strips?


derwin12:
No need spending time on these processes during batch render...
wonder if there is anything else that might be good to turn off?

cybercop23:
I'd say, don't even load the view, but I  haven't looked to see if/how it is possible

PR #5157, opened 2025-01-26
No bkgrd effect images or effect indicators during batch render #5155

derwin12:
Improve batch render performance by not doing the gui bling. #5155


dkulp:
This is definitely not how this needs to be done.  This just doesn't display them on the effect grid, but during batch render, the effect grid would only be rendered once at the beginning when none of these are calculated anyway.   This really needs to turn off the setting at the start and restore at the end or something so that the actual calculation and memory storage of the background effects is not calculated during render.

derwin12:
Oh .. okay will take a look into that..

PR #5166, opened 2025-01-28
Windows ISPC *For Consideration*

derwin12:
For discussions only .. This is my local code for running ispc - nothing special. 

removed the #ifndef __WXMSW__  and the updated the vs2022 project files.

One thing is ispc compiled for what? 
`ispc.exe "%(FullPath)" -o "$(IntDir)%(Filename).obj" --target=host`

This compiles based on my PC's capabilities - do we need to set that to a value that is our now new low-bar for cpu requirements.
What was used on the mac size?

### DIFF
--- a/README.windows
+++ b/README.windows
@@ -108,3 +108,16 @@ Here are the steps to Code::Blocks for Editing .wxs files in Windows.
      by double-clicking on the .cbp file in the xLights directory.
      Then Open a Dialog, Panel, or Frame in the Left Panel to edit it.
 
+
+Instructions for Setting up ISPC for Windows
+============================================
+
+ISPC is a compiler for a variant of the C programming language, with extensions for "single program, multiple data" (SPMD) programming. 
+It is used in xLights for the generation of AVX1, AVX2, SEE2 and SSE4 optimized code.
+
+  a) Download ISPC from
+        https://github.com/ispc/ispc/releases/download/v1.25.3/ispc-v1.25.3-windows.zip
+
+  b) Extract the zip file to a folder of your choice.
+
+  c) Add the path to the ISPC executable to your PATH environment variable.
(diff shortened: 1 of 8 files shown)


dkulp:
Most likely you want to use "--target=avx2-i32x16 --target=avx1-i32x16  --target=sse4.2-i32x8 --target=sse2-i32x8 --arch=x86_64" which is what is being used for Linux..   It compiles in all four and at runtime will select the one most appropriate based on the cpu.

computergeek1507:
can you pull in my workflow change and to see if it builds on github

derwin12:
I think I did what you asked .. 
Can I run a build/installer script on my machine? Is that msw\build_VS_x64.cmd ?
Would that create a zip I can pass to others to test?

derwin12:
Simple tests .. 5secs off of 60s render.

computergeek1507:
You didn't merge in the GitHub action change to see if it compiles in github

derwin12:
This latest push - I didnt have the HASISPC defined - I had it blocked out so it only did one file with ISPC.
I think it is in there .. I see the build ispc working..
image
Lets see the results of this next one. When I do release on my side now I am seeing upwards of 25-30% speed up.

computergeek1507:
If it builds why not remove the the special "-ISPC" modifier and merge it, I can added the install code to the release and nightly action. If it works ship it.

derwin12:
Nervous .. lol.  It is 30% faster in these test I have been doing. I know there is one big bug with ISPC reported last night.
Lets test for a couple days but it should work. Not sure what Dan wants to do with the #ifdef stuff - just remove it??
image

computergeek1507:
YOLO it, like the kids say

cybercop23:
Can the layering bug be affecting the performance here? If it doesn't process layers, it will be faster. I see your PR is compiled, I'll download and test it this afternoon.

dkulp:
I think the main thing left to update before merging is the README.windows.

PR #5175, opened 2025-02-01
ISPC VideoEffect

derwin12:
Trying out some ISPC coding. This works with one (big) exception. The transparentblack option - it leaves weird lines on say bars effects where there should be no impact. Anyways, looking for any comments on how this is coded, how to improve etc.
I used a CustomBuild in vs2022 to build the ispc.h fie, but I had to hard code in the directory. It didn't like ($Directory).

Shaved 20sec off the GreatestShowman - again about 20-30% faster.

`ispc.exe "%(FullPath)" -h "effects\\ispc\\%(Filename).ispc.h"`


xLights/Xlights.vcxproj.filters
```diff
@@ -2216,6 +2216,9 @@
       <Filter>Preferences</Filter>
     </ClInclude>
     <ClInclude Include="chatGPT.h" />
+=========
+    <ClInclude Include="effects\ispc\VideoFunctions.ispc.h" />
```
computergeek1507, line 2220: That isn't going to build

Issue #2787, opened 2021-08-31
Carry Forward Render Styles from Individual Models to Groups on Import


MrPaulAR:
**Describe the solution you'd like**
When importing individual models onto a group carry forward the render styles as expected when available.
* Default becomes Per Model Default
* Per Preview becomes Per Model Per Preview
* Single Line becomes Per Model Single Line

**Describe alternatives you've considered**
Manual

derwin12:
I was showing Paul that the option was added to change to per model afterwards via the right click option on the models list. But what a bout adding a check box on the mapping page, that would do that at import time for the selected mapped models. So if you know you are mapping from a single prop to a group in your layout, you could put the checkbox on, and it would do that "per model" conversion then? .. and ideally save that option in the "save mapping" function.

cybercop23:
@derwin12 now that you develop.. thoughts? Maybe that should be default with the model/group indicator in zip file imports.

AGFazio:
WooHoo!!!!!  This would be great!!!

PR #5178, opened 2025-02-04
Import Effects add option to convert to 'Per Model' #2787

derwin12:
Additional option on import mapping that when checked, if you map a single model to a group it will convert the render style to 'Per Model'. This is similar code that is available on the sequencer, model right click. It only convers Default, Single Line and Per Preview to 'Per Model'. #2787


derwin12:
Found an issue here..

Issue #5124, opened 2025-01-16
You cant import a Cube xlights xmodel file


derwin12:
You can export a cube with "export xlights model" but the import of the .xmodel does nothing. 
It simply says `log_base [ERROR] GetXlightsModel no code to convert to Cubemodel`

cybercop23:
If the XMLSerializer ever picks back up (@AzGilrock) then that will be automagically handled.

cybercop23:
@AzGilrock TBH I don't recall where I left it. I think the only part that's missing was the Layout 2/3 point view.
Here's the PR - lots of confilcts that need to be solved maybe new discussion points.
https://github.com/xLightsSequencer/xLights/pull/4536
If/when ready to start it backup, please LMK and we can figure out where we are and how to restart it.

PR #5203, opened 2025-02-15
Import of Cube Model #5124

derwin12:
Add ability to import the cube xmodel file. #5124


computergeek1507:
With changing the parameter names, you will break existing exported files

derwin12:
I found that the model property name didnt match the getproperty and write so it wouldnt have even written out the value into the .xmodel. This corrects that so that the value is indeed saved into the .xmodel and hence is available on the subsequent import.

PR #5205, opened 2025-02-15
Only sesnd examples that have actual mappings

cybercop23:
Keeps payload and token usage down

### DIFF
--- a/xLights/xLightsImportChannelMapDialog.cpp
+++ b/xLights/xLightsImportChannelMapDialog.cpp
@@ -2501,7 +2501,7 @@ std::string xLightsImportChannelMapDialog::BuildAlreadyMappedPrompt(const std::l
 
     std::string exampleMappings = "<exampleMappings>\\n";
     for (const auto& it : targetModels) {
-        if (filter(it)) {
+        if (filter(it) && it->HasMapping()) {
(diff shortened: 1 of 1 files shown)


keithsw1111:
I am not sure why you would need this ... the filter is already checking for the mapping. If it is not working you should modify the filter lambda instead.

cybercop23:
I think in AIStrandMap and AINodeMap => BuildTargetModelPrompt & BuildAlreadyMappedPrompt should be m->_mappingExists vs m->IsMapped(), however within BuildAlreadyMappedPrompt() the only way I can get a clean non-empty exampleMappings was to have => && it->HasMapping() Without it it created a 1.7M token promot and thankfully it got stopped so I didn't use all my $.

keithsw1111:
> I think in AIStrandMap and AINodeMap => BuildTargetModelPrompt & BuildAlreadyMappedPrompt should be m->_mappingExists vs m->IsMapped(), however within BuildAlreadyMappedPrompt() the only way I can get a clean non-empty exampleMappings was to have => && it->HasMapping() Without it it created a 1.7M token promot and thankfully it got stopped so I didn't use all my $.

So make the change in the lambda.

Issue #2642, opened 2021-05-15
Feature Request: Search box for Models and Groups


ejordan107:
Not finding this mentioned, so wanted to see what your thoughts are on this feature.  It is very helpful already when adding models to groups but unable to search the models and groups themselves at the top of the Layout view.

earft01:
I'm surprised nobody else has interest in this feature?

Clyde-L:
You have multiple ways to organize your sequence screen.  Use Display Elements and setup a sequencing view that you can click on  to move directly to a specific prop or group.  You can quickly from different areas of sequencing without the entire screen being cluttered with every prop and group.

PR #5218, opened 2025-02-19
Add search support for Layout Group/Model & SubModels

cybercop23:
Ctrl|Cmd & F to search start of Group or Models
Ctrl|Cmd & Shift & F to search within SubModels

#2642

### DIFF
--- a/xLights/LayoutPanel.cpp
+++ b/xLights/LayoutPanel.cpp
@@ -105,8 +105,6 @@ static wxRect scaledRect(int srcWidth, int srcHeight, int dstWidth, int dstHeigh
 	return r;
 }
 
-static bool ctrlFPressed = false;
-static bool ctrlshiftFPressed = false;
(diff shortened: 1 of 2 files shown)


cybercop23:
#5218

xLights/LayoutPanel.cpp
```diff
@@ -105,6 +105,10 @@ static wxRect scaledRect(int srcWidth, int srcHeight, int dstWidth, int dstHeigh
 	return r;
 }
 
+static bool ctrlFPressed = false;
+static bool ctrlshiftFPressed = false;
+wxTreeListItem lastFoundItem = nullptr;
+
```
dkulp, line 111: Why are these statics instead of instance variables in the LayoutPanel?

cybercop23: No real reason other than thought since we only have one LayoutPanel, would not matter. I've moved it to be more aligned and have it as part of the instance.

PR #5225, opened 2025-02-20
Add support organizational folder only

cybercop23:
This allows users to group multiple folders under one better organize the layout tab. It will not allow effects to be placed on it, copied over it or moved to it.  Import Effects will also skip displayting this group, however embeded groups will still display as normal.

image

image

### DIFF
--- a/xLights/BitmapCache.cpp
+++ b/xLights/BitmapCache.cpp
@@ -610,6 +610,11 @@ wxBitmapBundle BitmapCache::GetPapgayoXIcon() {
 wxBitmapBundle BitmapCache::GetModelGroupIcon() {
     return CreateBitmapBundleFromXPMs(16, "ModelGroup", {model_16, model_64, model_64, model_64, model_64});
 }
+
+wxBitmapBundle BitmapCache::GetModelOrgGroupIcon() {
+    return CreateBitmapBundleFromXPMs(16, "xlART_PADLOCK_CLOSED", { padlock_close_14, padlock_close_28, padlock_close_28, padlock_close_28, padlock_close_28 });
+}
+
 wxBitmapBundle BitmapCache::GetFPPIcon() {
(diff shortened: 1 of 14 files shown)


cybercop23:
Found and issue working on correcting it

dkulp:
I'm really not sure on this one as I'm struggling to see any benefit.   It's pretty much just the same groups, just with a bunch of restrictions.   Why wouldn't you just use a normal group?

cybercop23:
That's just it. This prevents users from putting effects on such group since they really shouldn't have any effects on them.
This group is to "ease' the ever growing list of models/submodels folders, but when you're in sequencer, you don't want to accidentely map an effect to this entire group. Obviouslly only kicks in if the folder is set as such.
IMHO makes the Model list be much more manageble and organized.

dkulp:
I doesn't change the model list at all other than the icon.   Thus, to me, it's completely pointless there.

On the sequence tab, it just ends up wasting valuable space with rows that aren't sequenceable.   Again, it's just a group that is basically "locked".      If you wanted to accomplish that, just allow any model in the sequence to be "locked" to prevent changes. 

Anyway, I'm against the change and don't plan on merging it.   Anyone else have comments?

computergeek1507:
I thought it was kind of pointless

cybercop23:
It makes the Model List be so much cleaner by moving all of the HD specific model groups into another group. YES you can do that now. The only thing this adds is a guiderail to help prevent users from accidentaly placing an effect on the whole group, vs the groups inside that. It obviously has no effect unless the user checks the folder attribute, so won't impact anyone that doesn't want to use it, but it will be great to have it you manage your showfolder this other way.

cybercop23:
Here's some more background.. Last year's xLights Around the World, Matt has a similar folder structure that he's using to organize groups and models. See below. So he could, but knows better, not to put effects on those groups with --- but all this will do is lock and prevent (if you so choose to set it up that way) putting effects on such groups.

image

AGFazio:
Get this pushed. This would be a great option to have.

dkulp:
Closing.  Not merging.

PR #5244, opened 2025-03-04
Default Galaxy, Shockwave, and Fan to Scale to Buffer for 2025.04+

computergeek1507:


marklancer:
Thanks Scott.  I tested it this morning and it works as expected and doesn't break existing sequences.

Issue #5243, opened 2025-03-04
Preset key bindings don't work for presets in a sub-folder


marklancer:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
If you create a key binding for a preset it only works if the preset is in the main preset folder.  If it is in a sub-folder nothing happens.  For example if you create subfolder called Megatree to help organize your presets you are unable to call it using a preset

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)

**Expected behavior**
Key bindings should be able to be linked to presets in either the main preset folder or a sub-folder

**Screenshots**

**Versions (please complete the following information):**
 - OS: [e.g. Windows 10, Mac 10.14.5, etc...]  Windows 11
 - xLights version [e.g. 2019.37] 2025.03

**Additional context**

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

computergeek1507:
are you putting the group name in the path in the key binding? ie "group/effect" would be the path in the key binding

cybercop23:
True story....

![Image](https://github.com/user-attachments/assets/64b1fe69-3e9d-4fb7-89d8-3edf509b6445)

computergeek1507:
Wrong slash

computergeek1507:
![Image](https://github.com/user-attachments/assets/197c7428-2a5e-4c54-a2ca-17d1a22d3c7c)

cybercop23:
You don't get to choose it. It is a dropdown prefilled... Ahhh.. So maybe replace the separator in the KeyBinding code.. let me try

![Image](https://github.com/user-attachments/assets/24550395-351d-4923-b3f4-3b4f8f312c0a)

cybercop23:
Yup.. fixed.

PR #5245, opened 2025-03-04
#5243 Fix apply preset using key bindings

cybercop23:


computergeek1507:
Did you verify this still works with the effects preset dialog?

cybercop23:
Yes sir.

Issue #5249, opened 2025-03-07
Dont allow Audio Pre/Post changes if no media.


derwin12:
Code was updated to not allow pre/post if animation, but it doesn't stop you from doing if it is set to media and the media is blank. Just need to add a if media <> blank in there.
ie. Open file settings, switch from anim to media then go put some values in the pre or post and Adjust.

computergeek1507:
Tell Alex to fix his code then

cybercop23:
Anybody can fix it. We *need* more time from PR merge to new version so we can get some others to test and break it. Really needs a better timeline for PR -> Merge -> Release.

computergeek1507:
While I do agree PR Merge Release is pretty quick. The bugs in this feature have been present since it was written in Augusts. The general rule is if you add a feature, you should support it. I have attempted to fix some bugs in it but I don't think I should have to spend my time fixing someone else's code because they didn't spend the time to test it adequately. If you are unwilling to fix/support the code you have written, I will remove it.

PR #5253, opened 2025-03-09
Check for missing media #5249

derwin12:
Dont allow you to fire off the pre/post audio if there is no media file entered.


xLights/SeqSettingsDialog.cpp
```diff
@@ -1875,6 +1875,10 @@ void SeqSettingsDialog::OnButton_MusicOpenClick(wxCommandEvent& event)
 
 void SeqSettingsDialog::OnButton_AddMillisecondsClick(wxCommandEvent& event) {
     const std::string inputFile = TextCtrl_Xml_Media_File->GetValue();
+    if (inputFile.empty()) {
```
cybercop23, line 1878: I was thinking of this in ProcessSequenceType()
Button_AddMilliseconds->Enable(type == "Media" && xml_file->HasAudioMedia());
like the other buttons that aren't enabled if the requiremnets aren't met.
Just my 2c. Was away almost all day and I'll be AFK tomorrow too.

computergeek1507:
Thanks Daryl

Issue #3725, opened 2023-02-11
Matrix with 1 string does not configure alternate nodes


sslupsky:
**Describe the bug**
When you create a matrix with one string and check the Alternate Nodes checkbox and view the node layout, the nodes are arranged the same as when the Alternate Nodes checkbox is unchecked.

**To Reproduce**
Creat a matrix
Set # Strings = 1
Check Alternate Nodes
View Node Layout

**Expected behavior**
The nodes should alternate

**Screenshots**
Alternate Nodes unchecked:
![Screen Shot 2023-02-11 at 3 50 01 PM](https://user-images.githubusercontent.com/16888734/218284410-e832a14f-996f-4073-9b54-20dc588ce7dc.png)

Alternate Nodes checked:
![Screen Shot 2023-02-11 at 3 50 35 PM](https://user-images.githubusercontent.com/16888734/218284406-56531786-72c9-46f0-878e-8edb5a2339c2.png)

xLights 2023.2

computergeek1507:
I was able to reproduce this, the model buffer is correct, the node layout view is just wrong. I have a fix but it might break other models. I have to do more testing.

cybercop23:
@computergeek1507 is this something that should still be considered or does this fall under custom model or other type, candy cane/stick or MT/SS to get this to work the way he's expecting?
![image](https://github.com/xLightsSequencer/xLights/assets/23619433/bcc9d58d-18ba-464b-b583-c535a3c597aa)


@sslupsky if these alternatives work for you, please close this issue.

computergeek1507:
It's actually a bug in the GUI

derwin12:
PR is in.

PR #5257, opened 2025-03-10
Account for alt nodes when display node layout #3725

derwin12:
The node layout didn't show the alt nodes for single row buffer scenarios #3725


derwin12:
I wondered about the purpose of the if then .. let me check and refactor.

computergeek1507:
I assume it's for Arch models that use parameter 1 for the arch count, not string count

derwin12:
Removed the if code .. didnt find any issues with my tests. The parm1 stuff is don't outside of this display code so not an issue here.

PR #5270, opened 2025-03-16
Add right click to maximize House Preview

derwin12:
Adding the ability to maximize the undocked house preview window. If there is a wxwidgets way to do this, I couldnt find it.
I also add a tooltip to the Save Color Pallete icon and the refactor of the colorpanel code as a result.


derwin12:
Found there is a hidden double click that accomplishes this.

Issue #5269, opened 2025-03-16
Crash on discover


derwin12:
Latest version 2025.04 - I believe he deleted the controller, so models were in limbo. He hits discover, it adds the controller. If he saves or visualize, it crashed.

Work around was to add a new ethernet controller and simply put the models on the visualizer.

Attached is his dbg file which hopefully will duplicate the issue.

[xLights_dbgrpt-8764-20250315T191624.zip](https://github.com/user-attachments/files/19267460/xLights_dbgrpt-8764-20250315T191624.zip)

jeffwalkinshaw:
This is for my XLights install.

cybercop23:
Unable to replicate in debug ot 2024.5 even with his show folder. 

- Discovered my controllers
- Moved all models from the House Falcon to mine ans saved
- Deleted Controller
- Discovery
- Was able to both save and visualize. - Models came back on contoller (as expected)
-

derwin12:
Jeff - if you can come back to the zoom room, Maybe we can see the values returned with the Discover or something.

derwin12:
Jeff - one more thing .. when you hit discover, does it come back as DDP or E131?
(You would have to temporarily delete the controller from your list then hit discover)

computergeek1507:
What type of controllers are on his network and is it a flat network or FPP bridged?

cybercop23:
Seems flat. There's a player in the mix. I can't duplicate it and also can't duplicate Ron's issue but he has a very similar issue on his network also with a Falcon.

derwin12:
Anyone that can test/verify this? @keithsw1111 -- need a newer Falcon on a flat network.

bloojoop:
I can setup this up and happy to supply any test results and even configure a packet capture if it would help.  

Ron A.

derwin12:
Even a crash report is a good first step..

bloojoop:
[xLights_dbgrpt-9776-20250603T191216.zip](https://github.com/user-attachments/files/20582164/xLights_dbgrpt-9776-20250603T191216.zip)

Of course there is one above attached but here is one from a show dir I just replicated the behavior.

bloojoop:
as I described in one of my comments, it occurs when the falcon is in DDP
and it gets discovered over the local network (not proxy config).    It
shows in xlights as e131 even though controller is in ddp mode. If I
manually change the controller to e131 then it will get
discovered correctly and no issue found.  It feels like the data that gets
[...]

PR #5272, opened 2025-03-16
Check for crash condition #5269

derwin12:
Not clear how it occurs or if this is the exact crash, but here is a couple of safe guards.


computergeek1507:
Are you looking at the commits before you make pull request, you included a temp file. I would also probably just put a null check at the beginning and return if null

derwin12:
Oh crud..missed that.

PR #5277, opened 2025-03-17
Check for crash condition #5269

derwin12:
Resubmission: Not clear how it occurs or if this is the exact crash but here is a couple of safe guards.

There is a check for empty _outputs at the start, I am thinking that the list is initialized but not correct and has pointers to nothing.

Scott, this is currently coded, is this what you were suggesting?
    if (_outputs.size() == 0) return false;

### DIFF
--- a/xLights/outputs/ControllerEthernet.cpp
+++ b/xLights/outputs/ControllerEthernet.cpp
@@ -840,8 +840,11 @@ bool ControllerEthernet::SetChannelSize(int32_t channels, std::list<Model*> mode
         it2->AllOff();
         it2->EndFrame(0);
     }
-    
+
     if (_type == OUTPUT_ZCPP || _type == OUTPUT_DDP || _type == OUTPUT_TWINKLY) {
+        if (_outputs.front() == nullptr) {
+            return false;
+        }
         _outputs.front()->SetChannels(channels);
         return true;
     }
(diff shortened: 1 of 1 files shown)


xLights/outputs/ControllerEthernet.cpp
```diff
@@ -840,8 +840,11 @@ bool ControllerEthernet::SetChannelSize(int32_t channels, std::list<Model*> mode
         it2->AllOff();
         it2->EndFrame(0);
     }
-    
+
     if (_type == OUTPUT_ZCPP || _type == OUTPUT_DDP || _type == OUTPUT_TWINKLY) {
+        if (_outputs.front() == nullptr) {
```
computergeek1507, line 845: This is the same as outputs.size() == 0 I believe by the logic, might be more obvious to do that instead

derwin12: That is already the first line in the proc. My thinking is the list is there but contains invalid pointers.

derwin12: Note Line 901-902 (below) also checks for != nullptr.

```
_outputs.push_back(new E131Output());
if (dynamic_cast<E131Output*>(_outputs.back()) != nullptr) {
```
It is like we know it is an E131 but we don't know how many outputs it has (yet)??

I can't duplicate it with my V3 Falcon.

derwin12:
Closing until we can find someone to duplicate the environment..

Issue #5278, opened 2025-03-17
Add Export States


cybercop23:
Currently SubModels has an Export SubModels to other Models feature. Do the same for States.

derwin12:
Added in next release

Issue #4356, opened 2024-02-18
Add Export Function to State Effect Similar to That in Sub-Models Dialog


MCStoff:
Presently, the only way to transfer state effects among models is to do an import from a model/file.  This allows for a one to one import that requires multiple iterations when importing to multiple identical models.

The sub-model dialog allows for an export of sub-models to other models.  This allows a one to many export.  Adding the export function to the state effects would allow a one to many export, thereby avoiding the necessity of multiple iterations.  Further, it would create internal consistency in the program.

cybercop23:
Done in 2025.6

PR #5279, opened 2025-03-17
#5278 Add Export Faces & States to other Models

cybercop23:
Works just like SubModels Export and replaces all States.


cybercop23:
Added faces too. So now all 3, Faces, States and SubModels can export to other Models.

cybercop23:
Add option to export Faces, States and SubModels to chosen models to make it easier to keep multiple of the same model in-sync

image

MCStoff:
This will close out #4356 as well.

PR #5284, opened 2025-03-21
Ability to change Video Render Algo

derwin12:
This came about with the ticket from Michael @ SS. The video gets washed out for some scenarios. With this change folks can alter what algo is used to render the video. The default remains as is with all the other ffmpeg options now selectable if desired.
From my readings it can vary based on the type of content and size/distance. The changes are subtle, basically how does it treat the small stuff in the background.

eg. Lanczos excels at maintaining sharpness and reducing aliasing, which is critical for cartoons with clean lines and high-contrast boundaries.
Bicubic is good for downsizing because it reduces aliasing (jagged edges) while keeping the output clean.

Thoughts? Any concerns or just leave existing code as is??

image


computergeek1507:
From a performance standpoint, it would be better if you stored the setting in the video render class as the enum and decoded it from the string setting once during the video render construction or in the video effect settings. You are currently "decoding" the string to enum on every frame. Probably in the real world it would not make much of a difference, but it is probably a little less performative

derwin12:
I will take a crack at doing the change..

Issue #5260, opened 2025-03-11
Color Coding Handle Bars


tgrinnall:
it would be nice when you click on a prop and scroll down to the size/location that the XYZ accordance would match the colors of the handle bars on the props. so when you look at the prop you know what number = what handle bar. I use this method to fine tune the placement of my props to get an effect to hit a prop correctly or to even a line them properly in the layout so they match what is actually outside for when i'm doing my horizontal and vertical testing. i believe the X should be Red, Y should be Green and Z should be Bue. i understand some people may have issues cause they are color blind. If this is truly and issues for this what if we added the wording in each field then or made a spot under prefrences color to check mark a box so it would change them to color.

PR #5285, opened 2025-03-22
Color X,Y,Z Coordinates #5260

derwin12:
Minor tweak so they match the color handle bars in the layout.


computergeek1507:
No screenshot of how light and dark mode look?

derwin12:
Used the same calls as the start/end of a poly.
ie
image
image

and
image
image

PR #5297, opened 2025-03-29
Potential Video Crash Fix #5296

### LINKED ISSUE #5296: Video Effect Renders but crashes if you click thru them
2025.04 Render All Works and it shows fine but if you click thru the various video effects, the 4 or 5th one will crash it.

Sequence reduced to a handful of effects. The videos are layered using submodels (left and right)

[Video Crash 3.zip](https://github.com/user-attachments/files/19522503/Video.Crash.3.zip)

[xLights_dbgrpt-560-20250329T142327.zip](https://github.com/user-attachments/files/19522504/xLights_dbgrpt-560-20250329T142327.zip)

derwin12:
The crop bottom was creating an exception esp if clicked quickly thru the effects. #5296

### DIFF
--- a/xLights/effects/ispc/VideoFunctions.ispc
+++ b/xLights/effects/ispc/VideoFunctions.ispc
@@ -24,6 +24,16 @@ export void VideoEffectProcess(const uniform VideoData &data,
                                uniform int startIdx, uniform int endIdx,
                                uniform uint8<4> result[])
 {
+    if (data.buffer_width <= 0 || data.image_width <= 0 || data.image_height <= 0) {
+        return;
+    }
+    if (data.ch != 3 && data.ch != 4) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


xLights/effects/ispc/VideoFunctions.ispc
```diff
@@ -24,14 +24,42 @@ export void VideoEffectProcess(const uniform VideoData &data,
                                uniform int startIdx, uniform int endIdx,
                                uniform uint8<4> result[])
 {
+    if (data.buffer_width <= 0 || data.image_width <= 0 || data.image_height <= 0) {
```
computergeek1507, line 27: Why only checking buffer width is zero or less and not buffer height

derwin12: Pretty sure buffer height as to be atleast 1 - but I can certainly add it for completeness.

PR #5313, opened 2025-04-08
Pack FSEQ into minimum number of Universes

jhinklejr:
HinksPix Long Range has required multiple small Universes to properly build a E131 table This change reformats the created FSEQ by packing and reording Universes to minimize E131 packets

### DIFF
--- a/controllers/hinkspix.xcontroller
+++ b/controllers/hinkspix.xcontroller
@@ -6,7 +6,6 @@
 			<SupportsPixelPortNullPixels/>
 			<SupportsPixelPortDirection/>
 			<SupportsPixelPortColourOrder/>
-			<SupportsUniversePerString/>
 			<DMXAfterPixels/>
 			<SupportsUpload/>
 			<SupportsRemotes/>
@@ -45,6 +44,7 @@
 			<SupportsPixelPortDirection/>
 			<SupportsPixelPortColourOrder/>
 			<SupportsUniversePerString/>
+			<SupportsModifyFseqFrame/>
 			<DMXAfterPixels/>
 			<SupportsRemotes/>
 			<SupportsSmartRemotes>16</SupportsSmartRemotes>
@@ -60,9 +60,6 @@
 			<DisableMonitoring/>
 			<PixelProtocols>
 				<Protocol>ws2811</Protocol>
-				<Protocol>ws2801</Protocol>
-				<Protocol>tls3001</Protocol>
-				<Protocol>apa102</Protocol>
 			</PixelProtocols>
 			<SerialProtocols>
 				<Protocol>dmx</Protocol>
@@ -92,6 +89,7 @@
 			<SupportsPixelPortDirection/>
 			<SupportsPixelPortColourOrder/>
 			<SupportsUniversePerString/>
+			<SupportsModifyFseqFrame/>
 			<DMXAfterPixels/>
 			<SupportsRemotes/>
 			<SupportsSmartRemotes>16</SupportsSmartRemotes>
@@ -107,9 +105,6 @@
 			<DisableMonitoring/>
 			<PixelProtocols>
 				<Protocol>ws2811</Protocol>
-				<Protocol>ws2801</Protocol>
-				<Protocol>tls3001</Protocol>
-				<Protocol>apa102</Protocol>
 			</PixelProtocols>
 			<SerialProtocols>
 				<Protocol>dmx</Protocol>
--- a/xLights/FileConverter.cpp
+++ b/xLights/FileConverter.cpp
@@ -24,6 +24,8 @@
 #include "FileConverter.h"
 #include "UtilFunctions.h"
 #include "outputs/OutputManager.h"
+#include "Controllers/BaseController.h"
+#include "Controllers/ControllerCaps.h"
 #include "outputs/Controller.h"
 #ifndef FPP
     #include "xLightsMain.h"
@@ -35,6 +37,12 @@
 #include "xLightsVersion.h"
 #include "ExternalHooks.h"
 #include <log4cpp/Category.hh>
+#include "outputs/ControllerEthernet.h"
+
+
+void PackUniverse_FseqFrame_Init(ConvertParameters* params);
+void PackUniverse_FseqFrame_Release(void);
+void PackUniverse_ModifyFseqFrame(uint8_t* data, int datasize);
 
 static const int MAX_READ_BLOCK_SIZE = 4096 * 1024;
 
@@ -1624,6 +1632,7 @@ void FileConverter::ReadFalconFile(ConvertParameters& params)
     delete file;
 }
 
+
 void FileConverter::WriteFalconPiFile(ConvertParameters& params)
 {
     static log4cpp::Category &logger_conversion = log4cpp::Category::getInstance(std::string("log_conversion"));
@@ -1752,14 +1761,19 @@ void FileConverter::WriteFalconPiFile(ConvertParameters& params)
             }
         }
     }
-    
+
+    PackUniverse_FseqFrame_Init(&params);
 
     file->writeHeader();
+
     size_t size = params.seq_data.NumFrames();
     for (int x = 0; x < size; x++) {
+        PackUniverse_ModifyFseqFrame(&params.seq_data[x][0], stepSize);
         file->addFrame(x, &params.seq_data[x][0]);
     }
     file->finalize();
     delete file;
     logger_conversion.debug("End fseq write");
+
+    PackUniverse_FseqFrame_Release();
 }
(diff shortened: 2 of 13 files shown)


dkulp:
This is not likely to be merged... Anything that changes the data in FSEQ file such that it does not match the channel numbers and values that are represented in memory is a non-starter.

jhinklejr:
> This is not likely to be merged... Anything that changes the data in FSEQ file such that it does not match the channel numbers and values that are represented in memory is a non-starter.

Dan - it does NOT change anything in the FSEQ.  The channels that make up universes are compacted so that they are minimized.  Your FSEQ format and or data is not being affected.

jhinklejr:
> I have no idea what you mean by minimized. But if it changes the channel numbers AT ALL, then it's a no-go.

Please define your meaning of channel numbers.  The number of Universes AND the number of channels that are identified in the xlights "Controller" tab does NOT change.  I am just changing which channels goes into which universe.

dkulp:
In xLights, if channel 6382 maps the red channel of "Snowflake 1, pixel 1", then in the FSEQ file, channel 6382 MUST map to that same channel.

jhinklejr:
> In xLights, if channel 6382 maps the red channel of "Snowflake 1, pixel 1", then in the FSEQ file, channel 6382 MUST map to that same channel.

I did NOT change that in any way.  Are you THEN implying that channel 6382 that was assigned to reside in Universe 5 can NOT be relocated to Universe 6?

dkulp:
The fseq file has no concept of a universe.   Thus, I have no idea what you are talking about with minimizing universes in the FSEQ file.

jhinklejr:
I did NOT make any changes to your FSEQ file or header.  My change is a couple of function calls that occur at the bottom of the FileConverter.cpp file.

dkulp:
That's a big part of the issue.. If ANYTHING in the fseq needs to change based on the target controllers, that's a problem.   Everything you are describing needs to be done at the auto-layout level, not fseq generation, and should be made "generic" and not tied to Hinks at all.   Reducing universes/packets can benefit all the various controllers, but that has to be done during the assignment of channels for each prop/string.

jhinklejr:
As a follow up - I did a bench test on an xlights user that was using FPP to send E131 packets to his HinksPix controller and he was receiving 79 E131 universe/packets each frame (he had a dozen pease stacks or more).  Using the changes I propose - the number of universes being sent by FPP would drop to 27.  That results in less stress on the network, less packets being sent by FPP and less packets being received and processed by the controller.  I hope that helps explain my statement "minimize universes in the FSEQ".

jhinklejr:
> As a follow up - I did a bench test on an xlights user that was using FPP to send E131 packets to his HinksPix controller and he was receiving 79 E131 universe/packets each frame (he had a dozen pease stacks or more). Using the changes I propose - the number of universes being sent by FPP would drop to 27. That results in less stress on the network, less packets being sent by FPP and less packets being received and processed by the controller. I hope that helps explain my statement "minimize universes in the FSEQ".

jhinklejr:
> That's a big part of the issue.. If ANYTHING in the fseq needs to change based on the target controllers, that's a problem. Everything you are describing needs to be done at the auto-layout level, not fseq generation, and should be made "generic" and not tied to Hinks at all. Reducing universes/packets can benefit all the various controllers, but that has to be done during the assignment of channels for each prop/string.

jhinklejr:
> > That's a big part of the issue.. If ANYTHING in the fseq needs to change based on the target controllers, that's a problem. Everything you are describing needs to be done at the auto-layout level, not fseq generation, and should be made "generic" and not tied to Hinks at all. Reducing universes/packets can benefit all the various controllers, but that has to be done during the assignment of channels for each prop/string.

#1.  Only HinksPix controllers have access to "Universe per String".  All other controllers are all packed to 510/512 to start with.
#2.  If I understand the code in xlights, your models and sequence generations demands and assumes model data appears in sequential memory.  Your models (assume a 16 string model) requires all 16 strings to be sequential 1,2....15, 16.  In many cases, a users is using one of my Long Range receivers that require the data to be in a different order (say 1,3,6,2,...13).  Today - the "String per Universe" solves that issue because not the strings can be arranged simply by ordering them differently by Universe.   I looked hard to see if I could rearrange memory at the model layer but I did NOT see an easy or even possible way of doing that.  So I took the direction of changing it prior to building the file (FSEQ) that is used to transmit the data.  Since changing it there does not impact anything in the FPP and does not impact any other controller except the HinksPix - I saw it as a elegant solution.  If I was wrong and model data can be made non-sequential, I will be please to implement the change but I don't see a way.  Am I missing something?  Thanks.

dkulp:
And just to follow up...   digging into the code, it does exactly what it CANNOT do.   For each frame, it's doing memmove and other things to change the channels and stuff around.   Since this is on the live data, that means after save, the user would have to re-render to get their preview correct.     What's worse, if the user hits save twice, the resulting fseq would have things double applied which would really be screwed up.   Also, when opening a sequence, they would have to re-render everything to get the data properly setup to allow the preview to work.    Again, this CANNOT happen.

jhinklejr:
A question about model channel number.  When you generate data - say for a megatree, is the data generated per specific channel number or is the model assumed to be one big lump of memory and data generated to go into that lump?  If sequence data is generated by specific channel number then I could go into the model and rearrange channel locations but I did not see that in the code.

jhinklejr:
> And just to follow up... digging into the code, it does exactly what it CANNOT do. For each frame, it's doing memmove and other things to change the channels and stuff around. Since this is on the live data, that means after save, the user would have to re-render to get their preview correct. What's worse, if the user hits save twice, the resulting fseq would have things double applied which would really be screwed up. Also, when opening a sequence, they would have to re-render everything to get the data properly setup to allow the preview to work. Again, this CANNOT happen.

You statement is wrong as it appears you are saying I am taking FSEQ data to make the change.  Not correct.  I am taking the model data as it is going into a FSEQ file - NOT after it is in an FSEQ file.

dkulp:
> > And just to follow up... digging into the code, it does exactly what it CANNOT do. For each frame, it's doing memmove and other things to change the channels and stuff around. Since this is on the live data, that means after save, the user would have to re-render to get their preview correct. What's worse, if the user hits save twice, the resulting fseq would have things double applied which would really be screwed up. Also, when opening a sequence, they would have to re-render everything to get the data properly setup to allow the preview to work. Again, this CANNOT happen.
> 
> You statement is wrong as it appears you are saying I am taking FSEQ data to make the change. Not correct. I am taking the model data as it is going into a FSEQ file - NOT after it is in an FSEQ file.

`PackUniverse_ModifyFseqFrame(&params.seq_data[x][0], stepSize);`

That seq_data is the EXACT channel data buffer in memory that is used for the model previews and everything for each frame.   In PackUniverse_ModifyFseqFrame, there is memmove and other actions that are modifying that data.   Thus, it would be moving stuff around in memory that the previews are expecting.    Each model does not have its own data array, they just have indexes into the channel data buffer.   Thus, by moving anything around in there, you are messing things up and will require a re-render.

jhinklejr:
OK - I agree that my modified FSEQ would cause you an issue.  As an offering of compromise, what if I generated a FSEQ by a different name that the user could then decide to either use xlights's FSEQ and have a large number of universes or utilize my re-arranged FSEQ helping everyone associated with network traffic?

dkulp:
> A question about model channel number. When you generate data - say for a megatree, is the data generated per specific channel number or is the model assumed to be one big lump of memory and data generated to go into that lump? If sequence data is generated by specific channel number then I could go into the model and rearrange channel locations but I did not see that in the code.

Depends on the type of model and how it's configured.   When NOT using the auto layout stuff, many of the models have an "individual start channel" option where each string can be set to a unique start channel.  That COULD be via universe+offset or similar.  Not all models have this, but many do.   In those cases, the channels do not have to continuous and can be random blocks all over the place.      However, when using the auto layout stuff, I'm fairly certain it has to be continuous and the "individual start channel" checkbox and such is disabled.

jhinklejr:
> > A question about model channel number. When you generate data - say for a megatree, is the data generated per specific channel number or is the model assumed to be one big lump of memory and data generated to go into that lump? If sequence data is generated by specific channel number then I could go into the model and rearrange channel locations but I did not see that in the code.
> 
> Depends on the type of model and how it's configured. When NOT using the auto layout stuff, many of the models have an "individual start channel" option where each string can be set to a unique start channel. That COULD be via universe+offset or similar. Not all models have this, but many do. In those cases, the channels do not have to continuous and can be random blocks all over the place. However, when using the auto layout stuff, I'm fairly certain it has to be continuous and the "individual start channel" checkbox and such is disabled.

Thanks - I'll look at making "NOT auto layout" and see what happens.  Thoughts on the additional/differently name FSEQ file for transmitting only (NOT used in preview, etc) suggestion?

jhinklejr:
I give up - I'm all bloody - please close request.  Thanks for the replies.

derwin12:
Wouldn't  this create a mess when people use multiple controllers?

jhinklejr:
No

PR #5315, opened 2025-04-09
updating sol 3.5.0 library to the 'single' implementation.

kylegrymonprez:
Compiled the single lib version of sol vs the fully broken out version of the lib, consistent with 3.2.2


computergeek1507:
I thought I used the single header to prevent needing to compile it as a library and dy file

Issue #4184, opened 2023-12-02
XLights 2023.20


Rlbos:
Upgraded Xlights 2023.20 every time I open the save light is on. I haven't change anything on the desktop I have to render my sequences.

[Document 4.pdf](https://github.com/smeighan/xLights/files/13537069/Document.4.pdf)

dartunder:
This happens sometimes for some configurations.  For example, if universe per string is enabled.

dkulp:
Double check with .21.  I fixed at least one of these with .21, but that was related to using a base folder.

dartunder:
I don't know about his case, but the code forces a recalculation that triggers the save button if the universe per string is enabled. His situation may be different.

My case is being triggered ( outputManager->SomethingChanged()) by this code from ModeManager.cpp line 939

 if (it->IsAutoSize()) {
            auto eth = dynamic_cast<const ControllerEthernet*>(it);
            if (it->GetChannels() != std::max((int32_t)1, (int32_t)ch - 1) || (eth != nullptr && eth->IsUniversePerString())) {
                logger_zcpp.debug("    Resizing output to %d channels.", std::max((int32_t)1, (int32_t)ch - 1));

                auto oldC = it->GetChannels();
                // Set channel size won't always change the number of channels for some protocols
                it->SetChannelSize(std::max((int32_t)1, (int32_t)ch - 1), allSortedModels);
                if (it->GetChannels() != oldC || (eth != nullptr && eth->IsUniversePerString())) {
                    outputManager->SomethingChanged();

                    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_NETWORK_CHANGE, "ReworkStartChannel");
                    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_NETWORK_CHANNELSCHANGE, "ReworkStartChannel");
                    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_UPDATE_NETWORK_LIST, "ReworkStartChannel");
                    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_SAVE_NETWORKS, "ReworkStartChannel");
                    xlights->GetOutputModelManager()->AddASAPWork(OutputModelManager::WORK_RESEND_CONTROLLER_CONFIG, "ReworkStartChannel");
                }
            }
        }

derwin12:
He is/was in zoom..  happens in .21 nightly.
[...]

PR #5329, opened 2025-04-18
Universe Per String triggers the Save Layout Bug #4184

derwin12:
If Universe per string is enabled then don't trigger the save icon on initial load of xLights. Bypass the recalc on initial load. Use the mSequencerInitialize flag that indicates that the sequencer panel has been loaded to indicate if we are on the initial load or not.

### DIFF
--- a/xLights/TabSetup.cpp
+++ b/xLights/TabSetup.cpp
@@ -885,7 +885,7 @@ void xLightsFrame::NetworkChange() {
 void xLightsFrame::NetworkChannelsChange() {
 
     static log4cpp::Category& logger_work = log4cpp::Category::getInstance(std::string("log_work"));
-    logger_work.debug("        NetowrkChannelsChange.");
+    logger_work.debug("        NetworkChannelsChange.");
 
     _outputManager.SomethingChanged();
     _outputModelManager.AddASAPWork(OutputModelManager::WORK_RESEND_CONTROLLER_CONFIG, "NetworkChannelsChange");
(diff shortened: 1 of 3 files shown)


keithsw1111:
This opens an admitably small risk that it is not in sync ... at a minimum we should then force a recalc before any upload to ensure it is correct. Otherwise strange bugs could surface particularly for folks saving layout but not networks or vice versa.

derwin12:
Could the channels change from the last save/close of xlights to when it is opened that it requires a recalc? If you change anything in the controller tab it will catch that. If you change anything in the layout, it would catch that. So why is the recalc needed just on restart just for the Universe Per String case?

keithsw1111:
The problem is not when you make a change ... its when you fail to save one and end up with layout/controller out of sync. There are as i say unlikely but possible ways for this to happen.

cybercop23:
Wish we had telemetry data to see how many have the two buttons linked. If they are linked, doesn't that reduce the chances?
Maybe only skip it if buttons are linked together.

keithsw1111:
Yes and no ... the biggest cause is likely people closing xlights and being prompted to save. Even if the buttons are linked these saves are independent. Maybe they should not be if the linkage is there.

dkulp:
Personally, I think we should remove the setting and always have them "linked".    Kind of a distant first step in merging the three xml files into a single "show config" file.

keithsw1111:
Linking was the first step on this path ... I dont see why they need to be separate save steps ... joining the files will have impacts as many folks have built tools assuming they are separate.

Issue #5343, opened 2025-04-28
Copy State corrupts the output for a specific set of inputs


SopherConklin:
A video is here:  https://youtu.be/S-_S_ZhmBl0

I have found this problem on the Patriotic Gnome state on the ShowStopper MiniTree.   I have not investigated to try to find other states that this may occur with.

Here's the summary
The Patriotic Gnome state seems pretty generic, although it does have some blank lines scattered throughout (I note you rejected my earlier suggestions to improve the editing of the States screen to make editing easier)

If you Select the Patriotic Gnome state and then select "Copy" from the "..." at the top, it creates an exact copy including the blank lines.  The result when applied to the model is identical to the original.

If you select "Add" from the top and make a new, blank state and then right click in the empty state and select Copy States and just select the Patriotic Gnome in the list - it creates a copy of the original with all blank lines removed.  It appears that all the lines are the same, including the node lists and the color, but if you apply this state to the model the "u" state is very different from the original.  Note the "a" and "s" states are identical to the original.

xLights version 2025_05

Link to the xmodel file (I'm amazed that this doesn't allow the attachment of xmodel files)
https://www.facebook.com/groups/1052895671989826/permalink/1751109175501802/

cybercop23:
This video is private. Can you please adjust.

SopherConklin:
Fixed - I was sure I set it as Unlisted, but sometimes if you do it too
early in the upload process it doesn't seem to take.

derwin12:
Not sure we guarantee a render order in the state itself .. but there are definitely duplicates in there..

![Image](https://github.com/user-attachments/assets/83cacebc-abbb-47f4-95f3-ade6debe167f)

Will take some combing thru the code to see how it determines the order - or if it just "luck". The real answer is don't duplicate the nodes in the particular state item.

derwin12:
Wow .. ok so line 9 is stored as "S9" in the system, line 10,11,12 etc are stored as "S10", "S11" etc. Well you guessed it .. "S10" sorts ahead of "S9". That means Line 9 is applied AFTER the others and hence hides those notes on line 17. That is one weird bug/feature.

cybercop23:
Guess the infamous wxWidget sort - may need the numbers aware sort logic.

computergeek1507:
More like how string are sorting in most computer languages, s1,s10,s11,s2 without a custom compare function.

computergeek1507:
Utils class should already has number aware sorting methods

[...]

PR #5344, opened 2025-04-29
Add number aware sort to State defn #5343

derwin12:
Handles and enforces the case where state definition lines need the render to happen in numeric order. #5343

### DIFF
--- a/xLights/effects/StateEffect.cpp
+++ b/xLights/effects/StateEffect.cpp
@@ -443,9 +443,20 @@ void StateEffect::RenderState(RenderBuffer& buffer,
 
     bool customColor = found ? findKey(definitionSi, "CustomColors") == "1" : false;
 
+    std::vector<std::pair<std::string, std::string>> sortedStateDefinitions;
+    for (const auto& it : definitionSi) {
+        if (EndsWith(it.first, "-Name")) {
+            sortedStateDefinitions.emplace_back(it);
+        }
+    }
+    std::sort(sortedStateDefinitions.begin(), sortedStateDefinitions.end(),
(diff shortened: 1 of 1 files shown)


keithsw1111:
They render today in the stored order right which is also the displayed order in the dialog. Silently and invisibly reordering them during the render does not sem like a good idea. If you want the ability to reorder make it an option in the dialog.

derwin12:
But they dont Keith .. that is the issue. 
Two items, only difference is 4+5 or 9+10.  
The red of 4 renders before the green of 5 - correct.
The green of 10 renders before the red of 9 - wrong.
image
image
image
Also here is the package of this bug..
[State Bug.zip](https://github.com/user-attachments/files/20001223/State.Bug.zip)

cybercop23:
And the original model worked because all of the "u" states were 11-17
"a" also works because s19 will aways be less than s20.  but in the case of the copy "u" since it skips blanks, s9 and s10, s9 will render last.

computergeek1507:
Sounds like copy shouldnt skip blanks and move the states around

cybercop23:
> Sounds like copy shouldnt skip blanks and move the states around

that's exactly what i'm thinking... but think if you create them you'll always have to be aware not to cross the 1-9 with 1x or 2x for the same named state. Do we skip 1-9 or make them s01-s09? Else we should have a warning that the same state croses the imaginary boundry.

derwin12:
> Sounds like copy shouldnt skip blanks and move the states around

but that wont fix the bug.

derwin12:
Addressed with a new PR.

PR #5357, opened 2025-05-06
Count blanks/zeros as a pixel

cybercop23:
When using placeholders, count them as a node. Ensure last one is also counted, due to wxStringTokenizer.
Standardize Zeros and Blanks
Make Compress use 0s by default - can always rightclick and remove - makes it easier to read - although count if not accounting for them.

image

### DIFF
--- a/xLights/SubModelsDialog.cpp
+++ b/xLights/SubModelsDialog.cpp
@@ -504,9 +504,6 @@ int SubModelsDialog::CountNodesInRange(const wxString& range) {
     wxStringTokenizer tokenizer(range, ",");
     while (tokenizer.HasMoreTokens()) {
         wxString nodeRange = tokenizer.GetNextToken();
-        if (nodeRange.IsEmpty())
-            continue;
-
         if (nodeRange.Contains("-")) {
             int dashPosition = nodeRange.Index('-');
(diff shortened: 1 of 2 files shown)


dkulp:
I'm trying to figure out why blanks would be counted as a node.   The node count is there to explicitly count nodes.   

A third column could likely be added for "Width" or something (Height if Vertical is checked), but Nodes have a very definitive meaning and this confuses it.

cybercop23:
When a SS effect runs across a submodel, it "counts" the node/place-holder. Meaning that if it needs to be on the 4th pixel, it "knows" that blank should count as the 4th one. This is the special case, when you purposely set the submodel to have the same length, but maybe not have all of the pixels, so when you look at the node count, it will look like it is wrong. 
Maybe Width can work, but adding yet another column, I think it will be confusing to the users - but I can add it if that's necessary.

dkulp:
This isn't any different than a "Custom Model".    It may have a width of 50, height of 50, but only 100 "Nodes".  There would be lots of blanks all over the place, but that doesn't change the "Node Count" of the model.  

Again, it's counting Nodes, then it needs to actually count Nodes.

cybercop23:
So just a 3rd column then, Width. I can change it if we're in aggrement.

Issue #4169, opened 2023-11-29
Color picker does not honor pure green


no1unorightnow:
**Describe the bug**
Color picker replaces hex #00FF00 with #008000 on selection.

**To Reproduce**
0. Be using the OS and version specified below - this does not appear to affect the Windows version from my understanding, but I have not tested Windows myself.
1. click on Sequencer tab
2. if none open, click "new sequence" button - any will work (I've attached an example blank sequence)
3. click any square in "Color" sub-window to pick a color
4. set the color to just green (#00FF00), using any of the methods (wheel, hex entry, green to 255, etc.)
5. click the "Select" button
6. observe that it changes the color to #008000 instead (dark green)
7. If you have #00FF00 already in the 3rd square (as is default), you can even just click the bright green square and then the "select" button without anything else, and note that it turns to the dull green.

**Expected behavior**
The color picker should retain the color picked by the end-user.

**Screenshots**
Screenshots attached in attachments section.

**Versions (please complete the following information):**
 - OS: Ubuntu 22.04 and 23.10 tested
 - xLights version 2022.24, 2023.17, and 2023.21 tested on Ubuntu 22.04; 2023.21 tested on Ubuntu 23.10

**Additional context**
I have yet to find a version of the xLights AppImage that does not exhibit this behavior. All other colors I've tested (which granted, is a small subset of the 2^24 possible) work just fine, it's just the full green (#00FF00) that seems to fail.

**Attachments**
[none.zip](https://github.com/smeighan/xLights/files/13495701/none.zip)
![Screenshot from 2023-11-28 20-47-31](https://github.com/smeighan/xLights/assets/5455204/d9f31d80-b783-4c7b-9f85-543b3f4018d0)
![Screenshot from 2023-11-28 21-02-07](https://github.com/smeighan/xLights/assets/5455204/d792921e-c834-4b80-92f8-1eb4fca9a495)

derwin12:
Can you test this again - this seems more like a wxwidgets issue than an xlights issue. Can you just enter the color name?

PR #5368, opened 2025-05-20
Ability to import Color Hex Values

derwin12:
Wanted to use AI to generate palettes but then there was no easy way to add them other than at the file system. With this change there is a prompt to enter the hex string. May also help on #4169 
(Many unrelated changes with the code::blocks generated stuff)

Can enter string like ,, #808080 to add that value as the third swatch or just send it 8 comma separated values.

Aside: I use this in chatgpt..
"xlights color paletes are 8 unique colors. Can you create a color palette that would represent the moods and imagery from the song ""song name"" by ""artist"". Avoid dark, near black colors. Create a React component that shows a preview of the colors in the browser. Also provide the hex values with a trailing comma and no spaces, so I can use them in xlights"

image
image


derwin12:
What if I change the save icon to a menu icon or .... then make that a menu popup. Save, Save As and Import. ?

PR #5370, opened 2025-05-21
This change only affects HinksPix controllers.

jhinklejr:
I did make a change to 2 xlights files: When "Universe per String" was UN-Checked the Number of channels per Universe did NOT go to 510 but remained.  I made a change that would reduce the Universe count and pack them to 510. With "Universe per String" NOT checked, the software for the HinksPix creates a memory re-mapping table and sends it to the controller during the "Upload to Controller" process.


derwin12:
Joe .. the build failed. I think you need to look at the build error and perhaps tweak an include or something.

jhinklejr:
I don't think there is a failure.  I clicked a button that I don't think I should have since I did not have the "privilege" to actually build.  The code I submitted has NO conflicts and should merge ok.  If you look at the failure - it has to do with building - and I am not allowed to do so - so it failed.  Please let me know what you think.  If I just "pull" again and not press that option button - no failure.  Thanks.

PR #5373, opened 2025-05-22
Add optional RGB palette-based coloring to Candle effect.

gstrike:
This PR adds support for optionally using up to two palette colors in the Candle effect.

- RGB values are calculated dynamically based on flame intensity.
- If only one palette color is selected, black is used as the second color.
- The default Candle behavior remains unchanged unless 'Use Palette' is enabled.
- 'Use Palette' is disabled by default to ensure compatibility with existing sequences.

The original Candle effect uses hard-coded colors that cannot be changed. This update introduces greater creative flexibility by allowing users to customize the flame’s color behavior via the palette.

Example of the Candle effect using different color selections:
Xlights_CBLaAjnUt7

### DIFF
--- a/xLights/effects/CandleEffect.cpp
+++ b/xLights/effects/CandleEffect.cpp
@@ -110,7 +110,7 @@ void CandleEffect::SetDefaultParameters()
     SetSliderValue(fp->Slider_Candle_WindVariability, 5);
 
     SetCheckBoxValue(fp->CheckBox_PerNode, false);
-	SetCheckBoxValue(fp->CheckBox_UsePalette, false);
+    SetCheckBoxValue(fp->CheckBox_UsePalette, false);
 }
 
 void CandleEffect::Update(wxByte& flameprime, wxByte& flame, wxByte& wind, size_t windVariability, size_t flameAgility, size_t windCalmness, size_t windBaseline)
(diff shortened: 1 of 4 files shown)


xLights/effects/CandleEffect.cpp
```diff
@@ -160,6 +161,14 @@ void CandleEffect::Render(Effect* effect, const SettingsMap& SettingsMap, Render
     int windVariability = GetValueCurveInt("Candle_WindVariability", 5, SettingsMap, oset, CANDLE_WINDVARIABILITY_MIN, CANDLE_WINDVARIABILITY_MAX, buffer.GetStartTimeMS(), buffer.GetEndTimeMS());
     int windBaseline = GetValueCurveInt("Candle_WindBaseline", 30, SettingsMap, oset, CANDLE_WINDBASELINE_MIN, CANDLE_WINDBASELINE_MAX, buffer.GetStartTimeMS(), buffer.GetEndTimeMS());
     bool perNode = SettingsMap.GetBool("CHECKBOX_PerNode", false);
+	bool usePalette = SettingsMap.GetBool("CHECKBOX_UsePalette", false);
```
derwin12, line 164: Spacing seems to be wrong here?

gstrike:
Indeed.  Should be better now!

keithsw1111 (commented):
All other effects default to white if no colour is selected.

gstrike:
Thanks for pointing that out @keithsw1111.  

For the Candle Effect to work there are two colors that need to be set.  

In the original implementation (when not using the palette) , these colors defaulted to red & yellow and could not be changed.  This PR does not effect this behavior.

However, when using the palette is enabled, there are two cases where the system needs to decide on one or both colors for the user:  

**1. Palette Enabled & no colors selected.**
Initially my PR set both colors to black.  However to @keithsw1111 's point other effects default to white.  I've updated the code to now use white as the first color and black as the second. (https://github.com/xLightsSequencer/xLights/pull/5373/commits/142ee2677a3c72f3815583173841ed904c25a16f) 

**2. Palette Enabled & one color selected.**
In this case we use the palette selected color for the first color, and default the second color to black.  This was the behavior in the initial PR.

Of course, if two colors are selected, we just use the colors they have selected.

Thanks guys!

Issue #5396, opened 2025-06-11
Old Shockwave's turn on Scale To Buffer When Selected


kevinsaucier:
Mac 2025.05 - 2025-04-14 (just tested with the 2025-06-10 release and this still exists)

Admittedly, a relatively small issue, but one that has a decent impact without being apparent.  I noticed tonight while updating one of Tom's sequences (we all know he used to love those shockwaves) that most of the  shockwaves looked great when I first rendered and tested but, after I clicked the effect itself to check settings or mess with it, it got tiny and didn't look right.  From what I can tell/guess, selecting the effect is applying the Scale to Buffer setting as Checked, which is messing up the look.

Attached a test.  The first shockwave on Roof Line Props has already been selected.  The others haven't.

Expectation would be that, since Scale to Buffer didn't exist when this sequence was created, Scale to Buffer would remain unselected when selecting any of the effects.

PR #5397, opened 2025-06-12
Bug with Scaling default option #5396

derwin12:
When an old sequence was loaded, the new scale to buffer was incorrectly being set to true.


computergeek1507:
Did you check Galaxy too?

derwin12:
yes .. I checked the others.

Issue #5398, opened 2025-06-12
2025.06 States not working correctly


csepesy:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
The face effect isn't honoring the states created for the EFL Snowman

**To Reproduce**
Click on the snowman face effect

**Expected behavior**
It should be using the forced colors in states

**Screenshots**

<img width="321" alt="Image" src="https://github.com/user-attachments/assets/1967d3bc-8a62-4775-b222-fbaf90e1883a" />

**Versions (please complete the following information):**
 - OS: Sequoia 15.3.2
 - xLights version [e.g. 2025.06]

**Additional context**
n/a

**Attachments**

[EFL Snowman State Effects.zip](https://github.com/user-attachments/files/20700777/EFL.Snowman.State.Effects.zip)

kylegrymonprez:
Was in the zoom room and someone also reported this - checked .05 and it worked fine with states as outline. Upgraded to .06 and state as outline would not render in the UI.

derwin12:
Addressed in PR.

PR #5399, opened 2025-06-12
Correct the state as outline  in Face Effect #5398

derwin12:
With the change to correct the render order, missed this code to account for the state names/etc now being 0 padded. #5398


derwin12:
Basically the items in the map are s1,s10,s11,s2  and now they are stored as s001,s002,s010,s011.
So that they will render in the proper order .. just as they appear in the list. It is not just a compare taking place, it is storing the contents in the map. It needs to be stored as s001 s001-name and s001-color so that any operations take place in the right order.

computergeek1507:
does that mean existing layouts have to open the dialog and/or resave for the keys to update to the new format?

derwin12:
the stored values are still s1,s2, etc only internally when loaded do they get zero padded so that they sort properly.

Issue #5403, opened 2025-06-14
Node Level Skipping At End Of Screen When Arrow Down


ChooChooMan74:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

THIS IS A ZOOM ROOM REPORTED BUG

**Describe the bug**
Issues scrolling the on effect up and down nodes.

**To Reproduce**
Open up Node Level
Put an on effect.
Arrow the effect down
When it gets to the bottom of the viewable nodes, it starts to act funny with a pattern.
Say the bottom node is Node 10.  Arrow down once, it goes to node 11 (+1 node).  Arrow again, it goes to node 13 (+2 node), arrow down again, goes to node 16 (+3). Then, scroll to get the on effect towards the top of the viewable screen, and up arrow and the effect actually moved down on the first up.  Then it will move up.

When scrolling up, it will go to the top visible node and not scroll up any further.

**Expected behavior**
Go down one node at a time.  Goes up one node at a time.

**Screenshots**

**Versions (please complete the following information):**
Win 11, XL 2025.06

**Additional context**

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

derwin12:
Just a FYI that is has been this way for a couple of years in my quick testing.

derwin12:
This seems to be intentional - ie allows you to quickly scroll down a long list.  Thoughts?

ChooChooMan74:
Sounds good to me. I was just reporting what JoelD was reporting.  Let the devs figure it out. I am not smart enough for that.

PR #5411, opened 2025-06-25
Handle effect movement #5403

derwin12:
The code used for handle a single effect movement would did not account for expanding strands, timing marks, nodes etc. The range selection is cleaner and works correctly even with one effect selected. No apparent performance issue found. #5403


derwin12:
What I found was selecting effects wasnt working - I had a timing track selected - so it is doing what is expected in that scenario.

Issue #5416, opened 2025-06-27
Hinkspix V2 Crash on visualizer


derwin12:
From zoom room - when clicking universe per string or trying to add more props. 
Short term fix was to disable autosize and add a bunch of universes.

To Duplicate - just added the 8 port MT .. cascade over 2 remotes. Was getting a 2x4 pattern that would exceed the pixel count. Tried click universe per string and it crashed.

[xLights_dbgrpt-17432-20250626T163127.zip](https://github.com/user-attachments/files/20937047/xLights_dbgrpt-17432-20250626T163127.zip)

derwin12:
Thank .. I was indeed looking at the commit trying to figure out what it was doing.

derwin12:
The commit code was not the issue .. going to close this as just a really bad config - he places a Virtual Matrix on the hinks which cant handle all those universes (per string). Also the mh shadow models were pointing to crud. I will ping him on messenger. He basically deleted his Christmas layout and tried to make Halloween and left things in a weird state.
It works just find if I remove just the one Pixel2DMX (MH1/MH2) misconfiguration.

computergeek1507:
We should at least prevent a crash, user error should never crash xLights

derwin12:
Agree but it was over my head.. reopen if you think someone would work on
it.

derwin12:
I put in code to handle it .. Scott can you take a look.

PR #5418, opened 2025-06-27
Handle the misconfigured outputs #5416

derwin12:
This logs the error when visualizing with too many outputs configured #5416


derwin12:
Not really .. I had noticed it was used before and it seemed to work. I can add the usual logger code in there instead.

derwin12:
Switched it over the logger_base code..

Issue #3617, opened 2022-12-06
Win xScheduleWeb Interface Tweak Request


TripleJ3EB:
The  xScheduleWeb/index.html page, when viewed on a mobile device, it stacks the Controller Status above the Available Playlists or Song List if playing a list.

If you have a lot of controllers, like the 50+ I have in our show, thats a lot of scrolling to get to the playlist or song selection part of the page.  Would appreciate it if it reversed those two.

Also, I believe the repeat one step is missing, you can just repeat the entire playlist.  

Thanks

PR #5443, opened 2025-07-13
Make the controller status section collapsable #3617

derwin12:
Rather than mess with order, this change simply makes the controller section collapsable. When using a narrow device you can reduce the area for that section. #3617 

image


derwin12:
@keithsw1111  Keith, are you okay with this change?

Issue #5446, opened 2025-07-14
remove "-shimmer" function.


Clyde-L:
You guys are really great at adding nice touches to the menu, thank you!  But I'm a bone head and accidentally selected the Add "-shimmer" to the wrong phoneme. LOL can another function be added to right-click menu to "remove -shimmer"?  I know, I know, first world problems haha. 

As always, thank you for looking!
Clyde L

<img width="911" height="332" alt="Image" src="https://github.com/user-attachments/assets/a2a37a3c-6398-4f46-81d1-8c33292e75b6" />

Clyde-L:
> Bottom of the queue this goes :-)

LOL Yep, I knew it haha.  Thanks!

PR #5467, opened 2025-07-31
Add right click to remove -shimmer #5446

derwin12:
Seems folks want to remove -shimmer as much as they want to add -shimmer #5446 
image


computergeek1507:
The utils class has inline method you can use as well

derwin12:
Switched it to the c++ code .. couldn't use the contains as I need to find it confidently at the end of the string.

computergeek1507:
You could make it smarter, only show add if it doesn't have it, and remove if it does

computergeek1507:
Does it allow double add at the moment?

derwin12:
It doesn't allow the double add - and I didn't want slow down the right click  to scan all the select effects to see if it needed an add or delete. So purposely left it out.

Issue #5501, opened 2025-08-25
Filter out the metadata mac files from Batch Render


derwin12:
If you store your files on non Mac FS (exFat) it creates these ._ files for meta data. No need to show them when doing a batch render.

<img width="763" height="261" alt="Image" src="https://github.com/user-attachments/assets/daa3f7c7-1ba4-408e-967d-19bec19f3aab" />

derwin12:
Testflight shows that this change didnt work.

PR #5527, opened 2025-09-15
Exclude the mac metadata files from the batch render picklist #5501

derwin12:
Found that MacOS creates meta files if you are not using Apple FS - so exclude those on the batch render pic list. #5501


computergeek1507:
the "wxDIR_DEFAULT" enum value includes hidden files, try passing "wxDIR_FILES | wxDIR_DIRS" instead to fix it.
 
image

Issue #5515, opened 2025-09-05
Orientation metadata tag is ignored in background images causing some images to be displayed upside down.


bl4krat:
Orientation metadata tag is ignored in background images causing some images to be displayed upside down.

**To Reproduce**
1) Take an image of your house with a smart phone (iphone in my case)
2) verify the image displays the right way up on the phone
3) email it to yourself
4) on your PC, save the attachment and verify the OS displays it the right way up
5) In xLights, start with a new show. Go to the LAYOUT tab
6) for "Background image", click on "..." and navigate to the image
7) click "Open"

The background image will appear, but depending on the orientation of the phone at the time the image was taken, the image may appear upside down (rotated through 180 degrees).

**Expected behavior**
The image should be displayed the right way up. The "Orientation" metadata field contains information on how to achieve this (for example a value of 1 means no rotation is required, a value of 3 means  'Rotate through 180 degrees')  which xLights is ignoring.

**Versions (please complete the following information):**
- OS - Win 11
 - xLights version 2025.8 64bit

**Additional context**
This issue has been known for several years and often dismissed as images themselves not being correctly rotated. The workaround is for the user to have to manually rotate the image outside of xLights.

derwin12:
if easy enough - can you attach the photo for testing.

bl4krat:
![Image](https://github.com/user-attachments/assets/f3ac1c3c-69de-468f-bdac-a06b223031dc)

bl4krat:
Try this one as an example.

PR #5544, opened 2025-09-24
Use the orientation metadata tag for jpg images #5515

derwin12:
Jpg files can have a rotation in the metadata, if found, honor that setting in picture effects and background images #5515

### DIFF
--- a/xLights/effects/PicturesEffect.cpp
+++ b/xLights/effects/PicturesEffect.cpp
@@ -240,7 +240,7 @@ typedef std::vector< std::pair<wxPoint, xlColor> > PixelVector;
 
 class PicturesRenderCache : public EffectRenderCache {
 public:
-    PicturesRenderCache() : imageCount(0), frame(0),  gifImage(nullptr), maxmovieframes(0) {};
+    PicturesRenderCache() : imageCount(0), frame(0),  gifImage(nullptr), maxmovieframes(0), orientation(1) {};
     virtual ~PicturesRenderCache()
     {
         if (gifImage != nullptr) {
@@ -255,6 +255,7 @@ public:
     int frame;
     int maxmovieframes;
     wxString PictureName;
+    int orientation;
     GIFImage* gifImage;
     std::vector<PixelVector> PixelsByFrame;
 };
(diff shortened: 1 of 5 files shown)


computergeek1507:
I think this is a good idea, but I'm a little nervous this will rotate pictures that have this flag and people didn't know it, especially in the picture effect, everything could all the sudden upside down

xLights/effects/PicturesEffect.cpp
```diff
@@ -655,10 +660,12 @@ void PicturesEffect::Render(RenderBuffer& buffer,
 
             if (loopGIF) {
                 image = gifImage->GetFrameForTime((buffer.curPeriod - buffer.curEffStartPer) * buffer.frameTimeInMs * frameRateAdj, true);
+                image = ApplyOrientation(image, cache->orientation);
             }
             else {
                 int ii = cache->imageCount * buffer.GetEffectTimeIntervalPosition(frameRateAdj) * 0.99;
                 image = gifImage->GetFrame(ii);
+                image = ApplyOrientation(image, cache->orientation);
```
computergeek1507, line 668: does the gif file format support this rotation meta data? maybe just wasting CPU cycles checking it on GIF frames

xLights/effects/PicturesEffect.cpp
```diff
@@ -631,16 +633,19 @@ void PicturesEffect::Render(RenderBuffer& buffer,
                             logger_base.error("Error loading image file: %s.", (const char*)NewPictureName.c_str());
                             image.Create(5, 5, true);
                         }
+                        image = ApplyOrientation(image, cache->orientation);
                         rawimage = image;
                     } else {
                         image = gifImage->GetFrame(0);
+                        image = ApplyOrientation(image, cache->orientation);
```
computergeek1507, line 640: GIF too

xLights/effects/PicturesEffect.cpp
```diff
@@ -631,16 +633,19 @@ void PicturesEffect::Render(RenderBuffer& buffer,
                             logger_base.error("Error loading image file: %s.", (const char*)NewPictureName.c_str());
                             image.Create(5, 5, true);
                         }
+                        image = ApplyOrientation(image, cache->orientation);
```
computergeek1507, line 636: remove too?

derwin12: Perhaps I am missing something .. how do we know it is not a jpg? I think the code calls everything a gif :-)

PR #5557, opened 2025-10-01
Refactored EspixelStick support.

MartinMueller2003:
Corrected color handling
Corrected how Gamma is handled in json
Refactored the json handling on the output stage

### DIFF
--- a/xLights/controllers/ESPixelStick.cpp
+++ b/xLights/controllers/ESPixelStick.cpp
@@ -23,6 +23,30 @@
 
 #include <log4cpp/Category.hh>
 
+#pragma region Global Data
+static std::map<wxString, wxString> EspsV4ColorOrders = {
+    { "rgb nodes",  "rgb" },
+    { "rbg nodes",  "rbg" },
+    { "gbr nodes",  "gbr" },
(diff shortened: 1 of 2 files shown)


dkulp:
That definitely seems completely wrong to me... not merging.    If it's set at the model node type level, then yes, xLights will arrange the channels.   If it's a controller connection level thing like pretty much every controller commonly used, then xLights will not.

MartinMueller2003:
What about port level? Espsv4 supports up to 8 ports and each one can have its own color order.

MartinMueller2003:
While I agree with you, the outer model seems to be the only place you can get the information. If you uncheck the port color order on the UI, the call returns the passed in default value. If you check the port color order on the UI, the call returns the color order selected on the UI. It seems to be returning the correct data. FYI: This is the way it works in the current "main".

To get to the string color order, I needed to get the model within the model. That structure returns a color order that follows the string attributes setting.

MartinMueller2003:
Hi I had gotten into a few other issues. I would like to have a conversation about why you think this is completely wrong. It is using the data that the other controllers are using.

Issue #5566, opened 2025-10-05
Upload Output causes instant crash on 2025.10


Unit-108:
**Describe the bug**
After updating to 2025.10, Upload Output to my F16V4 with 16 port expansion board, xLights seems to just instantly crash without any dialogue windows or crash dumps that I can see. I do see the bottom left status text for Uploading, then xLights crashes when that popup window with upload progress is supposed to appear.

I can Upload Output to my F16V3 without expansion just fine. I have tested this on two Windows 11 systems running 2025.10 and it happens on both.

I tested with different show directories and setup's for the controller to use, same outcome. It crashes, and does not upload, my controller has the same config as it did before I tried to Upload Output. No sequences were open during these attempts.

I was able to upload output by installing 2025.09 and it worked, but when I try the same again on 2025.10 it crashes.

**To Reproduce**
Controller configured as following:
Vendor: Falcon
Model: F16V4
Variant: 16 + 16 Local Ports
Active: Active
Monitor: True
IP Address: Set to a local IP
Protocol: E131
Priority: 100
Models: A lot of them
Start Universe: 1
Universe Count: 64
Universes: 1-64
Channels per Universe: 510
Managed: True

1. Select the controller
2. Click Upload Output

**Expected behavior**
I would expect a small dialogue window to appear followed by progress of upload and upload completion. I would then expect my controller to have the configuration I uploaded to it.

**Screenshots**
<img width="1488" height="1099" alt="Image" src="https://github.com/user-attachments/assets/2495636a-10a2-4d92-9c68-ac7a228cfad4" />

**Versions (please complete the following information):**
 - OS: Windows 11 24H2 & 25H2
 - xLights version: 2025.10
 - F16V4 Build: 26

**Additional context**
Worked on 2025.09, but not 2025.10. My setup is a bit complex but I hope the log files are somewhat helpful.

I am also using DMX on this controller to control flood lights, might be helpful to know or not.

**Attachments**
I have included my logs from the 24H2 system, screenshot, and what should be a copy of my setup. I may not have provided it in the best way but you can load it up and see my layout and controllers.
I also have included the settings json for the controller, it has the layout I uploaded using 2025.09, unsure if needed but might be helpful.
(Controller named Falcon32 is the problematic one)

[...]

PR #5574, opened 2025-10-07
Fix Upload Output for FPP

zackbcom:
Fixes an issue https://github.com/xLightsSequencer/xLights/issues/5566#issuecomment-3375269394

I only fixed the non safe json access for my issue, Probably want to refactor all the calls to `.get` that replaced `.AsString` and the others.

# Fix JSON Access Safety Issues in FPP.cpp

## Problem Description

The `FPP.cpp` file contained unsafe JSON field access patterns that would cause crashes when processing JSON payloads with missing fields. This was particularly problematic for certain channel output types like `HTTPVirtualDisplay` that don't include a `device` field.

### Root Cause

During the migration from `wxJSONValue` to `nlohmann::json`, the safe `AsString()` method was replaced with the unsafe `.get<std::string>()` method:

**Before (wxJSONValue - safe):**
```cpp
wxString device = otherDmxData["channelOutputs"][x]["device"].AsString();
```

**After Migration (nlohmann::json - unsafe):**
```cpp
std::string device = otherDmxData["channelOutputs"][x]["device"].get<std::string>();
```

### The Issue

The `nlohmann::json::get<T>()` method throws an exception if the requested key doesn't exist, while `wxJSONValue::AsString()` handled missing keys gracefully by returning empty strings.

### Example Problematic Payload

```json
{
    "enabled": 0,
    "type": "HTTPVirtualDisplay",
    "startChannel": 1,
    "channelCount": 8388608,
    "width": 1280,
    "height": 1024,
    "pixelSize": 2
}
```

This payload lacks a `device` field, causing the application to crash when the code tried to access `["device"].get<std::string>()`.

### Safe JSON Access Pattern

Replaced all unsafe direct JSON access with the existing safe utility functions for my issue:

**Correct Migration (nlohmann::json - safe):**
```cpp
std::string device = GetJSONStringValue(otherDmxData["channelOutputs"][x], "device");
```

### Utility Function Used

The codebase already contained a safe accessor function:

```cpp
static std::string GetJSONStringValue(const nlohmann::json& val, const std::string &key, const std::string &def = "") {
    if (val.contains(key) && val[key].is_string()) {
        return val[key].get<std::string>();
    }
    return def;
}
```

This function:
- Checks if the key exists using `contains()`
- Validates the value is a string using `is_string()`
- Returns the actual value or a default (empty string) if missing/invalid

### DIFF
--- a/xLights/controllers/FPP.cpp
+++ b/xLights/controllers/FPP.cpp
@@ -2051,7 +2051,7 @@ static bool mergeSerialInto(nlohmann::json &otherDmxData, nlohmann::json &otherO
             if (!device.empty() && !origDevice.empty() && origDevice == device) {
                 //same device, see if type matches and update or disable
                 std::string origType = GetJSONStringValue(otherOrigRoot["channelOutputs"][y], "type");
-                if (type == origType) {
+                if (!type.empty() && !origType.empty() && origType == type) {
                     //device and type the same, update values
                     found = true;
                     changed |= UpdateJSONValue(otherOrigRoot["channelOutputs"][y], "description", GetJSONStringValue(otherDmxData["channelOutputs"][x], "description"));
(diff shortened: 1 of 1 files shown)


computergeek1507:
nlohmann has this functionality with the "value" method. We could and probably should use it by default. I normally avoid it as you have to be more intentional with types and type conversions. 

```
    int v_integer = j.value("integer", 0);
    double v_floating = j.value("floating", 47.11);
    std::string v_string = j.value("nonexisting", "oops");
    bool v_boolean = j.value("nonexisting", false);
```

https://json.nlohmann.me/api/basic_json/value/

Issue #5602, opened 2025-10-20
Long directory names on batch render overlap


cf-mp:
**Describe the bug**
I work on a Mac but I think this was the same case on my windows 10 PC a few years ago before I switched.  I store my sequences in /user/my name/documents/xlights/Holday 2025 v1-1/sequences directory.  When I batch render sequences, the input file name of the sequence and the output file names are too long and it overlaps in the middle making it hard to read what file it is actually on.  This is just a cosmetic error and in no way causes problems besides it being hard to tell what file it is currently rendering.

**To Reproduce**
Put your sequences in a slightly deep folder structure and run batch render.  See attached screen shot to see how the input and output file names overlap each other on the screen.

**Expected behavior**
I am not sure the best way to fix this.  Maybe remove the output folder name for the fseq so the overall line is short enough to fit on an average resolution monitor or have it put the input xsq file on one line and the output on a second line.  

**Screenshots**

<img width="1505" height="82" alt="Image" src="https://github.com/user-attachments/assets/e34e2a1b-0b3b-401b-b8c4-3048ed708ca7" />

**Versions (please complete the following information):**
 - OS: MacBook Pro M3 Max running Tahoe 26.0.1
 - xLights version 2025.10.1

PR #5607, opened 2025-10-21
Truncate the batch render status to the last 100 chars #5602

derwin12:
Just show the last 100 characters in the status line. #5602


cybercop23:
I'd say the latter, anything from the show folder and below.

derwin12:
Oops...

PR #5624, opened 2025-10-30
Add SM16825 RGBCW LED support for WLED controllers

jjarmoc:
## Summary
Adds support for SM16825 RGBCW (5-channel) LEDs to WLED controller configurations in xLights.

## Changes
- Added `sm16825` protocol to the list of supported pixel protocols in `wled.xcontroller`
- Added protocol encoding mapping (sm16825 → type 34) in `WLED.cpp` `EncodeStringPortProtocol` function

## Motivation
The SM16825 chip is becoming increasingly common in LED products (such as Govee Outdoor Permanent Lights Pro H706A). This change enables xLights to properly configure WLED controllers to use SM16825 LEDs.

## Technical Details
- **Protocol type**: 34 (TYPE_SM16825 in WLED firmware)
- **Characteristics**: RGBCW (RGB + Warm White + Cold White), 16-bit
- **Compatible with**: WLED firmware 0.15.0-b5 and later

## Test Plan
- [x] Verify sm16825 appears in LED protocol dropdown for WLED controllers
- [ ] Verify xLights correctly configures WLED controller with sm16825 protocol
- [ ] Verify WLED controller properly controls SM16825 LEDs after upload

### DIFF
--- a/controllers/wled.xcontroller
+++ b/controllers/wled.xcontroller
@@ -21,6 +21,7 @@
             <Protocol>tm1829</Protocol>
             <Protocol>ucs8903</Protocol>
             <Protocol>ucs8904</Protocol>
+            <Protocol>sm16825</Protocol>
             <Protocol>ws2801</Protocol>
             <Protocol>apa102</Protocol>
             <Protocol>lpd8806</Protocol>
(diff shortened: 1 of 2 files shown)


computergeek1507:
This will not work, the issue with WLED that makes it different from Falcon/Kulp, it expects the data over E131/DDP as three channel RGB and then it converts it to the 5 channel RGBWW. If you enable SM16825 in xLights, xLights will send 5 channels of data causing everything to be wrong in WLED. Falcons/Kulps don't do this conversion, they take 5 channels in and push the five channels out the pixel output. Anything that isn't 3 channels is normally a problem. The WLED E131 does have a setting to accept 4 channel input, but this is global and can cause mixture of LEDs protocols to be wrong. As stated on Facebook it's better to manually setup WLED and xLights when using non 3 channel LEDs.

derwin12:
It seems this is noted as not being safe nor a working solution

PR #5671, opened 2025-11-17
Fix accidental effect/timing shift on click

o0charlie0o:
When clicking on effects or timing marks in the sequencer, even tiny mouse movements (1-2 pixels) would trigger the drag/resize logic, causing unwanted shifts in effect timing. Added a 3-pixel drag threshold that distinguishes between intentional drags and simple clicks, preventing accidental timing changes while keeping edge resizing responsive.

Before:

https://github.com/user-attachments/assets/54941d20-6834-4c62-9f92-6f9fd1e61118

After:

https://github.com/user-attachments/assets/2fb5352e-f157-45da-b560-070940f1f88a


o0charlie0o:
@cybercop23 yeah me as well. I guess this approach just removes that extra step of having to lock and unlock every time. I don't personally lock effects ever.

PR #5711, opened 2025-12-01
Update manifest.json to fix icons paths

jasoveen:
Fixed xScheduler icon paths in the manifest.json to resolve issue on Android Firefox. Since the images are in the same folder as the manifest.json we don't need any path info. Also changed the "name" and added the "short_name" and "display" values since those are usually required in the manifest.json.

This fixes an issue on Android when using Firefox as the browser where the xScheduler icon was not used when you added the site to your "Home Screen".

xScheduler Home Icons


derwin12:
Moved over the xschedule repo and commited.

PR #5712, opened 2025-12-01
Update xSchedule webSocket.js

jasoveen:
Added support for connecting to secure WebSocket (wss) if browser is using a https connection. Useful when a secure reverse proxy is in front of xSchedule.


derwin12:
Moved this over to the other repo and will submit there.

PR #5748, opened 2025-12-16
Fix #5707: Improve group text contrast in Import Effects dark mode

### LINKED ISSUE #5707: Low contrast on 'Import Effects' screen when in Dark Mode
**Describe the bug**
The 'Import Effects' screen displays groups in dark blue text. When running in Dark Mode on OSX, this leads to very low contrast which impacts usability.

**To Reproduce**
1. Enable Dark Mode on OSX
2. Launch Xlights and load a sequence
3. Select 'Import Effects' and point to a sequence.
4. Note that groups are in dark blue text on a black background.

**Expected behavior**
Groups are in a higher contrast color and thus are more legible.

**Screenshots**
Image

 - OS: OSX Tahoe 26.1 (25B78)
 - xLights version 2025.12

rshambaugh:
## Summary
- Groups in the Import Effects dialog were displayed in dark blue text on a dark background in macOS dark mode
- Changed from `CyanOrBlueOverride()` to `CyanOrBlue()` which directly checks dark mode and returns cyan for better contrast

## Root Cause
`CyanOrBlueOverride()` checks if `COLOR_TEXT_HIGHLIGHTED` is set to black before falling back to `CyanOrBlue()`. This additional logic could result in a less visible color. Using `CyanOrBlue()` directly ensures proper dark mode handling.

## Test Plan
- [x] Build and run xLights on macOS in dark mode
- [x] Open File → Import → Import Effects
- [x] Select a sequence file with model groups
- [x] Verify groups now display in cyan (readable) instead of dark blue

### DIFF
--- a/xLights/xLightsImportChannelMapDialog.cpp
+++ b/xLights/xLightsImportChannelMapDialog.cpp
@@ -189,7 +189,7 @@ bool xLightsImportTreeModel::GetAttr(const wxDataViewItem &item, unsigned int co
     }
 
     if (node->IsGroup()) {
-        attr.SetColour(ColorManager::instance()->CyanOrBlueOverride());
+        attr.SetColour(CyanOrBlue());
         set = true;
     }
 
@@ -1151,7 +1151,7 @@ void xLightsImportChannelMapDialog::PopulateAvailable(bool ccr)
 
             // If importing from xsqPkg flag known groups by color like is currently done in mapped list
             if (m->type == "ModelGroup") {
-                ListCtrl_Available->SetItemTextColour(j, ColorManager::instance()->CyanOrBlueOverride());
+                ListCtrl_Available->SetItemTextColour(j, CyanOrBlue());
             }
             j++;
         }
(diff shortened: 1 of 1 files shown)


rshambaugh:
Thanks for the feedback. I want to make sure I understand the issue correctly.

Looking at CyanOrBlueOverride():
- If COLOR_TEXT_HIGHLIGHTED is black (the default), it calls CyanOrBlue() which returns cyan in dark mode
- If the user has customized COLOR_TEXT_HIGHLIGHTED, it returns their custom color

The reporter is seeing dark blue in dark mode. If the default path were working, they'd see cyan. So either:
1. IsDarkMode() is returning false on macOS Tahoe
2. The user has COLOR_TEXT_HIGHLIGHTED set to a custom (blue) value
3. Something else is going on

What's the right approach here? Should I:
- Investigate why the dark mode detection might be failing?
- Make CyanOrBlueOverride() apply dark mode adjustments even to custom colors?
- Something else?

PR #5765, opened 2025-12-23
Fix #5323: Meteor effect speed=1 now works at high frame rates

### LINKED ISSUE #5323: Meteor Effect - Speed of 1 with 40fps does not work
It works fine at 20fps. Ziggy saw this when he converted his working 20fps sequence to 40fps.

The meteors don't fall :-)

Can be duplicated by just dropping a meteor effect with speed of 1 on a 40fps sequence.

Image

rshambaugh:
## Summary

The Meteor effect with speed=1 would freeze at frame rates above 20fps (e.g., 40fps, 100fps) because the speed calculation resulted in 0 due to integer division.

Fixes #5323

## Root Cause

The speed calculation `mSpeed * buffer.frameTimeInMs / 50` uses integer math:
- At 20fps (50ms): `1 * 50 / 50 = 1` ✓
- At 40fps (25ms): `1 * 25 / 50 = 0` ✗
- At 100fps (10ms): `1 * 10 / 50 = 0` ✗

When the result is 0, `effectState` never increments and meteors don't move.

## Changes

Applied `std::max(..., 1)` to ensure the speed increment is at least 1, allowing the effect to work at any frame rate.

## Testing

- [x] Built and verified on macOS
- [x] Meteor effect with speed=1 now animates at 40fps and higher frame rates

### DIFF
--- a/xLights/effects/MeteorsEffect.cpp
+++ b/xLights/effects/MeteorsEffect.cpp
@@ -197,7 +197,7 @@ void MeteorsEffect::Render(Effect *effect, const SettingsMap &SettingsMap, Rende
         buffer.needToInit = false;
         cache->meteors.clear();
         cache->meteorsRadial.clear();
-        cache->effectState = mSpeed * buffer.frameTimeInMs / 50;
+        cache->effectState = std::max(mSpeed * buffer.frameTimeInMs / 50, 1);
     }
 
     switch (MeteorsEffect) {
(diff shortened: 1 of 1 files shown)


rshambaugh:
Closing this PR as the issue has been fixed in a better way by @dkulp in commit 85e91e4c4 - using float for effectState instead of the std::max approach. Thanks!

PR #5773, opened 2026-01-02
Update WLED/Bong69 to reflect support for new PCB v4.

gstrike:
This PR updates the device identification string to reflect support for the new WLED Bong69 PCB v4 board, which is backward compatible with PCB v2 and PCB v3.  

The PR updates the displayed board name / identifier only.

### DIFF
--- a/controllers/wled.xcontroller
+++ b/controllers/wled.xcontroller
@@ -198,7 +198,7 @@
             <Port7>5</Port7>
             <Port8>33</Port8>
         </Variant>
-        <Variant Name="8-Port Distro (PCB v2/v3)" Base="WLED:ESP32WLEDSettings">
+        <Variant Name="8-Port Distro (PCB v2/v3/v4)" Base="WLED:ESP32WLEDSettings">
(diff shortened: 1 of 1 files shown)


computergeek1507:
The Variant Name is used to determine what controller is used in Controller tab. If you change it, any user with it selected will revert to the first variant (v1 in this case) as the name stored in the RGB effects file does not match the controller file.

gstrike:
I didn’t realize the Variant Name was used that way, thanks for calling that out.  Given that, I can update the PR so it doesn't change the the existing variant name at all.

1. Keep the "8-Port Distro (PCB v2/v3)" name
2. Add a "8-Port Distro (PCB v4)" name with the same pin mappings.

If you'd prefer a different naming scheme for the variant, I'm happy to adjust.

This keeps existing users stable and also makes it clear and obvious that PCB v4 is explicitly supported.

derwin12:
Going to close this as it is dated and seems like another solution is suggested/implemented.

PR #5780, opened 2026-01-07
Add AI Image Generation

derwin12:
Add Gemini as a service pointing to the image model. Add Generate AI Image item to the menu.
Gemini costs 2-4cents per image but they do offer $300 90day trial as well.
Pulls the ai prompt from the txt file that is stored on your installation - similar to what was done with AI Auto map.
Ideally the user can place a very simple item in the prompt "A Christmas Tree" - and the ai prompt will do the rest.
Note that code was added to remove the black background since Gemini cannot create transparent images.
The Directory is stored as a sys variable so folks are encouraged to create their own Images directory for these images.
Also I have ifdef code in there so folks can test the dialog etc with a local image, without having the gemini api paid service.
Gemini creates a 1Kx1K image by default - cant make it smaller.
**UNTESTED ON MAC**
image

### DIFF
--- a/xLights/ai/AIImageDialog.cpp
+++ b/xLights/ai/AIImageDialog.cpp
@@ -1,0 +1,547 @@
+#include "AIImageDialog.h"
+
+//(*InternalHeaders(AIImageDialog)
+#include <wx/string.h>
+//*)
+#include <wx/mstream.h>
+
+//#define BYPASS_GEMINI // set this to use a local image for testing local functions
+
+//(*IdInit(AIImageDialog)
+const wxWindowID AIImageDialog::ID_STATICTEXT1 = wxNewId();
+const wxWindowID AIImageDialog::ID_PROMPT = wxNewId();
+const wxWindowID AIImageDialog::ID_STATICTEXT3 = wxNewId();
+const wxWindowID AIImageDialog::ID_DIRBUTTON = wxNewId();
+const wxWindowID AIImageDialog::ID_DIRECTORY = wxNewId();
+const wxWindowID AIImageDialog::ID_STATICTEXT6 = wxNewId();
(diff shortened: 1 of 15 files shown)


xLights/xLightsMain.cpp
```diff
@@ -10746,3 +10752,25 @@ void xLightsFrame::OnMenuItemFindShowFolderSelected(wxCommandEvent& event)
 aiBase* xLightsFrame::GetAIService(aiType::TYPE serviceType) {
     return _serviceManager->findService(serviceType);
 }
+
+void xLightsFrame::OnMenuItem_GenerateAIImageSelected(wxCommandEvent& event) {
```
computergeek1507: I would make a new aiType called IMAGEGENERATION then use the GetAIService method above to pass the aibase class into the dialog. This would make it more polymorphic so other ai classes could implement it.

xLights/ai/AIImageDialog.cpp
```diff
@@ -0,0 +1,547 @@
...
+    // === NORMAL MODE: Call Gemini API ===
+
+    prompt = AIPrompt + " " + prompt;
+
+    ResultHTMLCtrl->SetPage("<html><body><i>Generating image with Gemini...<br>This may take 10-30 seconds.</i></body></html>");
+    ResultHTMLCtrl->Refresh();
+    wxYield();
+
+    std::string model = "gemini-2.5-flash-image";
+    model = _serviceManager->getServiceSetting("GeminiModel", model);
```
computergeek1507: In conjunction with the other change, make a virtual method in the base AI class to return the image and other data as a struct or tuple, them move the image generation and retrieval code to a override for the gemini child class.

Issue #5737, opened 2025-12-13
Changing FPS causes render glitches


Mat-Moo:
**Describe the bug**
AS I mostly had WiFi device setup I've been using 25fps for a lot of my sequences, however this year I've moved almost half my props to wired, so thought I'd look at switching up to 40fps. All runs ok, but the rendering at the new frame right is wrong, and I have to go to every effect and click around to get it to update correctly.

**To Reproduce**
https://github.com/user-attachments/assets/941687fa-4c75-4750-874a-9a6e94010c0f
You can see on my mega tree a chase effect, and after switching and re-rendering it's only shows on the second part of the effect, but it still plays the whole effect correctly but glitches at the end. Changing to skips and back, fixes it

**Expected behaviour**
After switching from 25fps to 40fps a re-render should work as expected :)

**Versions (please complete the following information):**
 - OS: Windows 11
 - xLights 2025.13

**Additional context**
Hope the video shows enough

PR #5795, opened 2026-01-18
Add warning to clear render cache if FPS is modified. #5737

derwin12:
if render cache is enabled the effects may not show the expected results without clearing the render cache. Add the warning to inform the user. #5737

### DIFF
--- a/xLights/RenderCache.cpp
+++ b/xLights/RenderCache.cpp
@@ -18,6 +18,7 @@
 #include <wx/filename.h>
 #include <wx/dir.h>
 #include <functional>
+#include "xLightsMain.h"
 #include "xLightsVersion.h"
 #include "UtilFunctions.h"
 #include "TraceLog.h"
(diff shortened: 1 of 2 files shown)


computergeek1507:
Why not automatically clear the render cache and not warn?

derwin12:
I could, I considered it .. should I just do it?

derwin12:
Code updated.

derwin12:
Would that mean opening each piece of the cache, checking the fps in there, then discard it if not the right fps?
Every sequence could be a different fps.
```
                            if (effectObj != nullptr && reff->SupportsRenderCache(SettingsMap) && _renderCache.IsEnabled()) {
                                if (!effectObj->GetFrame(*rb, _renderCache)) {
                                    reff->Render(effectObj, SettingsMap, *rb);
                                    GPURenderUtils::waitForRenderCompletion(rb);
                                    effectObj->AddFrame(*rb, _renderCache);
                                }
                            }
```

PR #5808, opened 2026-02-05
Add palette size preference to Color Manager settings

MrPierreB:
I recently got asked if we can make the color squares a bit bigger - especially in the dropdown of the saved palettes.

I added an Color Panel section to the Colors tab in preferences with a standard or large option.

Here is an image of how it looks normally ,  I have called it standard under the options, then 2nd image is the preferences screen, and the final one is how it looks once you have it set to large.  (I kept the ratios,  I would actually prefer the color palettes to be square block :) - easy to change if that is what we agree on)

@dkulp - First time being in this part of the code, so I would appreciate if somebody can test it out (especially on mac as I don't have one)

ColorPanelSize

### DIFF
--- a/xLights/ColorPanel.cpp
+++ b/xLights/ColorPanel.cpp
@@ -109,7 +109,7 @@ const wxWindowID ColorPanel::ID_MNU_GENERATE = wxNewId();
 #define PALETTE_CC_SIZE_LARGE 18
 
 static bool IsLargePalette() {
-    return wxConfigBase::Get()->ReadLong("PaletteSizeIndex", 0) == 1;
+    return wxConfigBase::Get()->Read("PaletteSize", "Normal") == "Large";
 }
 
 class ColourList : public wxOwnerDrawnComboBox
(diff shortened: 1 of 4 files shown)


MrPierreB:
> Might be easier to place this setting over on the Effects Grid or View settings .. these colors are not really the same thing.

Fair :)
I wouldn't say easier, but better place indeed is on the View Tab.
I have now moved it over to the View Tab,  please have a test and play and let me know how it goes.
image

xLights/UtilFunctions.h
```diff
@@ -190,3 +190,6 @@ bool IsFloat(const std::string& number);
 bool IsDarkMode();
 void SetSuppressDarkMode(bool suppress);
 bool IsSuppressDarkMode();
+int GetPaletteSizeIndex();
+void SetPaletteSizeIndex(int index);
+bool IsLargePalette();
```
dkulp: Why are you adding this setting to UtilFunctions instead of handled in xLightsFrame like every other setting?  Dark mode is special as that is needed before the frame is created, but this doesn't fall into that.  

It should also be possible to have the size change immediately when the setting changes like the effect icon sizes.

Issue #3829, opened 2023-06-03
Setting start time >1 on video effects where xLights has "Hardware Video Decoding" Enabled causes the effect to display a blue screen.


bjhiltbrand:
**Describe the bug**
Either importing or creating a new sequence, if you have xLights preferences set to enable "Hardware Video Decoding", and if you work with a video effect and set the "Start Time" value to anything other than 0, 1, or 7(oddly enough), The video effect will only show a blue screen. I noticed this first when importing Visionary Light Show's "Master of Puppets" sequence. After working with Robert Gray and Kyle Bostick, we found that Robert could not reproduce the problem on his PC, and Kyle could not reproduce it on his Mac. Then, I unchecked "Hardware Video Decoding" on my PC and re-rendered, and it worked for me. Kyle verified that the issue was reproducible on his PC, but not his Mac. Therefore, it looks like it only affects PC installations. Kyle mentioned that he had hardware decoding on his PC enabled last Halloween when he used the sequence last, and it worked then. So we think it was a bug introduced somewhere between Halloween and now.

**To Reproduce**
Links:
Original question in Facebook:
https://www.facebook.com/groups/628061113896314/?multi_permalinks=6342861952416173&ref=share
Screen capture reproducing bug:
https://nas.hiltbrandhome.com/d/s/tpzhgsMYwj69YBl68VUxSjUUuYKpl2SW/ma9wKv0L8TnHMmnjs1dY9MZg5qMVlZPk-prXgNjJ4fQo
Please let me know where to send the sequence if necessary, as I don't want to leave it available in this bug report (It's a paid sequence from a vendor)

1. Ensure that Hardware Video Decoding is enabled in xLights Preferences
2. Import an existing sequence or create a new one
3. Select a video effect or add a new one.
4. Set video start time to any number greater than 1
5. See the model that you are working with go blue

**Expected behavior**
Video should play at the specified start time on the model.

**Screenshots**
https://nas.hiltbrandhome.com/d/s/tpzhgsMYwj69YBl68VUxSjUUuYKpl2SW/ma9wKv0L8TnHMmnjs1dY9MZg5qMVlZPk-prXgNjJ4fQo

**Versions (please complete the following information):**
 - OS: Windows 10 and Windows 11 - Affected; Mac (Not sure of the version) - Not Affected
 - xLights version Tested on 2023.01, 2023.02, 2023.06, 2023.07, and 2023.08 - All Affected

bjhiltbrand:
[...]

PR #5829, opened 2026-02-17
Fix video missing frames on start bug #3829

derwin12:
On some DirectX11 edge cases, the start position can result in missing the frames and hence show blue. The blue is the same blue that you would get if you try to play past a video end. #3829  Applicable on windows hardware video decoding.
Also added a couple lines to clear memory on exit.
2026.02
image

After
image


derwin12:
90% of AI is figuring out what it is lying to you about :-)

PR #5858, opened 2026-02-27
0000 windows fixes

derwin12:
Needed the includes and the "f" was added to clear some conversion noise in the build.


derwin12:
Oops.

PR #5869, opened 2026-03-05
Include the new model types

derwin12:
Just a quick change to handle the new DisplayAs types. Probably a fancier way to do it..


computergeek1507:
Oops I missed that....

Issue #5873, opened 2026-03-07
Check Sequence is reporting errors that do not exist. (2026.03.02)


cfudala:
Version 2026.03.2
Windows 11 PC

The error section of the check sequence report is showing unusual results.
Example:
"_rings
has invalid nodes outside the range of the parent model._"
"_❌
Custom model 'Burtons_Cross-1' Multiple strings but none starting at node 1._"

These models are configured correctly and do not have the reported errors.

Screen shots attached
Logs attached
Sequence attached

<img width="560" height="235" alt="Image" src="https://github.com/user-attachments/assets/1a1c0efe-7cf6-4d74-a18b-84d3481c6912" />

<img width="900" height="652" alt="Image" src="https://github.com/user-attachments/assets/420b42ce-e1d0-42be-813d-e85537c98e04" />

<img width="602" height="400" alt="Image" src="https://github.com/user-attachments/assets/1d206456-fead-4618-a4ab-e54de7171ec0" />

[xLightsProblem.zip](https://github.com/user-attachments/files/25806726/xLightsProblem.zip)

[AA_Initial (2).zip](https://github.com/user-attachments/files/25806727/AA_Initial.2.zip)

cfudala:
I am pretty sure I downloaded a nightly right before we left on a cruise (Tuesday after the FMM) I’ve been off line since then and yes the things I reported seem to have been fixed.   Any other areas I should jump into?  ThanksCarlSent from my iPhoneOn Mar 24, 2026, at 10:01 AM, Daryl ***@***.***> wrote:﻿derwin12 left a comment (xLightsSequencer/xLights#5873)
Pretty sure we have got most of these .. Note .. you really want to get a clean .03 layout and use a more recent nightly when it comes.

PR #5875, opened 2026-03-07
Check Sequence - DO NOT MERGE

derwin12:
This is from codex and is over my head in some aspects - so just for review and to see if it helps.
It does allow the check sequence to match the prior but there is some changes that might impact render etc.
#5873


derwin12:
we will re-assess with the next version....

Issue #5876, opened 2026-03-07
2026.03.02  Models with start nodes of 0


AGFazio:
2026.03.02 nightly 

**Describe the bug**
DMX floods showing invalid protocol and start channels of 0.

<img width="2090" height="1069" alt="Image" src="https://github.com/user-attachments/assets/99d3ce5a-3d80-45a2-82b9-70875ec7341e" />
<img width="875" height="991" alt="Image" src="https://github.com/user-attachments/assets/2aa856f2-e1a3-4196-b78c-b4b0fb0e36b7" />

[xlights_networks20260201.zip](https://github.com/user-attachments/files/25817461/xlights_networks20260201.zip)

[xLightsProblem20260302.zip](https://github.com/user-attachments/files/25817464/xLightsProblem20260302.zip)

derwin12:
Submitted fix that covers my version.. will check yours.

PR #5883, opened 2026-03-09
Fix default dmx channel #5876

derwin12:
The first dmx prop doesn't have the "channel" in the xml, it incorrectly defaults to 0. #5876


derwin12:
I will retest with these copilot commits...

Issue #5892, opened 2026-03-11
2026.03.2 Nightly - moving prop in visualizer gets stuck in recalc loop


derwin12:
Simply open the 2 model layout. One model with multiple ports. Drag the prop from the port to the next port and it gets stuck in a loop.
Recalc channels.
No need for a sequence. Tried various props - dont think it matters what the prop.
Both single string props but also had the issue with multi string props.

[Blank Sequence.zip](https://github.com/user-attachments/files/25907144/Blank.Sequence.zip)

PR #5919, opened 2026-03-20
 Moving prop in layout visualizer gets stuck in infinite recalculation loop #5892

derwin12:
When moving a prop/model in the layout visualizer, xLights could get stuck in an endless recalculation loop, making the application unresponsive.

Also fixes a bug where clearing a model's chain assignment (setting it back to "Beginning") was silently ignored, leaving the old chain value in place.


derwin12:
will redo with new updates.. (even though it says no conflicts)

Issue #5954, opened 2026-03-27
2026-03.03 New Shader Effect Crashing..


derwin12:
Shows this but also crashed 2 times.. never got a chance to select a shader.

<img width="529" height="523" alt="Image" src="https://github.com/user-attachments/assets/dccb40bd-8d23-4a63-ac3a-eb5093b2c7e6" />

PR #5961, opened 2026-03-29
Shader Effect preview and sequence rendering

derwin12:
Shaders render in sequence and preview. Fix the [x] buttons so the show in Windows (hoping they work right in Mac as well). #5954


xLights/ui/effectpanels/ShaderPanel.cpp
```diff
@@ -272,7 +272,7 @@ void ShaderPanel::OnFilePickerCtrl1FileChanged(wxFileDirPickerEvent& event)
     ObtainAccessToURL(newf.ToStdString());
 
     // if shader name hasnt changed dont reset
-    if (newf == last) {
+    if (newf == last && newf == "") {
```
computergeek1507: empty()

PR #6073, opened 2026-04-10
Add multi-select support to vendor model catalog download

heffneil:
## Summary
- Enable Ctrl+click and Shift+click to select multiple models in the vendor catalog dialog (Layout → Download)
- Batch download all selected models in one operation
- Auto-place downloaded models side by side in the layout
- Button label updates to show count (e.g. "Insert 7 Models") and auto-resizes

## Changes
- **VendorModelDialog.h/cpp**: Changed tree from `wxTR_SINGLE` to `wxTR_MULTIPLE`, added `GetSelectedWirings()` for filtering valid selections, `DownloadSelectedModels()` for batch download, and `GetFocusedItem()` helper to replace `GetSelection()` calls (required for multi-select trees)
- **LayoutPanel.cpp**: Extended `GetXlightsModel()` to capture additional downloads, loop through extras in `FinalizeModel()` creating and placing each model offset horizontally

## Backward Compatibility
- Single-select works identically to before
- Double-click still downloads a single item immediately
- All existing per-model prompts (groups, aliases, vendor certification) fire per model as before

## Test plan
- [ ] Single-select: select one wiring node, click Insert Model — works as before
- [ ] Multi-select: Ctrl+click multiple wiring nodes, verify button shows "Insert N Models"
- [ ] Click Insert — all models download and appear side by side in layout
- [ ] Double-click a wiring node — downloads single model (unchanged behavior)
- [ ] Mixed selection (vendor + wiring nodes) — only wirings are processed
- [ ] Cancel from dialog — no models placed, clean state


cybercop23:
@heffneil did you compile xLights and tested this? If so, can you please post an image of how the vendor model tree looks like or better yet a video showing exactly how this works?

heffneil:
It allows you to select multiple models like anything else on a computer - if you hold control you can pick individually and if you use shift you can select a series of them. Perhaps I don't know procedure.  Send me links to yours and andy's that show images and videos so I cam comply.  Of course I tested this and it works.  This is for the mac to be clear.  I do not run Xlights on windows.
Thanks

cybercop23:
You compiled xLights on the Mac and run it. Use the screenshot to record the flow and include it here.

src-ui-wx/ui/layout/LayoutPanel.cpp
```diff
@@ -4029,7 +4029,7 @@ void LayoutPanel::FinalizeModel()
 
         // Process additional models from multi-select download
         if (!additionalModels.empty()) {
-            float xOffset = 200.0f; // horizontal spacing between models in layout units
+            float xOffset = 200.0f;
```
computergeek1507: Why remove comments

heffneil: I didn't remove them - I would say thats a claude (intern) suggestion.  I have added Daryl's recommended guidance and I did notice that it said "keep comments in code to a minium" so perhaps it was trying to clean them up?

heffneil:
Pushed a commit addressing two of the Copilot review findings:

1. **Dedup in GetSelectedWirings()** - Now prevents duplicate MModelWiring* entries when a user selects both a Model node and its child Wiring node (Copilot comment #5)
2. **Cache GetFocusedItem() in OnTreeCtrl_NavigatorSelectionChanged()** - Reuses the already-captured startid instead of calling GetFocusedItem() again, avoiding potential inconsistency if the selection changes between calls (Copilot comments #2-4)

The remaining Copilot suggestions (C-style casts, ::tolower signed char, DownloadedModelInfo location, AddASAPWork in loop) were reviewed and determined to be either not applicable to this PR or already correct in the existing code.

heffneil:
I have updated based on the recommendations from co-pilot.  I have instructed to re-install the original comments.  There might need an updated PR for this.

heffneil:
Merged master and applied the fix from PR #6075 (SetControllerName(NO_CONTROLLER, true) after CreateDefaultModelFromSavedModelNode) to the multi-select batch download path as well. Without this, extra downloaded models in a batch would get stuck at start channel 1 instead of being auto-assigned.

Also addressed the Copilot memory-leak concern by using a temporary return pointer so the original allocation is deterministically freed on failure.

Build verified on macOS Debug.

heffneil:
Addressed the three remaining Copilot findings:

1. **Safe tolower**: batch extension check now uses a lambda with unsigned char cast to avoid undefined behavior on signed char values
2. **Single Layout() call**: ValidateWindow button-state block now sets enable/label in each branch then calls Layout() once at the end, reducing UI churn on frequent selection changes
3. **Clean state on cancel**: DownloadModel now returns bool. DownloadSelectedModels aborts the whole batch and clears all state if any download fails or is cancelled, matching the PR test plan promise of no partial results on cancel

Build verified on macOS Debug.

heffneil:
Addressed the second round of Copilot findings:

1. **Stale additionalModels vector**: clear() before assign so reused vectors do not carry stale entries when a new download returns one or fewer models
2. **Dynamic batch spacing**: replaced hard-coded xOffset = 200.0f with per-model GetRestorableMWidth() + padding (named constants, min-offset floor) so side-by-side placement scales with model size
3. **Null-safe Layout()**: guard against null parent/sizer in ValidateWindow before calling Layout()

Confirmed branch is up to date with master (includes 2026.06 tag and post-release commits). Build green on macOS Debug.

heffneil:
Addressed latest Copilot findings:

1. **Double-delete fix (LayoutPanel.cpp)**: CreateDefaultModelFromSavedModelNode takes ownership of the passed-in model and deletes it internally before returning. On failure it returns nullptr with the original pointer already freed, so the manual delete extraModel in the failure path was a double-delete. Removed it to match the primary model path which just continues without manual cleanup.

2. **Native wxTreeCtrl::GetFocusedItem() (VendorModelDialog.cpp)**: confirmed the API is available and already used elsewhere in the codebase (SeqSettingsDialog.cpp). Replaced the GetSelections-based heuristic with a direct call to TreeCtrl_Navigator->GetFocusedItem() so the actual focused item is returned rather than the last-selected one.

Build green on macOS Debug.

heffneil:
## Complete Copilot Review Audit

20 total Copilot comments across 4 review rounds. Current HEAD: `d80306e3c`.

### Round 1 (April 10) — 11 comments

| # | Finding | Status |
|---|---|---|
| 1 | `GetFocusedItem()` not actually focused | Fixed — uses `wxTreeCtrl::GetFocusedItem()` |
| 2-4 | Re-entrancy guard recomputes `GetFocusedItem()` | Fixed — cached `startid` reused |
| 5 | Duplicate `MModelWiring*` entries possible | Fixed — `std::find` dedup |
| 6 | C-style casts | Skipped (matches 24 existing casts in file) |
| 7 | `::tolower` UB on signed `char` | Fixed — unsigned char lambda |
| 8-10 | `AddASAPWork` in loop + magic `xOffset` | Fixed — single call, dynamic `GetRestorableMWidth` spacing |
| 11 | `DownloadedModelInfo` coupling | Skipped (both files same UI layer, all 4 consumers already include dialog header) |

### Round 2 (April 11) — 3 comments

| # | Finding | Status |
|---|---|---|
| 12 | Leak in `extraModel` overwrite | Fixed — later refined to match primary path |
| 13 | `Layout()` in every branch | Fixed — single call at end |
| 14 | Partial batch on cancel | Fixed — `DownloadModel` returns `bool`, batch aborts + clears |

### Round 3 (April 16 AM) — 4 comments

| # | Finding | Status |
|---|---|---|
| 15 | Need null-safe `Layout()` | Fixed — `wxWindow*`/`wxSizer*` checks |
| 16 | `additionalModels` not cleared before assign | Fixed — explicit `clear()` first |
| 17 | `::tolower` UB (re-flag) | Already fixed in Round 1 |
| 18 | Hard-coded `xOffset` (re-flag) | Fixed — dynamic per-model width |

### Round 4 (April 16 PM) — 2 comments

| # | Finding | Status |
|---|---|---|
| 19 | Double-delete: `CreateDefaultModelFromSavedModelNode` takes ownership | Fixed — removed manual delete, matches primary path |
| 20 | `GetFocusedItem` should delegate to native API | Fixed |

### Summary

- **18 distinct findings addressed** (several duplicates across rounds)
- **2 intentionally deferred** with documented rationale:
  - C-style casts: 24 identical pre-existing casts in the same file; fixing only the new ones creates inconsistent style — right fix is a separate cleanup PR
  - `DownloadedModelInfo` coupling: both files in `src-ui-wx/ui/` (not a core/UI boundary), all 4 existing consumers of `VendorModelDialog.h` already include it, extraction would add a file without reducing dependency graph

**Build:** Clean on macOS Debug.  
**PR mergeability:** `MERGEABLE`, `UNSTABLE` only from pending CI (no conflicts).  
**Platform scope:** All three modified files (`VendorModelDialog.cpp/.h`, `LayoutPanel.cpp`) live in `src-ui-wx/` — the cross-platform wxWidgets UI layer. No platform-specific APIs used, no new files added, so no project file updates needed for Windows, macOS, or Linux.

heffneil:
holy crap! merged!

derwin12:
Now xlights crashes on close :-(

heffneil:
i believe it :)

Issue #6077, opened 2026-04-11
Model list in layout default size


AGFazio:
Make the model list have a minimum size so the model settings window below can't completely hide the list if the seperator bar is raised too high.

<img width="882" height="702" alt="Image" src="https://github.com/user-attachments/assets/b3cb4242-aeb8-47a0-9e41-0803a5fa9bed" />

AGFazio:
PR 6082

PR #6082, opened 2026-04-11
#6077 Don't completely hide model/group list in Layout

AGFazio:
Previously you could drag the seperator bar between the model/group list and model/group settings window to be able to completely hide the list. This change makes the model/group list a fixed size that can't be hidden.


AGFazio:
This is also addressed in PR 6090. This can be closed if that is implemented.

AGFazio:
addressed in PR 6090.No longer needed.

Issue #5047, opened 2024-12-10
Dock Effects Presets Window


McCoyCOREUSA:
Would be great to dock the Effects Preset window. For those of us who use it a lot, this would be handy.

AGFazio:
add to view...windows menu list

AGFazio:
PR 6128

PR #6128, opened 2026-04-16
#5047 Make Effect Presets panel dockable

AGFazio:
Converts EffectTreeDialog from a wxDialog to a wxPanel hosted in the AUI manager, making it dockable/floatable like other sequencer panels. Adds a View menu item to show/hide it and a lazy-init flag so presets are only loaded when first shown.


AGFazio:
made wrong changes new is 6152

Issue #6142, opened 2026-04-17
xlights media conversion error


dreamwishhope365-netizen:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
downloaded 2026.06 opened up sequence, prompted with a file conversion warning. Tried to convert and received error.

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)

**Expected behavior**
The video is already a .mov video, I don't think it needs conversion. 

**Screenshots**

**Versions (please complete the following information):**
 - OS: Windows 11

[xLights_spdlog.log](https://github.com/user-attachments/files/26823669/xLights_spdlog.log)

<img width="915" height="598" alt="Image" src="https://github.com/user-attachments/assets/51e4b019-ad82-4989-9d2b-208783987b3d" />
[xLights_spdlog.log](https://github.com/user-attachments/files/26823818/xLights_spdlog.log)

<img width="580" height="223" alt="Image" src="https://github.com/user-attachments/assets/7eaab7d1-f3bf-4b57-95cf-1699a2cfd970" />

 - xLights version 2026.06

**Additional context**

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

derwin12:
I think what it is saying is the conversion we have in xlights cant handle your file. You need to do the conversion outside of xlights.

derwin12:
In the next update

derwin12:
Fix is being reviewed..

PR #6154, opened 2026-04-17
Handle qtrle codec mov conversions (#6142)

derwin12:
Add the ability to convert this codec. 

[skip ci]


derwin12:
Handle conversion of various non supported mov formats as well. (#6142)

[skip ci]

PR #6172, opened 2026-04-19
Replace P3/P4 zero-sentinel with explicit _hasStartEndLevel flag

MrPierreB:
Addresses code review feedback from Copilot on #6150 (which landed before the review was actioned).

  The original implementation detected "legacy" curves (pre-Start/End Level feature) by checking if both P3 and P4 were
  zero. Copilot correctly flagged that this conflates "unset" with "intentionally set to zero", making Start=0 / End=0
  impossible to represent correctly.

  Changes:
  - Added explicit `_hasStartEndLevel` bool flag (defaults false)
  - Added `SE=Y` serialization key so the flag survives save/load
  - All RenderType and scale function guards now use the flag instead of the value check

  Backward compatible — existing sequences have no `SE=Y` key, so the flag stays false and rendering is identical to
  before.

### DIFF
--- a/src-core/render/ValueCurve.cpp
+++ b/src-core/render/ValueCurve.cpp
@@ -748,15 +748,14 @@ void ValueCurve::UnFixChangedScale(float newmin, float newmax)
         _parameter2 = _parameter2 * newrange / oldrange + mindiff;
     }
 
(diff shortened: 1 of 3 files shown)


MrPierreB:
Addressed all Copilot review feedback in follow-up commit.

PR #6184, opened 2026-04-20
Add Bulk Edit Rotate X / Y / Z to the Layout tab

heffneil:
## Summary

Adds three new items to the existing **Bulk Edit** right-click submenu in the Layout tab so a rotation can be applied to every selected model in one operation:

- **Rotate X**
- **Rotate Y**
- **Rotate Z**

Each option opens a text entry dialog pre-filled with the first selected unlocked model's current rotation on that axis, and applies the entered angle (in degrees, decimals supported) to every selected non-locked model. The existing tree selection is preserved through the reload, and a single Ctrl-Z reverts the entire batch as one undo step.

## Design choices

- **Three menu items, one axis each** rather than a combined X/Y/Z dialog, since the most common workflow is rotating a group of props around a single axis at a time.
- **Locked models are skipped** silently (same rule the property-grid rotation handler uses via `loc.IsLocked()`).
- **Float parsing uses `std::strtod`** instead of `std::stod` per CLAUDE.md ("xLights has effectively no exception handling").
- **Undo via `CreateUndoPoint("All", ...)`** takes a full snapshot before any change, so one undo rolls back all affected models together. The snapshot is deferred until after the dialog is accepted and the value parses successfully, so Cancel or bad input doesn't leave a phantom entry on the undo stack.
- **Selection preservation** mirrors `BulkEditPixelSize` — capture `GetSelectedTreeModelPaths()` before the dialog, then call `ReselectTreeModels()` after the reload.
- The three public entry points (`BulkEditRotateX/Y/Z`) delegate to a single `BulkEditRotateAxis(char axis)` to avoid triplicate code.

## Files changed

- `src-ui-wx/layout/LayoutPanel.h` — three new `ID_PREVIEW_BULKEDIT_ROTATE{X,Y,Z}` IDs and method declarations
- `src-ui-wx/layout/LayoutPanel.cpp` — ID definitions, menu entries under Bulk Edit (after Shadow Model For, before the controller port section), dispatch branches in both `OnPreviewModelPopup` sites, and the `BulkEditRotateAxis` implementation
- `README.txt` — release note under 2026.07

No new files added, so no `Xlights.vcxproj` / `xLights.cbp` / Xcode project updates needed. All changes live in `src-ui-wx/` (cross-platform UI layer), so the feature applies equally to Windows, macOS, and Linux builds.

## Test plan

- [x] Build green on macOS Debug
- [x] Multi-select 3+ models, Bulk Edit → Rotate X/Y/Z → value applied to all
- [x] Locked model mixed into selection — skipped without error, others still rotate
- [x] Tree selection preserved after the operation
- [x] Cancel button on the dialog — no change, no undo entry
- [x] Invalid text (e.g. `abc`) — aborts silently, no change
- [x] Ctrl-Z after a bulk rotate — all models revert in one step
- [ ] Windows smoke test (code is pure wx + std, no platform-specific APIs; would appreciate a Windows reviewer confirming)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,9 +11,6 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.07  April ??, 2026
-    -enh (heffneil)             Layout tab: Bulk Edit menu now includes Rotate X / Rotate Y / Rotate Z items so a
-                                rotation can be applied to every selected model in one operation, with the tree
-                                selection preserved after the change.
     -enh (dkulp)                Shape effect: Effect panel now organises its 27 properties into Shape / Size / Motion /
                                 Triggers tabs instead of a flat scroll, matching the grouped layout other large effects use.
     -enh (PB)                   Value curve Exponential, Logarithmic, and Parabolic types now support Start/End.
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,9 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.07  April ??, 2026
+    -enh (heffneil)             Layout tab: Bulk Edit menu now includes Rotate X / Rotate Y / Rotate Z items so a
+                                rotation can be applied to every selected model in one operation, with the tree
+                                selection preserved after the change.
     -enh (dkulp)                Shape effect: Effect panel now organises its 27 properties into Shape / Size / Motion /
                                 Triggers tabs instead of a flat scroll, matching the grouped layout other large effects use.
     -enh (PB)                   Value curve Exponential, Logarithmic, and Parabolic types now support Start/End.
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -2310,6 +2310,15 @@ void LayoutPanel::BulkEditRotateAxis(char axis) {
     }
     float newAngle = static_cast<float>(angle);
 
+    // Snapshot the full models XML before any changes so a single Ctrl-Z
+    // rolls back the entire bulk rotation as one undo step. "All" type
+    // captures the complete state - this is the same pattern used by
+    // multi-model delete / group operations elsewhere in LayoutPanel.
+    // We defer taking the snapshot until after the dialog is accepted and
+    // the value parsed, so Cancel / invalid input doesn't leave a no-op
+    // entry on the undo stack.
+    CreateUndoPoint("All", wxString::Format("BulkRotate%c", axis).ToStdString(), entered);
+
     int changed = 0;
     for (Model* model : modelsToEdit) {
         if (model == nullptr) continue;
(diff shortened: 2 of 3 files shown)


heffneil:
Pushed a follow-up commit fixing two real bugs found in live testing:

**1. Wrong work item type**: was using WORK_RELOAD_ALLMODELS (which tears down and rebuilds every model from XML). Switched to WORK_SCREEN_LOCATION_CHANGE to match how Nudge() and the property-grid rotation handlers persist geometry changes. This also removes the need for the ClearSelectedModel / ReselectTreeModels dance since the models do not get torn down.

**2. Selection bounding box not rotating with the model**: BoxedScreenLocation caches a rotate_quat quaternion that is rebuilt only inside Init() when rotation_init==true. SetRotateX/Y/Z only writes the raw rotatex/y/z fields; Reload() sets the flag but nothing in the draw loop calls Init() to honor it. So the model drew at the new angle (ApplyModelViewMatrices uses the raw fields directly) but TranslatePoint (which drives the selection bounding box bounds transform) kept using the stale quaternion. The user found they could work around it by closing and reopening the Layout tab, which naturally triggers a load-time Init. Calling Reload() + Init() immediately after each SetRotate*() makes the cache consistent on the first apply.

Live-tested on macOS Debug: white selection wireframe now rotates with the model on the first click; undo still rolls the whole batch back in one step.

heffneil:
Addressed the Copilot finding plus reverted the README change (per maintainer guidance not to touch release notes):

**Copilot: undo snapshot before editability check**
The CreateUndoPoint("All", ...) call was running before the loop could determine whether any model would actually change. When every selected model is locked, the function would return early with changed == 0 but the undo stack already had a no-op entry and the full XML snapshot work was wasted.

Restructured to filter editableModels (non-null, non-locked) up front, early-return if empty, then take the undo snapshot only after we know at least one model will change AND the entered angle parsed successfully.

**Release-note revert**
Dropped the README.txt lines this PR previously added.

Build green on macOS Debug.

heffneil:
Addressed the two new Copilot findings from the latest review round:

**1. Input validation (non-finite + out-of-range)**
std::strtod accepts "nan" / "inf" and any numeric string. Non-finite would propagate garbage into the transform matrices; out-of-range values get silently reset to 0 by BoxedScreenLocation::Init() - so a user typing 200 would lose their input without warning.

Now rejects !std::isfinite() silently, and shows a wxMessageBox warning for values outside [-180, 180] (the same range the rotation property grid enforces via Min/Max attributes in ScreenLocationPropertyHelper).

**2. Undo selection hint**
CreateUndoPoint("All", model, ...) stores the second argument into undoBuffer[idx].model, which DoUndo() later passes as the "selectedModel" hint to AddASAPWork (LayoutPanel.cpp:8209). A label like "BulkRotateX" is not a real model name - post-undo selection logic would try to resolve a nonexistent model.

Now passes editableModels.front()->name as the model hint, with the operation label moved to the key field.

Also added <cmath> include explicitly so std::isfinite resolves under NO_PCH per CLAUDE.md guidance.

Build green on macOS Debug.

hssusc:
This is another step toward making "organic" Peace Stakes (or similar props) easier to set up. 

When Peace Stakes were introduced, 3D was just coming into xLights. The cube model quickly became the preferred method for these props just for simplicity and ease of setup. That is great, but makes it difficult for those that want to lay them out in a more "organic" or random pattern.

heffneil:
Hi @dkulp / maintainers — friendly bump on this one. PR has been open since 2026-04-20, CI is green, mergeable, all Copilot threads addressed. Other PRs of mine that landed after it (#6186, #6241) have already merged.

Happy to rebase if it's drifted, just let me know if anything is still blocking. Thanks!

PR #6189, opened 2026-04-21
Replace model(s) with this model: multi-target dialog with filter and checkboxes

### LINKED ISSUE #4462: Add option to replace multiple models with a new model
Currently there is an option to replace a model with this model, which is useful when someone makes an incorrect model, and sequences on it, and then wants to replace the model with a corrected model.  Many times, the incorrect model was either downloaded from a vendor or was copied and pasted to multiple locations on the layout during the initial setup.  In order to replace multiple incorrect models, the user has to make multiple copies of the good model, and replace each incorrect model one by one.  Would be nice if there were an option to replace the multiple incorrect models with the single corrected model.

heffneil:
Fixes #4462
Ref: [comment on the issue](https://github.com/xLightsSequencer/xLights/issues/4462#issuecomment-3793533322)

## Summary

Replaces the old `wxSingleChoiceDialog`-based "Replace A Model With This Model" flow with a richer dialog that can replace **multiple** target models with a copy of the current one in a single operation. The single-select case still works — you just tick one checkbox instead of picking from a radio list.

## What's new

- **Filtered model picker** — live, case-insensitive substring filter. Check state persists across filter changes so you can narrow down one subset, check those, narrow down another, and keep both groups selected.
- **Tri-state master checkbox in the header** — sits next to the bold "Model Name" label. Clicking it selects/clears every currently visible row in one action, and reflects the aggregate state (all / none / mixed).
- **Three batch options** replace the three separate Yes/No prompts of the old flow — asked once, applied to every replaced target:
  - Copy target's start channel + controller settings to the replacement
  - Merge target's submodels into the replacement
  - Use target's original size and position
- **Result confirmation** — info / warning / error `wxMessageBox` at the end summarising how many models were replaced and naming any that failed, so the user knows the operation completed.

## Implementation notes

- Source model's XML is snapshotted once via `CopyPasteBaseObject`; each target gets a fresh `AllModels.CreateModel()` clone from that snapshot. The original source stays intact across the whole batch.
- The per-target rename dance (park target under a unique trash name → rename the clone into the target's original name → delete the trashed target) mirrors the existing single-replace pattern, extended to N targets in a loop.
- The alias-prompt behaviour ("Save old name as alias?") is saved and set to `"Always No"` for the duration of the batch, then restored on every exit path. Prevents the user from being asked N × 3 times about aliasing when they're conceptually **replacing** models rather than renaming them. Global preference is untouched for subsequent rename operations.
- Single `CreateUndoPoint("All", ...)` snapshot before any change — one Ctrl-Z rolls the entire batch back.
- Dialog's list is built from real `wxCheckBox` + `wxStaticText` widgets inside a `wxScrolledWindow`. Every row (including the header) uses the same 2-column sizer pattern, so alignment between the header checkbox and list rows is automatic. Gives native `wxCheckBox` rendering on Windows, macOS, and Linux — no custom drawing.

## Scope

- Only `src-ui-wx/layout/LayoutPanel.{cpp,h}` changed
- No new source files → no project file updates needed for `Xlights.vcxproj` / `xLights.cbp` / Xcode pbxproj
- Pure wxWidgets + STL, cross-platform
- The old `ReplaceModel()` body was removed; the new `ReplaceModel()` handles both single and multi-target cases

## Test plan

- [x] Single-target replace works end-to-end (verified from preview right-click menu)
- [x] Multi-target replace (2+ selections) works
- [x] Filter narrows the list and check state persists across filter changes
- [x] Master tri-state checkbox correctly reflects and drives row state
- [x] Each of the three batch options is honoured (start channel / submodel merge / size+pos)
- [x] No alias prompts appear during the batch (preference is restored afterward)
- [x] Ctrl-Z reverts the entire batch in one step
- [x] Confirmation box appears at the end in all three outcome shapes
- [ ] Windows / Linux smoke test — code is pure wx + std, no platform-specific APIs, but a Windows reviewer confirming would be appreciated

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -8588,11 +8588,11 @@ public:
         top->Add(new wxStaticText(this, wxID_ANY, "For each replaced model, also:"),
                  0, wxLEFT | wxRIGHT | wxTOP, 8);
         _copyStartChannelCB = new wxCheckBox(this, wxID_ANY,
-            "Copy target's start channel and controller settings");
+            "Keep each target's start channel and controller settings");
         _mergeSubmodelsCB = new wxCheckBox(this, wxID_ANY,
-            "Merge target's submodels into the replacement");
+            "Keep each target's submodels");
         _copySizePosCB = new wxCheckBox(this, wxID_ANY,
-            "Use target's original size and position");
+            "Keep each target's position and size");
         _copyStartChannelCB->SetValue(true);
         _copySizePosCB->SetValue(true);
         const int cbFlags = wxLEFT | wxRIGHT | wxTOP;
@@ -8615,10 +8615,10 @@ public:
             UpdateMasterState();
         });
         _masterCB->Bind(wxEVT_CHECKBOX, [this](wxCommandEvent&) {
-            // wx's default tri-state click cycles 2-state visibly (unchecked
-            // -> checked -> unchecked). IsChecked() reflects the desired new
-            // state after the click - apply to every visible row.
-            const bool shouldCheck = _masterCB->IsChecked();
+            // 3-state checkbox: use Get3StateValue() — IsChecked() asserts
+            // on a 3-state in wxWidgets debug builds. Treat anything that
+            // isn't explicitly "checked" as an uncheck-all.
+            const bool shouldCheck = _masterCB->Get3StateValue() == wxCHK_CHECKED;
             for (const auto& row : _rows) {
                 row.cb->SetValue(shouldCheck);
                 if (shouldCheck) {
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -8766,14 +8766,17 @@ void LayoutPanel::ReplaceModel()
     // temp-name, target trash-name, clone final-name) and the user didn't ask
     // to rename anything - they asked to replace wholesale, so the alias
     // question doesn't apply. Restore the previous setting after the batch so
-    // normal rename operations still obey the user's preference. Uses a
-    // scope_guard-style RAII restore so we don't leak the override on early
-    // return paths.
-    const wxString savedAliasBehavior = xlights->GetRenameModelAliasPromptBehavior();
+    // normal rename operations still obey the user's preference.
+    //
+    // Real RAII guard so the restore is resilient to any early return path
+    // (current or future) - the destructor fires on scope exit regardless of
+    // how we leave this function.
+    struct AliasBehaviorRestore {
+        xLightsFrame* frame;
+        wxString saved;
+        ~AliasBehaviorRestore() { frame->SetRenameModelAliasPromptBehavior(saved); }
+    } aliasGuard{ xlights, xlights->GetRenameModelAliasPromptBehavior() };
     xlights->SetRenameModelAliasPromptBehavior("Always No");
-    auto restoreAliasBehavior = [&]() {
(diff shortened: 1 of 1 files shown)


heffneil:
Addressed all three Copilot review findings in commit 5f917e066:

1. **Temp / trash name uniqueness** (lines 8045, 8091): both `__xl_rmm_new_N` and `__xl_rmm_old_N` now loop the suffix until `AllModels[name]` returns nullptr before committing. Prevents the catastrophic case where `AllModels.AddModel()` silently deletes a pre-existing user model with a colliding double-underscore name.
2. **`Delete()` return value handled**: if deletion of the trashed target fails, the failure is recorded in the `failures` list (so the final summary message surfaces it) and `successCount` is not incremented. Previously a failed delete would silently leave a renamed stray behind.
3. **Real RAII guard**: replaced the manual `restoreAliasBehavior()` lambda with an actual `AliasBehaviorRestore` struct whose destructor restores the saved preference on scope exit. Any future early-return paths stay correct automatically, and the misleading "scope_guard-style" comment now matches the code.

Build green on macOS Debug.

computergeek1507:
Do you have any screenshots or video of this in action?

heffneil:
Screenshot 2026-05-25 at 11 02 37 PM
Screenshot 2026-05-25 at 11 02 28 PM

heffneil:
Pushed two small fixes after a local re-test:

**Crash fix** — clicking the master "select all" checkbox tripped a wxWidgets debug assert (`checkbox.h:60 'IsChecked() doesn't make sense for a three state checkbox, Use Get3StateValue() instead'`). The handler was calling `IsChecked()` on the 3-state master. Switched to `Get3StateValue() == wxCHK_CHECKED`, preserving the two-state-visual cycle the original comment described.

**Label clarification** — the three per-target checkboxes now read "Keep each target's ..." instead of "Copy / Use target's ...", so the multi-target semantics are unambiguous (the code was already correct — each clone keeps its own target's start channel / submodels / position; no collisions or stacked models).

Branch was rebased onto current master to pick up the post-April toolkit changes (ARC-across-all-Apple-targets, etc.). PR is now mergeable on the latest Xcode/Mac submodule combo.

heffneil:
Addressed all four Copilot threads in 22855caaa:

1. **Unicode-safe filter** (line 8657) — `std::tolower` / `std::string` byte-folded; non-ASCII model names wouldn't match. Switched to `wxString::Lower()` + `Contains()` and convert candidate names via `wxString::FromUTF8`. Platform locale now handles case-folding correctly.

2. **OK disabled with empty selection** (line 8755) — cached the OK button on construction, toggled inside `UpdateMasterState()` which already fires on every row toggle + filter change. OK enables the moment a row is checked, disables when the last is cleared.

3. **Rollback on Delete failure** (line 8884) — if `Delete(trashName)` fails the swap is now reversed (clone → `tmpNewName`, original → `targetName`, clone deleted) so the layout returns to its pre-iteration state for that target. Best-effort; if the rollback renames also fail we record the failure and continue, but the partial state is no worse than the pre-fix code.

4. **AddASAPWork context** (line 8896) — now passes `'ReplaceModel: source=<name> replaced=<N>'` instead of bare `'ReplaceModel'` so downstream observers and logs can scope the change.

computergeek1507:
I want to double check the process flow, the user would download/import the "new" model, then they would replace a model with the new model, it would replace all the selected models and delete the "new" model.

computergeek1507:
I'm not a huge fan of infunction classes if you could put your ReplaceModelDialog class in a separate header file I think it is cleaner

heffneil:
I wish I could help alas I have no more tokens to spare.

computergeek1507:
I added this in a separate commit

PR #6191, opened 2026-04-21
Controller Visualiser: realtime filter that dims non-matching models on the wiring pane

heffneil:
## Summary

Adds a realtime name filter to the **left pane of the Controller Visualiser** (`ControllerModelDialog`). This is the pane where you actually see the models that are already wired onto the controller's ports — previously there was no way to find a specific model on a dense controller except scrolling.

The existing filter on the right pane ("available models to drag onto the controller") is untouched.

## How it works

- New `wxSearchCtrl` sits above `PanelController`
- Bound to `wxEVT_TEXT` so filtering is live on every keystroke (no Enter required), matching the pattern of the other filter improvements shipped in #6190
- Matching models: drawn at full opacity / full color
- Non-matching models: drawn with a pale grey brush, outline and text — still visible, still clickable, just visually de-emphasized so the user's eye finds the matches
- Cancel (X) button clears the filter
- Works during drag / copy / paste — only the paint is affected; the underlying selection, hit-testing, and click handling are unchanged
- Printing / export operations suppress the dimming so output is untouched regardless of filter state

## Implementation

- `BaseCMObject` gains `static wxString _visualizerFilterLower` so `ModelCMObject::Draw` can check "do I match?" with a cheap `Contains()` per model per paint, without plumbing a new parameter through every `Draw` call chain in the dialog
- The filter is lowercased once on each text change (inside the handler) — paint-time comparisons reuse that pre-lowercased string instead of re-transforming each frame
- Dimming uses solid pale greys (RGB 240/200/170) rather than alpha blending. `wxDC` alpha support is uneven across backends we ship (`wxMSW`, `wxOSX`, `wxGTK`) — solids render consistently on all three platforms
- The new widget is constructed **outside** the wxSmith `//(* ... //*)` guard. `Panel3`'s existing `FlexGridSizer5` is preserved intact and re-parented into a new `wxBoxSizer(wxVERTICAL)`. `SetSizer(wrap, false)` keeps the old sizer alive rather than deleting it. No `.wxs` file edit needed.

## Files changed

- `src-ui-wx/setup/ControllerModelDialog.h` — new `TextCtrl_VisualizerFilter` member + `OnTextCtrl_VisualizerFilterText` handler
- `src-ui-wx/setup/ControllerModelDialog.cpp` — static filter on `BaseCMObject`, dim logic in `ModelCMObject::Draw`, widget wiring after the wxSmith guard

No new files, no project file updates needed for Windows / macOS / Linux.

## Test plan

- [x] Filter appears above the visualizer canvas
- [x] Typing narrows the visible set in realtime — matches stay solid, non-matches fade
- [x] X / cancel button clears the filter
- [x] Matching / dimming is purely visual — dimmed models are still interactive (click, drag, menu)
- [x] Dragging a model does not fade the drag source mid-drag
- [ ] Windows / Linux smoke test — code is pure wx + std, no platform-specific APIs


heffneil:
Addressed all seven Copilot review findings in commit c3d46f3a2:

1. **Dimming scoped to controller pane only** - gated on portMargin>0 so the right-pane "available models" list (which calls Draw with portMargin==0) stays untouched. Printing paths also use 0 and are naturally excluded.
2. **Dropped the dead !_dragging check** - it read BaseCMObject::_dragging (a bool nothing ever sets) instead of the dialogs drag pointer, so the guard never fired anyway.
3. **Lowercase-caching comment** rewritten to match what the code actually does (filter side is pre-lowercased on the handler; name side is lowercased per call, which is negligible for the tens of models on a single controller).
4. **Theme-aware dim palette** - replaced the hard-coded 240/200/170 greys with blends between __backgroundBrush and __textForeground using IsDarkMode()-tuned weights, so dimmed items recede toward the background in both light and dark mode instead of standing out in dark mode.
5. **Tabs indentation** in the .h block now matches surrounding wxSmith style.
6. **OnTextCtrl_VisualizerFilterText declaration moved outside the //(*Handlers ...) wxSmith guard** so a future wxSmith regeneration cannot drop it.
7. **Static _visualizerFilterLower cleared** at the start of the dialog constructor and end of the destructor so closing + reopening the dialog never shows an empty SearchCtrl while a stale filter still dims models.

Build green on macOS Debug.

derwin12:
Any issue ticket with some screen grabs or anything?

heffneil:
happy to demo for you in the zoomy roomy

heffneil:
Screenshot 2026-04-21 at 12 53 14 PM
Here you go

heffneil:
Screenshot 2026-04-21 at 12 57 09 PM

cybercop23:
Should the filter be on both lists? I think so.

heffneil:
the list on the right? There is another filter for it?

cybercop23:
> the list on the right? There is another filter for it?

right, but that's for that side, you wouldn't want to do it on both. So say of the right/model pool doesn't have a filter, apply the same filter from the left, so you can see all wreaths (btw no more model for that - use the cirlce model)

heffneil:
updated to conform to the request:

Screenshot 2026-06-19 at 12 42 23 PM
Screenshot 2026-06-19 at 12 42 11 PM

PR #6192, opened 2026-04-21
Fix copy/paste: intelligent placement + only new copy selected after paste

### LINKED ISSUE #2: Meteor starfield doesnt work
Meteor starfield doesnt work

8/12/2013: Matt - Starfield changed to implode and explode, and are now working.

heffneil:
## Summary

Fixes two related user-visible bugs in copy/paste on the Layout tab:

1. **Copy lands on top of (or inside) another model** — previously the paste bumped the XML position by a fixed 40 world units when an exact integer-centre match was detected. Both parts were too weak: small enough not to matter visually against a wide matrix, and exact-centre matching missed the common case where the paste lands inside a wide model's bounding box without being on its exact centre.

2. **Both the original and the new copy were left selected** after a paste. The user expects only the fresh copy to be highlighted.

## Fix #1 — offset by the collided model's own width

The overlap test is now:

```cpp
|paste_x - other.pos.x| < other.GetRestorableMWidth() / 2
```

(plus exact integer match on Y/Z, which are historically not bumped). When a collision is detected, the bump is `other.width + 20 padding`. So:

- A small 50-unit arch bumps by ~70 world units
- A large 600-unit matrix bumps by ~620 world units
- The copy always lands *just past* whatever it would overlap, regardless of size

This also fixes the repeated-paste cascade: each subsequent `Ctrl-V` steps past the full extent of every already-placed copy, so stacking copies run out to the right in clean single-model-width increments instead of occasionally landing inside a neighbour.

A small `loopGuard` caps the search at 500 iterations as a belt-and-suspenders against pathological cases (no real layout has 500 identical models in a straight row).

## Fix #2 — clear the tree-widget selection before paste

Root cause: the models tree is a `wxTreeListCtrl` with `wxTL_MULTIPLE` style. `UnSelectAllModels()` clears our per-Model `Selected()` bools and internal tracking vectors but does **not** call `TreeListViewModels->UnselectAll()`. So when `WORK_RELOAD_MODELLIST` eventually runs `SelectBaseObject -> SelectModelInTree -> TreeListViewModels->Select(newItem)` on the new copy, the native widget **adds** the new row to the existing selection rather than replacing it — leaving both highlighted.

Fix: call both `UnSelectAllModels()` and `UnSelectAllModelsInTree()` at the start of `DoPaste`. The latter invokes `TreeListViewModels->UnselectAll() + PlatformHandleSelectionChanged()`, so the post-reload selection starts from a clean slate.

## Files changed

- `src-ui-wx/layout/LayoutPanel.cpp` — only

No new files, no project file updates needed for Windows / macOS / Linux.

## Test plan

- [x] Select a prop, Ctrl-C, Ctrl-V — copy lands clearly beside original, not overlapping
- [x] Press Ctrl-V repeatedly — each copy cascades further right, no stacking
- [x] Works for small props (arches) and large matrices alike — offset scales to each model's width
- [x] Only the newly-pasted copy is selected after paste (both in tree and preview)
- [ ] Windows / Linux smoke test — code is pure wx + std, no platform-specific APIs

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -7904,23 +7904,6 @@ void LayoutPanel::DoPaste(wxCommandEvent& event) {
 
                 if (nd)
                 {
-                    // Clear any existing selection ONLY after we've confirmed the
-                    // clipboard has valid paste data - otherwise a failed paste
-                    // (empty clipboard, bad XML) would wipe the user's selection
-                    // for nothing. See Copilot review on PR #6192.
-                    //
-                    // UnSelectAllModels() clears the per-Model Selected() flags
-                    // and our own tracking vectors. It does NOT clear the
-                    // wxTreeListCtrl widget's native selection though - on a
-                    // multi-select tree (wxTL_MULTIPLE) the old row lingers and
-                    // the post-reload SelectBaseObject->SelectModelInTree adds
-                    // the new row to it instead of replacing it. Hence we also
-                    // need UnSelectAllModelsInTree() to call
-                    // TreeListViewModels->UnselectAll() so only the new copy
-                    // ends up highlighted.
-                    UnSelectAllModels();
-                    UnSelectAllModelsInTree();
-
                     auto nda = DisplayAsTypeFromString(nd.attribute("DisplayAs").as_string());
                     auto nx = (int)nd.attribute("WorldPosX").as_double();
                     auto ny = (int)nd.attribute("WorldPosY").as_double();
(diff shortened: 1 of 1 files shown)


heffneil:
Addressed all three Copilot review findings in commit 7bb89da8e:

1. **Selection preserved on failed paste** - `UnSelectAllModels` / `UnSelectAllModelsInTree` moved inside the `if (nd)` block so they only run on the success path. Empty clipboard / bad XML / no element node no longer wipes the users selection for nothing.

2. **Collision loop scoped to Model pastes** - wrapped in `if (!copyData.IsViewObject())`. The loop compares against `xlights->AllModels` which is meaningful for model-vs-model placement but nonsensical for a ViewObject (Image / Mesh) paste. ViewObjects keep their existing `AddOffset(50, 0, 0)` nudge as before.

3. **Loop-guard trip is now visible** - if the 500-iteration safety cap is hit, we emit an `spdlog::warn` that includes the final X coordinate so the pathological case shows up in logs instead of silently pasting a potentially overlapping copy.

Build green on macOS Debug.

heffneil:
I think this is a worthwhile one, guys.  Instead of copying a prop or model over itself and also selecting the original, the new one is copied next to it, and the new one is selected itself.  The last 2 images demonstrate that if you pick a prop that has something adjecent it will move it to the next available area to the right



Screenshot 2026-04-29 at 10 57 10 AM

Screenshot 2026-04-29 at 10 57 18 AM

Screenshot 2026-04-29 at 10 57 22 AM

Screenshot 2026-04-29 at 10 57 25 AM


Screenshot 2026-04-29 at 10 57 35 AM

Screenshot 2026-04-29 at 10 57 39 AM

PR #6193, opened 2026-04-21
Remove leading comma in submodel selections 

AGFazio:
…when choosing nodes by dragging a box or mouse click.


derwin12:
Does this mess up the jacky models where he intentionally puts in the commas?

computergeek1507:
Im with Daryl, the commas are intentional to adjust/pad the size of the buffer

AGFazio:
No that I can see. I told it to leave that alone. When you drew a box, or clicked a single node on a new row it would add a comma in the beginning of line and add a node to the line count which wasn't wanted. If you intentionally add the commas or 0s it stays.

cybercop23:
It didn't use to add the comma, so we should target that change and fix the source of the problem.

AGFazio:
the add comma probaly came from when I added the new column. Intern looking at it now. Found if you intentionally add commas or 0s, then selected a range of nodes by dragging a box the placeholders would be removed...no bueno.

derwin12:
Tell the intern to review your commit (provide the url) to resolve the issue. See if it is a simpler fix.

derwin12:
Found the one liner for the fix .. not to point fingers, but traced it back to the utils commit where Scott (on the 6th) moved things around. Having the intern look at the commit it found the error in his intern's work.

cybercop23:
I think this can be closed, right?

AGFazio:
I believe Daryl made a different pr and merged that. This can be closed as I have a revised one that has less code if needed.

PR #6241, opened 2026-04-26
Add filter to Edit Display Elements Available list

heffneil:
## Summary

Adds a filter box above the **Available** list in *Edit Display Elements* (sequencer right-click → Edit Display Elements). Large shows can have hundreds of models/groups/timings, and scrolling to find the one you want to add to a view is tedious.

Same `wxSearchCtrl` pattern we already use in:
- ControllerModelDialog model filter
- Controllers tab visualizer filter (#6191)

so users get consistent UX across the app.

## Behaviour

- **Filter type:** case-insensitive substring match against the model/timing/group name.
- **Placeholder text:** "Filter models..."
- **Cancel button (×):** clears the filter.
- **Auto-clear on Add:** moving items into the view (Add Selected / Add All / drag-to-right / arrow-key right) clears the filter automatically — after moving, users almost always want the full list back to pick the next item. If they need the same filter, retyping is one click.
- **Scope:** Available list only, as agreed. Added list is unfiltered.

## Implementation notes

- All new UI/state lives **outside** the wxSmith generated block, so the `.wxs` file does not change.
- `ListCtrlNonModels` was at `GBPosition(1, 0) span(3, 1)`. We detach it, drop the search ctrl at `(1, 0)`, and re-attach the list at `(2, 0) span(2, 1)`. The growable row stays at row 3 so the list still fills the column.
- The filter is enforced inside `AddModelToNotList` / `AddTimingToNotList` (both gateways into the visible list), so any code path that calls `PopulateModels()` automatically respects the current filter without needing to know about it.
- New helper `IsFilteredOutOfNonModels(name)` does the comparison.

## Test plan

- [x] Build (Debug) macOS — clean compile, my files have zero warnings; only the pre-existing CoreML/SoundAnalysis link error which is unrelated and a separate Xcode-project regression.
- [x] Sequencer → right-click model → Edit Display Elements:
    - [x] Filter box appears above Available list with "Filter models..." placeholder
    - [x] Typing filters the list as you type (case-insensitive)
    - [x] × button clears
    - [x] Add Selected / Add All clears the filter and shows full list
    - [x] Drag to right clears the filter
    - [x] Right-arrow key on selected items clears the filter
- [ ] Verify on Windows/Linux — should work identically; `wxSearchCtrl` is cross-platform.
- [ ] Verify with views switching — `PopulateModels` is called on view switch, so the filter persists across views as long as the dialog is open.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -86,6 +86,9 @@ XLIGHTS/NUTCRACKER RELEASE NOTES
     -enh (dkulp)                Shader effect: dynamic uniforms emit JSON matching the effect-panel schema; JsonEffectPanel
                                 gains a reusable point2d control type, so iPad and desktop build the dynamic rows from the
                                 same description.
+    -enh (Neil)                 Edit Display Elements: add a filter box above the Available list so large shows can
+                                quickly find a model/timing/group to add. Filter clears automatically when items are
+                                moved into the view.
     -bug (dkulp)                Replaced throwing std::stoi calls in OutputManager / xxxSerialOutput / HinksPix with
                                 non-throwing std::strtol so corrupt config or controller responses no longer crash.
     -bug (dkulp)                State / Faces / Shockwave / VUMeter / Lyric / MatrixModel: added div-by-zero and
--- a/src-ui-wx/layout/ViewsModelsPanel.cpp
+++ b/src-ui-wx/layout/ViewsModelsPanel.cpp
@@ -254,18 +254,23 @@ ViewsModelsPanel::ViewsModelsPanel(xLightsFrame *frame, wxWindow* parent, wxWind
     _viewButtonsSizer = FlexGridSizer8;
 
     // Insert a filter control above the "Available" (non-models) list.
-    // The wxSmith block put ListCtrlNonModels at (1, 0) spanning rows 1-3.
-    // Detach it, drop the filter into row 1, and re-attach the list to
-    // rows 2-3 so the bottom row stays the growable one.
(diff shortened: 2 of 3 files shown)


computergeek1507:
I have worked on a lot of large shows over the last 10 years and I have never needed any of these search boxes....

heffneil:
i work on my large show everyday and its a useful feature.  Not sure why its a problem to incorporate a useful feature.  Takes up very little real estate and provides the end user a way to reduce the list to a manageable level to select the prop they are looking for.  I don't know if your shows are many props but when the list is greater than 25 it starts to become difficult and when its 50 its hard and by 100 its impossible.

heffneil:
Examples below:

Screenshot 2026-04-26 at 3 49 29 PM
Screenshot 2026-04-26 at 3 49 24 PM

MrPaulAR:
I also like the idea of a filter (without the need of clicking a button) anywhere that can have a long list of objects (display elements, groups in layout, import wizard, etc).  Even better if it supports regex.

jbs00210:
i like it and think its useful

wantmoore:
I wanted to chime in to support this. Filtering on long lists is one of those features that feels unnecessary right up until you have it, and then you can't imagine going back. Even folks who've gotten fast at scrolling and scanning end up reaching for it once it's there. Doesn't change anyone's existing workflow, just gives people another way in. Appreciate the effort on this!

jazakrzewski:
This would be nice to have, any list that can be bigger than the screen should have a filter option for good user experience

computergeek1507:
Can you fix the merge conflict and I will merge it

PR #6251, opened 2026-04-28
Fix Import Models from RGBEffects not honouring real-world dimensions (#6250)

### LINKED ISSUE #6250: Importing model from RGB effect broken
When you import a model from an rgb effect file. It will not honor the ruler on the layout you are importing to. This has been a known issue for quite some time. It didn't happen with any of the new releases.

heffneil:
## Summary

Fixes #6250 — *Import Models from RGBEffects* did not honour the Ruler/real-world dimensions of the source show. The bug had two halves; round-trip needs both to work.

### What was broken

**Writer side.** xLights' main save path (`xlights_rgbeffects.xml`) never emitted the portability hint `<dimensions units="…" width="…" height="…" depth="…"/>`. Single-model `.xmodel` exports did, but the full-show save did not. That's why reopening the same show always looked right — raw `ScaleX/Y/Z` and `ModelScreenLocation` *are* persisted and the real-world numbers shown in the model pane are derived from those under the show's Ruler — but cross-show import had no anchor in the destination's coordinate system.

**Reader side.** `XmlDeserializingModelFactory::Deserialize` already honoured `<dimensions>` and called `ApplyDimensions(...)`, but only when its caller passed `importing=true`. `ModelManager::CreateModel` hard-coded `importing=false`, so even on shows where the source XML *did* have `<dimensions>` (manually edited, or from `.xmodel` exports), the import path silently dropped the block.

### Fix

**Writer** — `src-core/XmlSerializer/BaseSerializingVisitor.cpp`
Hook `WriteDimensionsElement` into `WriteOtherElements`, the universal post-Visit hook every model type already calls (faces/states/dimming-curve/submodels/aliases). One line, covers every model type. The function self-no-ops when no Ruler is set, so shows that never configured one are unaffected.

**Reader** — `src-core/models/ModelManager.{h,cpp}`, `src-ui-wx/layout/LayoutPanel.cpp`
Add optional `importing = false` parameter to `CreateModel` and `createAndAddModel`. Plumb through to `DeserializeModel`. The two `LayoutPanel::ImportModelsFromPreview` call sites (one for models, one for groups) now pass `true`. All other callers — initial show load, base show folder merge — use the default `false` and stay byte-identical.

### Why same-show reopen always looked right

The underlying screen-space scale (`ScaleX/Y/Z`, `WorldPosX/Y/Z`, etc.) is persisted on every save. Real-world width/height/depth shown in the model pane is `(RenderWi · ScaleX) / pixelsPerUnit_from_Ruler` — derived on the fly. Same Ruler + same preview = same derived numbers. Cross-show, the inputs differ, so without the explicit `<dimensions>` hint the destination has no way to re-derive the right scale.

## Test plan

- [x] Built Debug, no warnings
- [x] Open source show with several mm-dimensioned models, save → confirmed 20 new `<dimensions>` child nodes appear in source `xlights_rgbeffects.xml`
- [x] Open destination show (with Ruler set), Layout → Import → Import Models from RGBEffects → pick source file → models land at source dimensions
- [x] Save/reopen of the same show is unchanged (regression check — raw `ScaleX/Y/Z` round-trip was already correct)
- [ ] Verify on Windows / Linux — pure C++/wx-free core path, no platform-specific code

## Out of scope

- Existing shows whose `xlights_rgbeffects.xml` predates this PR will start writing `<dimensions>` on the next save once a Ruler is set, but old archives remain anchorless until resaved.
- Real-world dimensions on `<modelGroup>` (groups don't carry physical sizes; the constituent models do).

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -14,6 +14,9 @@ XLIGHTS/NUTCRACKER RELEASE NOTES
     -enh (MrPierreB)            Add node animation playback to SubModels dialog.
     -enh (dkulp)                Media-compatibility "Convert Now" now special-cases animated GIFs that were used as
                                 Video effects and turns them into normal Pictures effects that properly handle animated GIFs
+    -bug (Neil)                 Import Models from RGBEffects: real-world dimensions now persist on save (xlights_rgbeffects.xml
+                                gains a <dimensions> child for every model when a Ruler is set) and are honoured on import,
+                                so models brought in from another show land at their original physical size (#6250)
     -bug (dkulp)                Replaced throwing std::stoi calls in OutputManager / xxxSerialOutput / HinksPix with
                                 non-throwing std::strtol so corrupt config or controller responses no longer crash.
     -bug (dkulp)                State / Faces / Shockwave / VUMeter / Lyric / MatrixModel: added div-by-zero and
--- a/src-core/XmlSerializer/BaseSerializingVisitor.cpp
+++ b/src-core/XmlSerializer/BaseSerializingVisitor.cpp
@@ -543,6 +543,14 @@ void BaseSerializingVisitor::WriteOtherElements(const Model* m) {
     WriteDimmingCurve(m);
     WriteAliases(m->GetAliases());
     WriteSubmodels(m);
+    // Persist real-world dimensions to xlights_rgbeffects.xml so a model's
+    // physical size (when a Ruler is set) survives a save/load and can be
+    // carried over by Import Models from RGBEffects. WriteDimensionsElement
+    // no-ops when no Ruler is set, so this is safe for shows that never
+    // configured a Ruler.
+    if (m != nullptr) {
+        WriteDimensionsElement(*m);
+    }
 }
 
 void BaseSerializingVisitor::WriteDimensionsElement(const Model& model) {
--- a/src-core/XmlSerializer/XmlSerializer.h
+++ b/src-core/XmlSerializer/XmlSerializer.h
@@ -160,10 +160,14 @@ struct XmlSerializer {
         docNode.append_attribute(XmlNodeKeys::TypeAttribute) = XmlNodeKeys::ExportedAttribute;
 
         XmlSerializingVisitor visitor{ docNode, includeGroups };
+        // The visitor's WriteOtherElements now emits <dimensions>
+        // directly inside the <model> node for every save path, so the
+        // separate XmlSerialize::AddDimensions call that .xmodel export
+        // used to make is no longer needed (and would produce a
+        // duplicate sibling <dimensions> child).
         model->Accept(visitor);
         if (includeGroups) {
             XmlSerialize::SerializeModelGroupsForModel(model, docNode);
-            XmlSerialize::AddDimensions(docNode, model);
         }
 
         return doc;
--- a/src-core/models/ModelManager.cpp
+++ b/src-core/models/ModelManager.cpp
@@ -1429,17 +1429,21 @@ Model* ModelManager::CreateDefaultModel(const std::string& type, const std::stri
     return model;
 }
 
-Model* ModelManager::CreateModel(pugi::xml_node node, int previewW, int previewH) const
+Model* ModelManager::CreateModel(pugi::xml_node node, int previewW, int previewH, bool importing) const
 {
     if (std::string_view(node.name()) == "modelGroup") {
         XmlDeserializingModelFactory factory;
-        return factory.Deserialize(node, const_cast<ModelManager&>(*this), false);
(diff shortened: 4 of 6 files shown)


computergeek1507:
Looks like a hammer looking for a nail

heffneil:
I am responding to feedback from users who expect a feature to behave in a certain (intuitive way) and making it work like that.  

I don't know if your comment is meant at "me" being the hammer, the approach of this fix, not entirely sure. But if you could be more specific I could comment.  Truth is claude is doing all of this so I can't really fix it without feedback thats more constructive other than involving hammer and nails and your experience with large light shows over 10 years. 

I hope you appreciate my attempt to help improve xlights and its not meant for anything other than that.

cybercop23:
Do all of the other ways of bringing in a model, download, import work as expected? If they do, that tells me there's a missing function when importing from an rgbeffect file. That's the fix. Adding a parameter to a function for when it is coming from a special place (Import from existing) doesn't seem the right fix. It is a fix, just not the right/optimal way. You need to review the change and make sure it makes sense, not that it just "works".

heffneil:
I don't really understand your comment. It doesn't work as you would expect - so its a bug.  Not really sure how to clarify what a bug is to you. 

Summary of the bug
When importing models from another show via Layout → Import → Import Models from RGBEffects, the destination ended up with default real-world dimensions instead of what was set in the source show. Two distinct things were wrong, both needed for the round-trip:

Bug 1 — the writer never ran
xLights has a function called WriteDimensionsElement that emits this XML:

<dimensions units="i" width="250" height="350" depth="..."/>
…but the only place it was hooked up was the single-model Export path (the .xmodel export when you right-click a model). The full-show save (xlights_rgbeffects.xml) never called it. So even shows where you'd carefully set physical dimensions never persisted those numbers — they only existed in memory while the show was open. Open the show next time and xLights re-derived dimensions from the model's screen-space size + the Ruler.

Bug 2 — the reader skipped the import
XmlDeserializingModelFactory::Deserialize does honor <dimensions> and call ApplyDimensions(...), but only when the caller passes importing=true. The import-from-previews path went through ModelManager::createAndAddModel → CreateModel, which hardcoded false for that flag. So even if <dimensions> had been there, it would have been ignored on import.

What we fixed
1. Persist dimensions on every save (BaseSerializingVisitor::WriteOtherElements)

Added a single line that calls WriteDimensionsElement(*m) in the universal post-Visit hook. Every model type uses this hook for faces/states/dimming-curve/submodels — adding dimensions there means it lands for all model types automatically. The function is no-op when the show has no Ruler set, so it's safe.

2. Thread importing through the import-from-previews path

Added an importing = false parameter to both ModelManager::CreateModel and ModelManager::createAndAddModel.
CreateModel now passes that through to XmlSerializer::DeserializeModel instead of hardcoding false.
The two call sites in LayoutPanel::ImportModelsFromPreview (one for models, one for groups) now pass true.
All other internal callers (initial show load, base-folder merge) keep the default false — their behaviour is unchanged.
Why the round-trip needed both
Old shows: source XML never had <dimensions> → import had nothing to copy. Bug 1 fixes the source side.
Even with <dimensions> present: import path ignored them. Bug 2 fixes the destination side.
You proved both worked in your test:

Resaving the donor show wrote 20 <dimensions> nodes into xlights_rgbeffects.xml (Bug 1 ✓)
Importing into the destination show pulled them in and applied them (Bug 2 ✓)

computergeek1507:
You(claude) describe this as a 'bug' but what if the user doen't want the model to use the source show folder dimensions.

heffneil:
What if a user imports a model from the catalog and doesn't want to use the source dimensions? Someone went to the trouble of sizing it in the other layout so I think they would expect it to persist between layouts.  The real world dimensions are the real world dimensions we aren't making this stuff up?

The referenced bug report shows that a user has over 200 models to import from another layout without persistence of size.  Resizing them manually (again) isn't an option.

cybercop23:
What we're questioning is... is this the right fix. You pasting answers from Claude to answer my question doesn't help matters either.

It was a simple question. The other ways of bringing in models, import & download, do they work as expected; mostly import? If so, then the real fix is making the import from a show work the same as import from a file, and not have to create a special way to handle imports. 

Clause will give you answers, YOU need to make sure they are correct.
Casein point:
image

heffneil:
Im not going to respond after this.  If you guys decide its not a worthwhile feature for a user who requested it then you can ignore it. To date every enhancement I have provided has been to benefit the user and you continue to shoot it down.  It's ok I hear you guys loud and clear and will no longer contribute.

derwin12:
The simpler fix is already implemented .. no need to update everyones rgbeffects file. It just looks at the ruler on the source and scales appropriately.

PR #6253, opened 2026-04-28
Fix vendor catalog Search regression after #6073 multi-select

heffneil:
## Summary

Fixes two regressions in the vendor model catalog Search after PR #6073 (multi-select catalog).

### 1. Search button perma-disabled on Windows
`ValidateWindow` was gating the Search button on `GetFocusedItem().IsOk()`. On the native Windows `wxTreeCtrl`, `GetFocusedItem()` returns invalid whenever keyboard focus is outside the tree — i.e. while the user is typing in the search textbox itself. Result: button stayed greyed out, search was effectively dead on Windows. macOS's generic implementation tracks focused item independently of keyboard focus, so the bug never surfaced there.

**Fix:** drop the focused-item gate. Only require non-empty search text and a populated tree.

### 2. `OnButton_SearchClick` bailed when no focused item
Even if Search was clickable, the search loop returned immediately when `GetFocusedItem()` was invalid (same Windows root cause).

**Fix:** when `GetFocusedItem()` is invalid, fall back to the first selected item, then to the root's first child, so the user can always kick off a search from the top of the tree.

### 3. Stepping through matches accumulated selection
PR #6073 switched the tree to `wxTR_MULTIPLE`. In that mode, `SelectItem()` *adds* to the selection set rather than replacing. Each Search step was silently piling matches into the multi-selection — searching for "pumpkin" and clicking Search repeatedly produced this:

[user's screenshot showing 6 highlighted pumpkin entries]

**Fix:** `UnselectAll()` before `SelectItem()` so each Search step shows exactly one match. This intentionally clears pre-existing user selection — Search is a navigation tool, not a multi-select builder. User can shift-click afterwards to build a download list from the search results.

## Test plan

- [x] macOS: type in search box → Search button enables. Type "pumpkin", click Search → first pumpkin highlights and scrolls into view. Click Search again → previous de-highlights, next pumpkin highlights. Confirmed by user.
- [ ] Windows: same flow. Was broken (button perma-disabled, search did nothing) before this PR.
- [ ] Linux: should match Windows behaviour since both use native wxTreeCtrl. No regression expected.

## Out of scope

The "Search ↔ multi-select for download" interaction has a residual mild UX wart: shift-clicking to build a multi-download from search hits requires the user to keep search results in mind manually. Could be worth a separate enhancement (e.g. "add all matches to selection" button) but that's a feature, not a regression.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,11 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.08  May ??, 2026
+    -bug (Neil)                 Vendor model catalog Search regression after #6073 multi-select: button stayed disabled
+                                while typing on Windows (GetFocusedItem() returns invalid when keyboard focus is in the
+                                search box), and stepping through matches added each one to the multi-selection instead
+                                of replacing it. Search button now only requires non-empty text + populated tree, and
+                                each step UnselectAll's before SelectItem so only the current match is highlighted.
     -enh (dkulp)                Linux: text rendering switched from wxGraphicsContext (Cairo+Pango) to a portable
                                 FreeType+HarfBuzz+Fontconfig backend in src-core/. Text and Shape effects can now
                                 render on background threads on Linux (previously forced to the main thread because
--- a/README.txt
+++ b/README.txt
@@ -15,7 +15,7 @@ XLIGHTS/NUTCRACKER RELEASE NOTES
                                 while typing on Windows (GetFocusedItem() returns invalid when keyboard focus is in the
                                 search box), and stepping through matches added each one to the multi-selection instead
                                 of replacing it. Search button now only requires non-empty text + populated tree, and
-                                each step UnselectAll's before SelectItem so only the current match is highlighted.
+                                each step calls UnselectAll before SelectItem so only the current match is highlighted.
     -enh (dkulp)                Linux: text rendering switched from wxGraphicsContext (Cairo+Pango) to a portable
                                 FreeType+HarfBuzz+Fontconfig backend in src-core/. Text and Shape effects can now
                                 render on background threads on Linux (previously forced to the main thread because
--- a/src-ui-wx/import_export/VendorModelDialog.cpp
+++ b/src-ui-wx/import_export/VendorModelDialog.cpp
@@ -1472,7 +1472,10 @@ void VendorModelDialog::OnTreeCtrl_NavigatorSelectionChanged(wxTreeEvent& event)
     // Because this code triggers a web download this function can be re-entered and this is not good
     busy = true;
 
-    wxTreeItemId startid = GetFocusedItem();
+    wxTreeItemId startid = event.GetItem();
+    if (!startid.IsOk()) {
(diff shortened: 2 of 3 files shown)


derwin12:
Windows .. still cant select the next item. I enter spinner, hit search. It finds the first one. Hit Search again .. still on the first item.

heffneil:
@derwin12 — pushed a follow-up that should fix the second-press behaviour you reported.

Root cause: after `UnselectAll()` + `SelectItem()` returns, keyboard focus on the Windows native `wxTreeCtrl` is on the Search button (not the tree), so `GetFocusedItem()` reports invalid, and `GetSelections()` after a programmatic `SelectItem` isn't reliable across the wxTR_MULTIPLE switch. The fallback chain landed back at the root's first child each time, and the loop's first advance just re-found the same first match.

Fix caches the last hit in a member (`_lastSearchHit`) so repeated Search clicks resume from where we left off, independent of platform tree-state quirks. The cache resets when the search text changes, so typing a new query starts fresh.

Verified on macOS — second/third/Nth press now advances. Windows artifact should be on the next CI run if you want to grab it: scroll to the latest "Windows xLights CI" check on this PR → Details → Artifacts → `xLights_Exe`. Appreciate the retest.

derwin12:
Now it goes to the next entry but it doesnt actually select it and hence show the prop on the right side. Confirmed it did that in the .01 release.

derwin12:
PS .. tell the intern to skip adding new comments to the code. It is adding more comments then code.

heffneil:
@derwin12 Pushed a fix. `OnTreeCtrl_NavigatorSelectionChanged` was reading the active item via `GetFocusedItem()` which returns invalid on Windows when focus is outside the tree (same root cause as the original Search-button issue). Switched to `event.GetItem()` so the right pane updates when Search programmatically selects.

Also — apologies on the verbose comments, will trim going forward.

heffneil:
@derwin12 — pushed three more rounds of fixes since your last report (commits `39e80844` through `11498399`):

- `OnTreeCtrl_NavigatorSelectionChanged` now reads `event.GetItem()` instead of `GetFocusedItem()` so the right pane updates when Search programmatically selects a model
- Search text is trimmed to match what `ValidateWindow` enables on
- User-manually-picked node (e.g. clicking a different vendor between searches) overrides the cached resume point
- Non-recursive `GetChildrenCount` for the empty-tree guard

Latest CI Windows artifact at https://github.com/xLightsSequencer/xLights/actions/runs/25123707052 if you have a moment to retest. Appreciate the patience.

heffneil:
Update — opened PR #6256 which adds a live whitespace-tokenized filter above the catalog tree. Tested on Windows: it both fixes the search-step issues this PR was chasing AND covers the wider 'huge catalog hard to navigate' UX in one shot. Filter narrows the tree as you type, hierarchy-aware so typing 'tree EFL' or 'boscoyo halloween' just works.

Suggest considering #6256 as the path forward and closing this one once #6256 lands. Happy to close this PR myself if maintainers prefer; leaving it open for now in case you want to use the small targeted fixes here as a stop-gap before #6256 merges.

derwin12:
Close this .. if you have a better one ...

heffneil:
This is replaced with a better filter coming in another PR

PR #6256, opened 2026-04-29
Add hierarchy-aware live filter to vendor model catalog

heffneil:
## Summary

Adds a live `wxSearchCtrl` filter input above the vendor model catalog tree (Layout → Import models from xlights.org) that **narrows** the tree as you type, instead of stepping through matches one at a time the way the existing Search button does.

## Why

The catalog has hundreds of nodes across a dozen vendors. Even after the multi-select work in #6073, finding what you want still meant scrolling past every non-matching vendor and category. Step-through Search is fine for "I know there's a pumpkin somewhere" but bad for "let me see all of Boscoyo's Christmas options."

## Behaviour

- **Single filter box, whitespace-tokenized AND-narrow.** Typing `tree EFL` is two terms — both must appear somewhere in the path. Replaces an earlier two-input prototype after feedback that one box with tokens is cleaner.
- **Hierarchy-aware match.** Each model's haystack is `Vendor / Category / Sub-category / ModelName` lower-cased, so typing a vendor or category name surfaces the entire sub-tree, not just leaves whose own names happen to contain the term.
- **Empty branches pruned.** A new bottom-up `PruneEmptyBranches` pass drops categories *and* vendors that have no surviving descendants. Suppressed vendors (e.g. "DMX Fixture Library", suppressed by default) stay visible when no filter is active so users can un-suppress them via the checkbox panel; under an active filter they get pruned along with everything else that doesn't match.
- **Auto-expand while filtering.** Categories already auto-expand (existing behaviour); vendors don't normally, but when a filter is active we `ExpandAll` so users see every survivor without re-expanding each vendor on every keystroke. Cleared filter returns to the default collapsed-vendor state.
- **Debounced 200 ms.** Full tree rebuild is O(N+M) and re-allocates `ModelElement` per visible node; collapsing fast keystrokes into one rebuild matters on big catalogs. Cancel button (×) rebuilds immediately — explicit user action shouldn't wait.
- **Search button untouched.** The existing step-through Search row at the bottom is preserved verbatim — power users who like next-match navigation still have it.

## Implementation notes

- All UI / state lives **outside** the wxSmith generated block; the `.wxs` file is unchanged.
- Filter row is non-growable; growable row marker is moved from row 0 → row 1 (the tree) so the tree fills vertically as before.
- Tree-build extracted into `RebuildTreeUI()` so it's callable on every filter change.
- `AddHierachy` and `AddModels` thread the lowercased ancestor path string down for the haystack match.
- `AddModels` early-returns on filter mismatch — every code path into the visible tree goes through it.
- Tokens parsed via `wxStringTokenizer` on space/tab; stored in a `std::vector<wxString>` (lower-cased).

## Test plan

- [x] Type "pumpkin" → all pumpkin models across vendors
- [x] Type "boscoyo pumpkin" → only Boscoyo's pumpkins (AND-narrow)
- [x] Type "halloween" → entire Halloween section per vendor
- [x] No filter → Suppressed vendors (DMX Fixture Library, DMX Model Library) still show so they can be un-suppressed
- [x] Filter active that excludes them → suppressed vendors hidden too
- [x] Cancel (×) → tree restores to original collapsed-vendor default
- [x] Original Search button at bottom still steps through matches
- [ ] CI Linux / Windows — pure wx APIs (wxSearchCtrl, wxTimer, wxStringTokenizer, wxFlexGridSizer, wxTreeCtrl), no platform-specific calls


computergeek1507:
I would add this as search modifiers on the current search box. i.e "pumpkin && boscoyo"  This would allow infinite length not just two

heffneil:
I understand, but && isn't very user-friendly or intuitive.  The second filter option lets you filter the results list.  In testing, it seems to do an admirable job of whittling the list down to a very very manageable level

computergeek1507:
I would also argue that two unlabeled search boxes is not very intuitive, windows explorer, finder, chrome, don't use double search boxes

cybercop23:
They are like mushrooms... now we'll have 3?

heffneil:
think of them as anchors eventually you'll have to come up for air

dkulp:
There are changes to controls and layouts in this dialog, but no changes to the wxs file which is used to actually define and layout of this dialog.   That's a big red flag to me and suggests that claude was  not following the instructions in the claude.md file.

heffneil:
Closing this PR. The hierarchy-aware filter works smoothly on macOS but Windows native wxTreeCtrl hangs during the rebuild for certain vendors (EFL Designs, Gilbert Engineering — possibly others with deep nesting or many multi-wiring models). After 20+ iterations and detailed per-step spdlog tracing the Windows hang point kept shifting — first the vendor Expand call, then post-rebuild paint sync, then queued selection events. We kept making progress on each one but always uncovered another platform-specific edge case. Without an interactive Windows debugger / repro on this side it's not solvable in this iteration. Filter runs beautifully on macOS. Reopening in the future when better Windows tooling is available.

heffneil:
Reopening — tested on macOS Debug and the Windows CI build (`xLights_Exe` artifact, current head `c1949da7b`). Filter behaves correctly on both:

- Catalog rebuild on every filter change completes cleanly (no more EFL hang on Windows after dropping the per-vendor / per-multi-wiring auto-Expand under filter)
- Hierarchy-aware matching narrows the tree as expected (typing `efl`, `gilbert`, `tree`, `arch`, `boscoyo` all work)
- Cancel button restores full tree
- Initial unfiltered build is unchanged

Full per-step `spdlog` tracing on Windows confirmed the rebuild reaches `RebuildTreeUI EXIT root_children=N tokens=M` without hanging on any of the previously-problematic vendors.

heffneil:
Bulk-resolved the prior Copilot review threads (27 of them). Most pointed at intermediate code paths from the debugging iteration on this branch — e.g. "both filter inputs" wording (only one input now), debug `spdlog::info` tracing (added during the Windows hang hunt), the static `bool rebuilding` guard (replaced by member `_treeRebuilding` + RAII `RebuildScope`), references to `SetFocus`-after-rebuild, `PruneEmptyBranches` recursive children-count, the deferred `_treeRebuilding` clear, etc. — all of which were superseded by subsequent commits leading to the current head `c1949da7b`. Anything still load-bearing is fine to flag again on a fresh review of the current diff.

heffneil:
Worth flagging: this PR took ~45 iterations to land cleanly. The hierarchy-aware filter logic itself was straightforward, but reaching a build that's stable on **both** macOS and Windows required a long debugging cycle — primarily chasing a Windows-only EFL hang in `wxTreeCtrl` triggered by auto-`Expand` calls during filter rebuilds, plus reentrancy between the debounce timer, `RebuildTreeUI`, and the selection-change handler that fires `DownloadImages`. The final shape (single filter input, RAII `RebuildScope` guard, member `_treeRebuilding` flag, no per-vendor/per-multi-wiring auto-Expand) is what survived that process.

heffneil:
@dkulp / @derwin12 — gentle bump on this one. Tested cleanly on macOS Debug + the Windows CI build. The 45-iteration debugging cycle turned up the EFL hang on Windows which is fully fixed at the current head. All Copilot threads resolved, CI green, mergeable. If anyone has bandwidth to give it another pass I'd really appreciate it — just trying to avoid the diff going stale against master.

heffneil:
Screenshot 2026-06-18 at 3 30 01 PM

heffneil:
guys - this is a worthwhile PR.  It really impoves finding appropriately models.  I would encourage you to test it yourself I really think it would be an improvement to the userbase.

dkulp:
Thanks for sticking with this — A few things need sorting out before this can go in, the biggest being how it interacts with the search control that's already in this dialog.

## 1. The existing search control — this needs a decision

Master already has a `wxSearchCtrl` here: `TextCtrl_Search` (`ID_TEXTCTRL3`), wxSmith-generated, at the bottom of `Panel3` (`VendorModelDialog.cpp:181-184`), bound at `:264-266` to `OnTextCtrl_SearchText`, `OnButton_SearchClick` and `OnSearchCancelClick`.

The important detail: #6267 ("Fix vendor model search on Windows") converted the old text-field-plus-button into a `wxSearchCtrl` and **deleted `Button_Search`**. `Button_Search` no longer exists anywhere in the `.cpp`/`.h`. The only surviving trigger for the step-through find-next (`OnButton_SearchClick`, `:1317`) is that control's built-in magnifier button.

So `TextCtrl_Search->Hide()` doesn't just tidy the layout — it makes step-through search **unreachable**. There's no accelerator, no F3, no menu item. `OnButton_SearchClick`, `OnSearchCancelClick`, `OnTextCtrl_SearchText`, `_lastSearchItem` and the bold-highlight/`EnsureVisible` machinery all become dead code (~100 lines).

That also means the PR description is no longer accurate:

> **Search button untouched.** The existing step-through Search row at the bottom is preserved verbatim — power users who like next-match navigation still have it.

True of an earlier commit, but `Hide bottom Search row; new live filter replaces it on all platforms` changed it.

Please pick one and drop the runtime `Hide()`:

- **(a) Repurpose in place — preferred.** Move `TextCtrl_Search` above the tree **in `src-ui-wx/wxsmith/VendorModelDialog.wxs`**, hang the debounce/live-filter logic off its existing `wxEVT_TEXT` binding, and keep `wxEVT_SEARCHCTRL_SEARCH_BTN` as next-match *within* the filtered tree. One control, both behaviours, and no hand-written sizer surgery.
- **(b) Remove it cleanly.** Delete `TextCtrl_Search`, its three handlers, `_lastSearchItem` and the `.wxs` entry outright, and say so in the description. Also fine — it drops ~100 lines.

Related: `VendorModelDialog.wxs:50` still declares `TextCtrl_Search` as a `wxTextCtrl` inside a `FlexGridSizer9` next to a `Button_Search`, none of which matches the generated `.cpp` any more — #6267 never updated the `.wxs`. That's pre-existing, but per AGENTS.md §4 those have to stay in sync, and option (a) is a natural place to fix it.

## 2. The description no longer matches the code

The body describes a hierarchy-preserving filter — `Vendor / Category / Sub-category / ModelName`, `PruneEmptyBranches` dropping categories *and* vendors, typing a category name surfacing a sub-tree. The code deletes `AddHierachy`, `AddModels` and `DeleteEmptyCategories` and builds a two-level Vendor → Model tree. No category rows, no wiring child nodes.

To be clear, I think the flattening is defensible on its merits:

- It converges with the iPad, which already browses flat (`src-iPad/App/VendorBrowserSheet.swift` — vendor list + `.searchable`, model list + `.searchable`). Both clients already share `src-core/import_export/VendorCatalog.h`.
- Iterating `vendor->_models` really does de-dup models listed in multiple categories, and it surfaces models with no category at all, which the old `GetModels(categoryId)` walk could never show.
- Dropping the parent Model node for multi-wiring models loses nothing — `Model` pushes its images into every wiring (`VendorCatalog.cpp:207-209`) and `ModelWiring::GetDescription()` prepends the model description (`:155-164`). I checked specifically; it's fine.

But it needs to be presented as the UX change it is, and it has some rough edges:

- With no filter typed, a vendor expands to a flat **unsorted** list of hundreds of models in raw XML order. Please sort leaves case-insensitively by `_name`.
- `PruneEmptyBranches` deletes **Vendor** nodes with zero children even with no filter active (only *suppressed* vendors are exempt). `DeleteEmptyCategories` never touched Vendor nodes. A non-suppressed vendor whose inventory parses to zero models now silently vanishes — along with the only way to reach its contact/website panel.
- The `type == "Category"` branch in `PruneEmptyBranches` is now dead, as is the "user selects both a Model node and its child Wiring node" dedup in `GetSelectedWirings()`.

Easiest path: land the flattening as its own PR with an accurate description, and keep this one to the filter.

## 3. Debug tracing needs removing

There are ~15 `spdlog::info` calls left from the `Add spdlog tracing for filter rebuild diagnosis` / `Add per-step tracing to find rebuild hang point` commits — `"VMD::RTUI step 1: UnselectAll"`, `"VMD::RTUI vendor[{}] '{}' begin"`, etc. That's per-vendor, per-rebuild, per keystroke-batch, at `info` level. Anything worth keeping should be `spdlog::debug`, matching the rest of the file (`:387`, `:402`, `:607`).

## 4. Correctness

1. **Reading a deleted tree item is undefined behaviour, not a null check.** In `OnTreeCtrl_NavigatorSelectionChanged`, `TreeCtrl_Navigator->GetItemData(startid) == nullptr` on an id from a deleted generation reads freed memory — the comment says as much ("GetItemData returns null for those on Windows"). Same for `GetItemData(_lastSearchItem)`. Use a generation counter bumped in `RebuildTreeUI` and drop events stamped with an older generation, or lean on `_treeRebuilding` + `CallAfter`.
2. **Removing the `if (!startid.IsOk()) startid = GetFocusedItem();` fallback** changes behaviour introduced for multi-select in #6073, where it was added deliberately. Worth verifying ctrl-click still updates the detail panel.
3. **`static bool busy` in `OnTreeCtrl_NavigatorSelectionChanged`** is a function-local static (shared across instances, not exception-safe) and redundant — the only non-trivial thing it calls is `UpdatePanelForItem`, which already has the identical guard. If a guard really is needed, make it a member with an RAII setter like your own `RebuildScope`.
4. **Hardcoded generated-layout indices.** `Insert(0, …)` / `RemoveGrowableRow(0)` / `AddGrowableRow(1)` silently marks the wrong row growable if wxSmith ever adds a control to `Panel3`. Option (a) above makes this go away.
5. **`_initialBuild` can stick.** It's only cleared inside `if (_initialBuild && first.IsOk() && first != root)`. A first build with no vendors leaves it `true`, so the next filter rebuild does an `EnsureVisible` that yanks scroll position. Clear it unconditionally at the end of the first rebuild.
6. **Ordering in `LoadTree`.** `RebuildTreeUI()` now runs *before* the `_vendors.size() == 0` check, so the zero-vendor path builds and prunes an empty tree before erroring. Swap them.
7. **`wxTE_PROCESS_ENTER` with no `wxEVT_TEXT_ENTER` handler** swallows Enter in the dialog, where it would otherwise reach the default button. Either bind it to force an immediate rebuild (nice, and consistent with your cancel-button rationale) or drop the style.
8. **Unconditional `SetFocus()` on every debounce fire.** Type, then click a tree node within 200 ms, and focus is yanked back to the filter. Guard with `if (FindFocus() == TextCtrl_Filter)`.

## 5. Performance — the haystack should be built once

Right now the match data is rebuilt from scratch on every rebuild:

- `AddVendorModelList` allocates a fresh `unordered_map<std::string, MVendorCategory*>` per vendor per rebuild, via a `std::function` recursive lambda.
- `ModelMatchesFilter` calls `Category::GetPath()` per model per category, and that recurses the parent chain concatenating `std::string`s each time (`VendorCatalog.cpp:367-370`).
- `CatalogFilterMatchesPath` then constructs and `MakeLower()`s a fresh `wxString` on *every token check*.

Compute one lower-cased haystack per model **once** after load (`"vendor / cat/subcat / model"`, joined across categories for multi-category models) into an `unordered_map<const MModel*, std::string>`. Filtering collapses to N substring tests. That removes the map rebuild, the `GetPath()` recursion and the `wxString` churn, lets you drop the new `<functional>`/`<unordered_map>` includes, and would probably let you shorten the 200 ms debounce.

Also, `IsVendorSuppressed` does a `config->Read` per call (`:1269-1276`) and is now called once per vendor per rebuild, plus again per empty vendor in `PruneEmptyBranches`. Hoist the suppress list into a `std::set` once per rebuild.

## 6. Style (AGENTS.md §6)

- §6 asks for comments only where the *why* is non-obvious. Several new blocks narrate what the code does or reference PR history ("Master's #6267 added a wxSearchCtrl…", "Experimental catalog filter") — that belongs in commit messages. Do keep the Windows `WM_SETFOCUS`/caret explanation and the `CallAfter`-after-`Thaw` note; those are exactly the kind of comment that earns its place.
- Manual `new wxTimer` + `delete` in the destructor — prefer a by-value member (`wxTimer _t;` plus `_t.SetOwner(this)` in the ctor) or a `std::unique_ptr`.


## Suggested order of work

1. Strip the `spdlog::info` tracing.
2. Decide (a) or (b) on the duplicate search control; no runtime `Hide()`.
3. Precompute the per-model haystack once at load.
4. Fix the stale-`wxTreeItemId` UB, the `_initialBuild` latch and the `LoadTree` ordering.
5. Split the Vendor → Model flattening into its own PR, with sorting and a guard for model-less vendors.

## BIG Request

I would strongly suggest closing the PR, squashing all the changes down into a single commit/change, and resubmitting as a new PR.   The 33 commits for changes in just two files is definitely confusing and as noted above, has caused descriptions and such to become out of date and/or inaccurate.

heffneil:
Thanks @dkulp for the thorough review — everything you raised is addressed in a fresh single-commit replacement: #6733. Per your suggestion this PR is closed in its favor. The new one repurposes the existing search control in place (option a, .wxs updated and re-synced), presents the Vendor → Model flattening honestly as the UX change it is (with sorting and the empty-vendor guard), precomputes the per-model search haystack at load, strips all the diagnostic tracing, and fixes the stale-item/UB, _initialBuild, LoadTree-ordering, Enter, and focus issues.

PR #6259, opened 2026-04-29
3D layout: click another model to switch selection (parity with 2D)

heffneil:
## Summary

In the 3D layout view, clicking a different model when one was already selected did **nothing** — the user had to first hit Esc, or double-click empty space (Mac) / single-click empty space (Windows) to drop the current selection, *then* click the new model. The 2D path has always done click-to-switch via `UnSelectAllModelsInTree + SelectSingleModel`; 3D never grew the equivalent.

This PR brings 3D into parity.

## Why it didn't already work in 3D

- `ProcessLeftMouseClick3D`, when `selectionLatched` is true, only checks if the click landed on the **current** selection's handles. If not, it sets `m_mouse_down = true` and waits for a drag — never considers the click might be on a different model.
- We can't lean on `highlightedBaseObject` either: `OnPreviewMouseMove3D` intentionally does NOT update `highlightedBaseObject` while a selection is latched (to keep visual noise down — see the *"require control to be active"* comment in that handler), so it always reads back as `selectedBaseObject` in the latched state.

## Fix

When `selectionLatched` and the click misses the current selection's handles, do an explicit `HitTest3D` against all models (or view objects, depending on `editing_models`) and:

| Input | Behaviour |
|---|---|
| Plain click on a **different** model | Drop the previous selection via `UnSelectAllModelsInTree` / `UnSelectAllModels` and select the new one. Mirrors the 2D path. |
| **Cmd / Ctrl-click** on a different model | Keep the previous selection in the `GroupSelected` set and add the new one as the active latched selection. Mirrors Ctrl-click multi-select in 2D. |
| Click on **empty space** (no hit) | Falls through to the original `m_mouse_down = true` path so lasso-drag and other modes still work. |
| Click on the **current selection's handles** | Unchanged — still routes through the existing handle-interaction code. |

Behaviour on the *non-latched* path is also unchanged: first click on any model still selects + latches it as before.

## Test plan

- [x] macOS, 3D mode:
    - [x] Click model A → click model B → A deselects, B becomes active
    - [x] Click A → Cmd-click B → both selected, B is active
    - [x] Cmd-click C → A, B, C all selected
    - [x] Click empty space → all deselect (lasso start)
    - [x] Click A's resize handle (not body) → drags handle as before
- [ ] Windows, 3D mode — should behave the same since `event.ControlDown()` covers Win Ctrl
- [ ] 2D mode — unchanged code path, not exercised by this PR
- [ ] Linux 3D — same code path as Windows

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,10 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -bug (Neil)                 3D layout: clicking a different model now switches the selection directly instead of
+                                requiring the user to first click empty space (or hit Esc) to deselect. Cmd / Ctrl-click
+                                still adds to the selection for multi-select. 2D path was already correct; this brings
+                                3D into parity.
     -bug (scott)                Fix HTDemucs ONNX model download failing: the 12-second total curl timeout set for
                                 short API calls was also killing the large-file download before it could complete.
     -bug (dkulp)                Backup: switch the per-file copy from wxCopyFile to std::filesystem::copy_file so
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -3693,20 +3693,6 @@ void LayoutPanel::ProcessLeftMouseClick3D(wxMouseEvent& event)
                 GetMouseLocation(event.GetX(), event.GetY(), ray_origin, ray_direction);
                 selectedBaseObject->GetBaseObjectScreenLocation().CheckIfOverHandles3D(ray_origin, ray_direction, m_over_handle, modelPreview->GetCameraZoomForHandles(), modelPreview->GetHandleScale());
             }
-            // If the click isn't on the current selection's handles, hit-test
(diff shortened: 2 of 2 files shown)


o0charlie0o:
Ah yes! Nice one @heffneil this has also bugged me but never thought to fix it...

derwin12:
Can we remove the 50 lines of comments?

heffneil:
@derwin12 just looked again — this diff is **66 lines of code in `LayoutPanel.cpp` with no inline comments added** (only a 4-line README release note). I count zero `//` comment lines in the additions. Can you point me to which lines you meant? Happy to trim if I'm missing something.

Also rebased onto current master to clear the merge conflict — branch is now mergeable, CI re-running on the new head.

dkulp:
I believe the latest changes to the Layout stuff addressed this.  All of the mouse/click/drag handling has been completely re-done and I believe this is now fixed as part of that.   Can you verify?

heffneil:
If you select a prop, and then click on another while the first one is yellow you still must hit escape or double click on the background to release the original selection.

cybercop23:
Confirmed on TestFlight. It now allows you to just click from model to model like 2D does.

heffneil:
Post .8? If so - great news!

derwin12:
(Dan did a layout change post .08)

heffneil:
I wouldn't know.

PR #6284, opened 2026-05-02
Layout Zoom Fix

AGFazio:
Finally pixel size is proportionate to the zoom.


derwin12:
This is just going back to the code that was there 2 versions ago..

Issue #6290, opened 2026-05-03
Add aliasing to dialog when opening old sequence


AGFazio:
Add option to save alias when opening an old sequence and you are asked to rename model or submodel

AGFazio:
PR 6296

PR #6296, opened 2026-05-03
#6290 Add alias when opening old sequence

AGFazio:
Add to submodel select dialog as well


derwin12:
Why would this be a whole new dialog? shouldnt it just be a check box added to the existing one?

AGFazio:
not new. Same ones that were there, just added aliasing to the the submodel one.

PR #6311, opened 2026-05-06
Add ctrls tab to Layout so you can select models by ctrl or ports

cybercop23:
Can easily see multi-port/string models. Also supports shadow models

https://github.com/user-attachments/assets/af66743c-4864-4b9d-8f2d-26cab69b8a32


cybercop23:
Will update this after the new controller tab redesign work is complete.

PR #6318, opened 2026-05-08
macOS: open .xsqz in a new xLights instance instead of replacing the current show

heffneil:
## Summary

On Windows, double-clicking a `.xsqz` launches a fresh `xLights.exe` process with the package extracted to a temp folder as the show — leaving any other running xLights instances untouched. On macOS, the running instance hijacked itself in-place via `MacOpenFiles`, replacing whatever the user had loaded with the package's contents. There was no way to open an `.xsqz` on macOS without losing your current sequence.

This PR brings macOS in line with the Windows behavior.

## Changes

**`src-ui-wx/xLightsApp.cpp` — `MacOpenFiles()`**
When the file is a `.xsqz` / `.zip` and a sequence is already loaded in this instance, spawn a separate xLights process via `/usr/bin/open -n -a <bundle> --args <file>` and return early. The new process starts cold and the existing argv path (lines 705-811) extracts the package and opens it in read-only mode — same flow as Windows.
- Cold start (no sequence loaded yet) still uses the existing in-place path, which is the correct behavior for the very first `.xsqz` open after launch.
- Bundle path is resolved from `wxStandardPaths::Get().GetExecutablePath()` so the spawned sibling is the same build the user is currently running (Debug spawns Debug, release spawns release).

**`src-ui-wx/xLightsMain.cpp` / `.h` — File menu**
- The pre-existing macOS-only **File → Open New xLights Instance** item now uses the running bundle path instead of hard-coding `/Applications/xLights.app`. Same fix as above applied to manual launches.
- Adds **File → Open New xLights Instance From File...** — file picker filtered to `.xsqz / .zip / .xsq` that spawns a new instance with the chosen file as argv. Useful when the user wants explicit control over which file goes to a sibling process without involving Finder file associations.

## Behavior matrix after this change

| Scenario | Before | After |
|---|---|---|
| Cold-launch xLights via Finder double-click on .xsqz | Loads xsqz in-place | Loads xsqz in-place (unchanged) |
| Already running, double-click another .xsqz in Finder | Replaces current sequence | Spawns new instance |
| File menu → Open New xLights Instance | Spawns /Applications/xLights.app | Spawns the running bundle |
| File menu → Open New xLights Instance From File... (new) | n/a | Pick file, spawn new instance |

## Test plan

- [x] macOS Debug build: with a sequence loaded, opening another .xsqz spawns a sibling instance, both windows independent
- [x] macOS Debug build: cold-launch via `open -a <Debug bundle> file.xsqz` still loads in-place
- [x] macOS Debug build: File → Open New xLights Instance From File... picks .xsqz and launches sibling
- [x] macOS Debug build: File → Open New xLights Instance launches an empty sibling
- [ ] Windows: untouched (changes are inside `#ifdef __WXOSX__` and `MacOpenFiles` is macOS-only)
- [ ] Linux: untouched (same reason)

A screenshot of the new menu items will be attached as a comment.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,12 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -bug (Neil)                 macOS: opening a .xsqz package while a sequence was already loaded replaced the
+                                current show folder in-place instead of launching a separate instance like Windows
+                                does. Now spawns a new xLights process for the package via `open -n`, leaving the
+                                current sequence untouched. Also adds File menu entries "Open New xLights Instance"
+                                (now uses the running bundle path) and "Open New xLights Instance From File..." for
+                                explicit second-instance launches.
 
 2026.08  May 7, 2026
     -enh (dkulp)                When a JobPool worker thread dies from an unhandled C++ exception, the log now
--- a/src-ui-wx/xLightsApp.cpp
+++ b/src-ui-wx/xLightsApp.cpp
@@ -146,6 +146,62 @@
 
 xLightsFrame* xLightsApp::__frame = nullptr;
 
+#ifdef __WXOSX__
+// Resolve the .app bundle path of the currently-running xLights process by
+// walking up from the executable. Returns empty if the layout doesn't match
+// the expected `<Bundle>.app/Contents/MacOS/<exe>` shape (e.g. unbundled dev
+// run). Callers must check for empty before using the result.
+wxString GetRunningAppBundlePath() {
+    wxString exePath = wxStandardPaths::Get().GetExecutablePath();
+    wxFileName fn(exePath);
+    if (fn.GetDirCount() < 2) {
+        spdlog::warn("Cannot resolve .app bundle: executable path too shallow: {}",
+                     exePath.ToStdString());
+        return wxEmptyString;
+    }
+    fn.RemoveLastDir(); // strip MacOS
+    fn.RemoveLastDir(); // strip Contents
+    wxString candidate = fn.GetPath();
+    if (!candidate.EndsWith(".app")) {
+        spdlog::warn("Cannot resolve .app bundle from executable path '{}' (got '{}')",
+                     exePath.ToStdString(), candidate.ToStdString());
+        return wxEmptyString;
+    }
+    return candidate;
+}
+
+// Spawn a new xLights instance using `open -n -a <bundle> [--args <file>]`.
+// Uses the argv form of wxExecute to avoid shell quoting / escaping issues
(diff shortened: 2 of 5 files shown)


heffneil:
Screenshot 2026-05-08 at 8 56 28 AM
Menu Option at the bottom.  If you think there is better wording feel free to change it.

cybercop23:
I don't think we want yet another menu option. The parity should be that when you open it from Finder, it should "know" that and start a new instance with that xsqz

heffneil:
I will work that way as well.  This is a file option which we never really had.  I agree that too many menu options is overwhelming, like the 100 effect buttons that only 20 are ever used. I have a large screen so I could be biased that the additional line doesn't really take up a lot of space.  We can work on cleaning up the File menu - I think theres too many unused things that could be nested?

cybercop23:
Not the same thing. Each effect does something totally different. That's why one should use more key bindings.

heffneil:
Lets change Backup to be a submenu if you're really worried about all the items in the menu.  We should have backup and then within the submenu you can have Back Up Now, Restore Backup, and Backup to Alternate Location. That will save 2 lines so we can add! I would even be open to making a Open menu with a submenu for Another Instance and Another Instance from File. The Import menu is one option.  It should be Import Export and we can move stuff there including the Export House Preview to reduce the number in the File section. And just for the record the Tool menu is much longer than File.  Not sure if I have proven that your argument is wrong.  But it clearly is.  Not to air dirty laundry Alex but you can either accept my improvement or not.  I don't need to be in the club but sometimes you just need to be open minded if I am NOT in the club.

heffneil:
Addressed all five Copilot threads in bdcd100e8:

- **Shell quoting (xLightsApp.cpp:470, xLightsMain.cpp:5456)**: switched both spawn sites from `wxExecute(wxString)` (shell-parsed on Unix) to `wxExecute(char**)` argv form. No more quoting issues for paths with spaces, quotes, or other shell metacharacters.
- **Brittle bundle path stripping (xLightsApp.cpp:468, xLightsMain.cpp:5435)**: extracted into `GetRunningAppBundlePath()` which walks up via `wxFileName::RemoveLastDir` and confirms the result ends in `.app`. Returns empty + logs a warning for unbundled / dev-run layouts; callers fall back gracefully (in-place handling for `MacOpenFiles`, error message for the menu items).
- **Ignored `ObtainAccessToURL` return value (xLightsMain.cpp:5450)**: now honored; shows an error dialog and returns if access is denied so we don't spawn a sibling that can't read the file.

Also centralized all three spawn call sites behind one helper `SpawnNewXLightsInstance(fileToOpen)` in `xLightsApp.cpp` so future callers don't re-introduce the same issues.

heffneil:
Superseded by #6319, which drops the redundant "Open New xLights Instance From File..." menu item. Same bug fix, smaller surface area.

PR #6319, opened 2026-05-08
macOS: open .xsqz in a new xLights instance instead of replacing the current show

heffneil:
## Summary

On Windows, double-clicking a `.xsqz` launches a fresh `xLights.exe` process with the package extracted to a temp folder as the show — leaving any other running xLights instances untouched. On macOS, the running instance hijacked itself in-place via `MacOpenFiles`, replacing whatever the user had loaded with the package's contents. There was no way to open an `.xsqz` on macOS without losing your current sequence.

This PR brings macOS in line with the Windows behavior.

(Replaces #6318. Same fix, but without the extra "Open New xLights Instance From File..." menu item — the double-click flow is sufficient and cleaner.)

## Changes

**`src-ui-wx/xLightsApp.cpp` — `MacOpenFiles()`**
When the file is a `.xsqz` / `.zip` and a sequence is already loaded in this instance, spawn a separate xLights process via `/usr/bin/open -n -a <bundle> --args <file>` and return early. The new process starts cold and the existing argv path (lines 705-811) extracts the package and opens it in read-only mode — same flow as Windows.
- Cold start (no sequence loaded yet) still uses the existing in-place path, which is the correct behavior for the very first `.xsqz` open after launch.
- Bundle path resolved via `wxFileName` walking up from `wxStandardPaths::GetExecutablePath()` and validating the `.app` suffix; returns empty (with logged warning) if the executable isn't inside an app bundle. Callers handle the empty case gracefully.
- Spawn logic uses `wxExecute(char**)` argv form, not the string form, so paths with spaces / quotes / shell metacharacters are safe.
- Centralized in `SpawnNewXLightsInstance()` so future callers share one implementation.

**`src-ui-wx/xLightsMain.cpp` — pre-existing menu item**
The pre-existing macOS-only **File → Open New xLights Instance** item now uses the running bundle path instead of hard-coding `/Applications/xLights.app`. Same fix as above applied to manual launches; Debug builds now spawn Debug builds.

## Behavior matrix after this change

| Scenario | Before | After |
|---|---|---|
| Cold-launch xLights via Finder double-click on .xsqz | Loads xsqz in-place | Loads xsqz in-place (unchanged) |
| Already running, double-click another .xsqz in Finder | Replaces current sequence | Spawns new instance |
| File menu → Open New xLights Instance | Spawns /Applications/xLights.app | Spawns the running bundle |

## Test plan

- [x] macOS Debug build: with a sequence loaded, opening another .xsqz spawns a sibling instance, both windows independent
- [x] macOS Debug build: cold-launch via `open -a <Debug bundle> file.xsqz` still loads in-place
- [x] macOS Debug build: File → Open New xLights Instance launches an empty sibling
- [ ] Windows: untouched (changes are inside `#ifdef __WXOSX__` and `MacOpenFiles` is macOS-only)
- [ ] Linux: untouched (same reason)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,11 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -bug (Neil)                 macOS: opening a .xsqz package while a sequence was already loaded replaced the
+                                current show folder in-place instead of launching a separate instance like Windows
+                                does. Now spawns a new xLights process for the package via `open -n`, leaving the
+                                current sequence untouched. The pre-existing File menu "Open New xLights Instance"
+                                also now uses the running bundle path so Debug builds spawn Debug builds.
 
 2026.08  May 7, 2026
     -enh (dkulp)                When a JobPool worker thread dies from an unhandled C++ exception, the log now
--- a/src-ui-wx/xLightsApp.cpp
+++ b/src-ui-wx/xLightsApp.cpp
@@ -187,16 +187,23 @@ bool SpawnNewXLightsInstance(const wxString& fileToOpen) {
     std::string sArgs = "--args";
     std::string sFile(fileToOpen.ToUTF8().data());
 
+    long pid = 0;
     if (fileToOpen.IsEmpty()) {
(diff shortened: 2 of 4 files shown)


heffneil:
Addressed both Copilot threads in 1dc854e39: `SpawnNewXLightsInstance` now captures `wxExecute`'s return value and returns false when it's 0 (async-launch failure). Callers fall back to in-place handling in `MacOpenFiles` and surface an error dialog from the menu handler, instead of silently doing nothing.

PR #6330, opened 2026-05-10
Add File menu "Open New xLights Instance From File..." (macOS + Windows)

heffneil:
## Summary

Adds a new **File → Open New xLights Instance From File...** menu item on macOS and Windows. Picks a sequence package (`.xsqz` / `.zip` / `.xsq`) via a file dialog and launches a new xLights process to open it.

## Why

On both macOS and Windows, `.zip` files are associated with the OS's archive utility (Archive Utility on macOS, File Explorer's built-in extractor on Windows), not with xLights. When a user receives a zipped sequence from a collaborator — which is the most common form they're shared in — they currently have to either:

1. Rename `.zip` to `.xsqz` to make double-click work, or
2. Extract the zip manually and open the contents through xLights

Neither is great. This menu option lets the user open any zipped or unzipped sequence file directly from inside xLights without having to mess with file associations or rename anything. The chosen file is passed as argv to a new xLights process so the existing launch-time argv path handles extraction exactly like a Finder / Explorer double-click would for `.xsqz`.

## How it works

The new instance always opens the package in **read-only** mode (matches existing `.xsqz` behavior — keeps the original archive intact). The current xLights instance is unaffected; user's open sequence is preserved.

| Platform | Spawn mechanism |
|---|---|
| macOS | `/usr/bin/open -n -a <bundle> --args <file>` (via the existing `SpawnNewXLightsInstance` helper added in #6319) |
| Windows | `wxExecute("xLights.exe <file>")` — xLights has no single-instance checker on Windows so re-running the same executable produces an independent process |

Both platforms use the `wxExecute(char**)` argv form so paths with spaces, quotes, or other shell metacharacters are handled safely.

## Changes

- **`src-ui-wx/xLightsApp.cpp`**: adds a Windows variant of `SpawnNewXLightsInstance()`. The macOS path is unchanged.
- **`src-ui-wx/xLightsApp.h`**: widens the declaration of `SpawnNewXLightsInstance` from `__WXOSX__` to `__WXOSX__ || __WXMSW__`.
- **`src-ui-wx/xLightsMain.cpp` / `.h`**: adds `OnMenuItem_File_NewXLightsInstanceFromFile` handler and wires it to a new menu entry. Both gated on `__WXOSX__ || __WXMSW__`. Honors `ObtainAccessToURL` return value on macOS so we don't spawn a sibling that can't read the file.
- **`README.txt`**: 2026.09 release note.

Linux is not currently supported because there's no `SpawnNewXLightsInstance` implementation for it. Could be added later (probably just `wxExecute(GetExecutablePath())` like Windows) but kept out of scope for this PR.

## Test plan

- [x] **macOS Debug**: File menu shows the new item; pick a `.zip` sequence package → new xLights instance launches and opens the package read-only; original instance unaffected
- [x] **macOS Debug**: pick a `.xsqz` → same flow works
- [x] **macOS Debug**: pick a plain `.xsq` → new instance opens with that file's directory as show folder
- [x] **macOS Debug**: cancel the file dialog → no spawn, no error
- [ ] **Windows**: same matrix on a Windows install (couldn't test locally — would appreciate Windows verification)
- [ ] **Linux**: untouched (gated out)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,11 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -enh (Neil)                 Add File menu "Open New xLights Instance From File..." (macOS + Windows). Picks an
+                                .xsqz / .zip / .xsq via a file dialog and launches a new xLights process to open it.
+                                Lets users open zipped sequence packages without renaming .zip to .xsqz first or
+                                relying on OS file associations. The new instance handles the file via the existing
+                                argv launch path so behavior matches Finder / Explorer double-click.
     -bug (scott)                Fix HTDemucs ONNX model download failing: the 12-second total curl timeout set for
                                 short API calls was also killing the large-file download before it could complete.
     -bug (dkulp)                Backup: switch the per-file copy from wxCopyFile to std::filesystem::copy_file so
--- a/src-ui-wx/xLightsApp.cpp
+++ b/src-ui-wx/xLightsApp.cpp
@@ -209,6 +209,39 @@ bool SpawnNewXLightsInstance(const wxString& fileToOpen) {
 }
 #endif
 
+#ifdef __WXMSW__
+// Spawn a new xLights.exe process, optionally passing a file path as the first
+// argv. xLights has no single-instance checker so simply re-running the same
+// executable produces an independent process. Uses the argv form of wxExecute
+// to avoid shell quoting issues for paths with spaces or other special chars.
+bool SpawnNewXLightsInstance(const wxString& fileToOpen) {
+    wxString exePath = wxStandardPaths::Get().GetExecutablePath();
+    if (exePath.IsEmpty()) {
+        spdlog::warn("Cannot resolve xLights executable path; new instance not spawned");
+        return false;
+    }
+    std::string sExe(exePath.ToUTF8().data());
+    std::string sFile(fileToOpen.ToUTF8().data());
+
+    long pid = 0;
+    if (fileToOpen.IsEmpty()) {
+        char* argv[] = { sExe.data(), nullptr };
+        spdlog::info("Spawning new xLights instance: '{}'", sExe);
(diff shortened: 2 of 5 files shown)


computergeek1507:
Looks like another hammer looking for a nail. I don't understand why the user would want to do this, maybe to preview a packaged sequence?

heffneil:
Precisely.  Open a purchased sequence to look at the effects you want to import.  In our most recent zoom room gems session people talked about opening multiple instances.  This wasn't an option in mac until I added the one feature to double click on the xsqz file but if you want to open a .zip you really can't.

computergeek1507:
On windows you can right click on a zip and select open with and select xLights

heffneil:
That is wonderful! On a Mac can you tell me how to do that?

computergeek1507:
I'll pass, I'm not a huge fan of facetious remarks and somewhat passive aggressiveness.

heffneil:
Well it was a genuine question. Since you think this is a hammer or a nail. I get confused which is which I want to know how to work around the problem I propose we solve here.

dkulp:
Right click -> "Open With..." -> Select "All Applications" -> Select xLIghts.  

That said, it's almost always quicker and easier to either rename the file to xsqz (if you intend to keep it compressed) or double click on it to expand it and then open the actual xsq.     In general, the xsqz option is the better one.   On the iPad, the xsqz is also not read-only.   It's fully editable.

heffneil:
Is the problem adding a menu option on the File menu?  I mean we just fixed opening a xsqz with my other PR and you can't unzip and open another xsq because you won't have the right layout? Still not sure why this is so hotly contested.

heffneil:
Right click -> "Open With..." -> Select "All Applications" -> Select xLIghts. 

this is not a solution.  This is a much more arduous process than it should be to open a file in a program.  I tried it. It doesn't retain all applications so you literally have to do those steps every single time?

dkulp:
Closing this.... if using things properly, this doesn't provide any value and just clutters the menus/interface with things people shouldn't normally be using.

heffneil:
I just eliminated the audio menu. The tools menu has about 100 things on it so im not sure how this cluters one of the least busy menus.

PR #6332, opened 2026-05-11
Custom Model Wiring view: show Total Nodes + color-code string start/end

heffneil:
## Summary

Two small but related improvements to the Custom Model Wiring view to make it easier to verify wiring against a physical prop:

1. **Total Nodes count** in the header overlay
2. **Green start / blue end** color-coding on each string, matching the Layout tab's existing convention

## Why

When wiring up a physical prop, users use the Wiring view as a guide and need to verify two things:

1. **Does the diagram show the same total number of nodes I have on hand?** Today the only way to figure that out is to visually scan the diagram for the largest pixel number — easy to miss on dense models (a 400-node wreath has its "400" label buried in the middle of the spiral, not in a corner where you'd look).

2. **Where does each string start and where does it end?** Today the first node of each string is drawn in blue, and the last node has no distinct color at all — so users have to trace the wiring lines to find where each string terminates. The Layout tab already uses **green = start, blue = end** for two-point models (see `TwoPointScreenLocation.cpp:310-311`), and the Wiring view was inconsistent with that.

## Changes

**Header overlay (`RenderNodes` + `RenderMultiLight`):**
- New `Total Nodes: N` line directly below `Rotation:`. Count computed from the same `points` map the renderer iterates so it's guaranteed to match what's drawn.

**Node color-coding (`RenderNodes` + `RenderMultiLight`):**
- First node of each string → **green** (was blue in `RenderNodes`; was cycling palette in `RenderMultiLight`)
- Last node of each string → **blue** (was theme `nodeFill` in `RenderNodes`; was cycling palette in `RenderMultiLight`)
- Middle nodes unchanged — theme `nodeFill` in `RenderNodes`, palette-cycling colors in `RenderMultiLight`
- Single-node strings get green (start wins over end since "begin wiring here" is the more important indicator)

**Drive-by fix:** `RenderNodes`'s "Model:" line was using `_selectedTheme.labelOutline` while every other header line in both render paths uses `messageOutline`. Brought into line with the rest.

**Code-quality fixes per Copilot:**
- `totalNodes` accumulator is `size_t` not `int`; dropped the C-style cast on `.size()`.

## Cross-platform

All changes are pure wxWidgets drawing primitives (`wxBrush`, `wxPen`, `*wxGREEN`, `*wxBLUE`, `wxMemoryDC`, `RenderText` helper) — no platform-specific code. Behavior is identical on macOS, Windows, and Linux.

## Test plan

- [x] macOS Debug: Custom Model → Show Wiring → header shows correct `Total Nodes: N` matching the highest node label
- [x] macOS Debug: first node of each string is green, last is blue, middle nodes unchanged
- [x] macOS Debug: matches Layout's green/blue start/end handle convention
- [ ] Windows: same draw paths, no `#ifdef` gates — should behave identically
- [ ] Linux: same as Windows
- [ ] Print preview also picks up both changes automatically (uses the same render paths)

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,8 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -enh (Neil)                 Custom Model Wiring view: header overlay now includes a "Total Nodes: N" line below
+                                Rotation so you can see the model's pixel count at a glance without counting.
     -bug (scott)                Fix HTDemucs ONNX model download failing: the 12-second total curl timeout set for
                                 short API calls was also killing the large-file download before it could complete.
     -bug (dkulp)                Backup: switch the per-file copy from wxCopyFile to std::filesystem::copy_file so
--- a/src-ui-wx/model/WiringDialog.cpp
+++ b/src-ui-wx/model/WiringDialog.cpp
@@ -424,12 +424,12 @@ void WiringDialog::RenderNodes(wxBitmap& bitmap, std::map<int, std::map<int, std
         RenderText("CAUTION: Front view", dc, AdjustX(0, printer) + _start.x, 20 + _start.y, _selectedTheme.messageAltFill, _selectedTheme.messageOutline);
     }
 
-    RenderText("Model: " + _modelname, dc, AdjustX(0, printer) + _start.x, 20 + fontSize + 4 * printScale + _start.y, _selectedTheme.messageFill, _selectedTheme.labelOutline);
+    RenderText("Model: " + _modelname, dc, AdjustX(0, printer) + _start.x, 20 + fontSize + 4 * printScale + _start.y, _selectedTheme.messageFill, _selectedTheme.messageOutline);
     RenderText("Rotation: " + std::to_string(_rotation), dc, AdjustX(0, printer) + _start.x, 35 + fontSize + 4 * printScale + _start.y, _selectedTheme.messageFill, _selectedTheme.messageOutline);
 
-    int totalNodes = 0;
+    size_t totalNodes = 0;
     for (const auto& itp : points) {
-        totalNodes += (int)itp.second.size();
+        totalNodes += itp.second.size();
     }
     RenderText("Total Nodes: " + std::to_string(totalNodes), dc, AdjustX(0, printer) + _start.x, 50 + fontSize + 4 * printScale + _start.y, _selectedTheme.messageFill, _selectedTheme.messageOutline);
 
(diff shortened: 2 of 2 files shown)


heffneil:
Total Nodes:  now added under rotation

Screenshot 2026-05-11 at 10 19 54 AM

heffneil:
Addressed all three Copilot threads in 8df488911:

- **`int` totalNodes accumulator (lines 432, 572)** — switched to `size_t` and dropped the C-style `(int)` cast on `.size()`. Type now matches the source; no narrowing.
- **Outline color inconsistency (line 431)** — Copilot was right that the pre-existing "Model:" line in `RenderNodes` was using `labelOutline` while every other header line in both render paths uses `messageOutline`. Fixed it as a drive-by so all three header lines (Model / Rotation / Total Nodes) in `RenderNodes` now match `RenderMultiLight`'s already-consistent styling.

cybercop23:
Why not use the node count already in the model object? model->GetNodeCount()

cybercop23:
I think a better enhancement would be to make the start of each string move visible/identifiable, so you can see it for each string on a multi-string model, and you will also have it for the model itself.

heffneil:
I don't use node when wiring a prop.  No one does.  I don't want different colors.  people are colorblind.  Why is this line of text an issue? Serious question not being argumentative.

cybercop23:
IMHO.. too narrow.. solve only a you problem

heffneil:
I guess I am a person who uses Xlights? I mean I can post it in the facebook support group? As a matter of fact I would like to do that with the 11 other PR's that are not getting any notice in here.  Let the people decide not 4 expert users of xlights who know the program in and out.

heffneil:
Screenshot 2026-05-11 at 11 48 38 AM

The old colors were wrong. They don't follow the layout coloring. Blue was the start. Changed it to be green for start blue for the end. Combined suggestion by @cybercop23

computergeek1507:
I thought this idea was helpful.

deplanche:
How does this handle the case of missing nodes or duplicate nodes in the model?  Showing the total may not be that useful if it doesn't show one of those occurred.

heffneil:
Im unclear of these circumstances? What is a duplicate node? I am told we aren't supposed to have missing nodes so i'm not clear on that situation as well?

@computergeek1507 did you see the update with the color coding to match layout start and finish denotations?

deplanche:
When making a custom model, it is very easy for the person making it to add the same node number twice (say 2 pixels labelled 210 in the grid), or to completely miss adding a mode (no pixels labelled 210 in the grid).  Duplicate nodes are almost always an error, but they do happen when making the custom model.  Skipping nodes may be an error, but also has been used to add null pixels to the middle of a string.  While that isn't a current recommended way of going it, it does happen and users do this. This can happen on both user made models and vendor imported models.  Missing nodes also can occur when exporting native models to custom models where the pixel density is high (a dense mega tree custom model has this problem near the top of the tree).  

Showing the total nodes could give a false sense that the model is correct, if one of both of these conditions happen.  It would be more useful if that also included if there were any duplicate or missed nodes.  Then someone would know they need to correct something... and can look to check sequence for more information.  For example, it could list "Total Nodes: 350 (2 duplicates, 1 missing)" for the case above.  If no duplicates or missing, that would not need to be shown.  

Please don't take this as me not liking adding this, just trying to make sure it is useful for all cases, even if those cases are when there is an error or not recommended way of making a model.

heffneil:
Sure I understand and appreciate your response.  I think this is the wiring view - i.e. not an edit screen. This is the screen you use when you sit down and wire the prop. If you want to add more or less then that number should and will change (in the real world).  I believe in the model data section there is a way to denote the duplicates (not sure about missing). I think that is where custom model checking would and should occur?

deplanche:
But if the model data is wrong, it would come into the wiring view as wrong, so I think you would want to flag it here too.

Think about a vendor making a model and there is an error in it.  Would be great if that didn't happen, but we know it does.  A user that downloads that custom model has no need to go into the model data, they can go right into the wiring view, not knowing there is a problem with it.  Making it as clear as possible helps them from pushing pixels wrong.

heffneil:
I understand. I wonder - if I can have repeating numbers RED? I would think the developers (not me) and what I alluded to would be that originally this was just a wiring dialogue not so much a check validity.

PR #6339, opened 2026-05-12
Layout pixel size part deaux

AGFazio:
Should be better. Looks good on my screens.


derwin12:
Its early .. lets push this and test test test..

PR #6344, opened 2026-05-12
Revert #6326

derwin12:
reverting this .. it is forcing folks into 3d


AGFazio:
Just built this and it seems to work fine.

PR #6352, opened 2026-05-13
Add Audio toolbar: Volume slider + Speed -/+ controls

heffneil:
## Summary

Adds a new dockable **Audio Tool Bar** (alongside the existing Play / Edit / View / Windows toolbars) giving users always-visible playback **Volume** and **Speed** controls. The existing Audio menu's 5 volume presets and 8 speed presets are buried in a menu drill-down; this surfaces them on the toolbar.

## Why

Volume and speed during playback are common-enough adjustments that the menu round-trip is friction. After this PR, both are one-click on the toolbar, and the layout can be moved / floated / hidden like any other toolbar.

## Controls

| | Widget | Behavior |
|---|---|---|
| **Volume** | `wxSlider` 0-100 with "Vol:" label | Live update on drag |
| **Speed** | `-` / `+` `wxButton`s with a `wxStaticText` showing current preset (1/4x .. 4x) | Each click steps one preset |

Both bidirectionally synced with the existing Audio menu radio items:
- Slider lands on 100/66/33/10/0 → matching menu radio checked; lands elsewhere → all unchecked.
- Menu item picked → slider snaps.
- Toolbar `+`/`-` → walks the preset list, mirrors to menu.
- Menu speed picked → toolbar label updates.

## Why `-`/`+` buttons instead of a dropdown for speed

Tried both `wxChoice` and a button-that-pops-a-`wxMenu`. Both go through wx's modal `PopupMenu` on macOS, which **blocks the wx event loop while the popup is shown**. The playback timer pauses for the duration, and on dismissal the audio engine plays catch-up — perceptible as a "stop/restart" at speed changes. `-`/`+` buttons each fire one direct event with no popup, so the dispatch path matches the menu-bar's native NSMenu handling (which doesn't block timers). Confirmed audibly: toolbar speed change is now indistinguishable from menu-bar speed change.

## Implementation

- New `xlAuiToolBar* AudioToolBar` member; added as its own `wxAuiPaneInfo()` pane on `Row(1)` so it doesn't clip the existing top row.
- All UI changes are wxWidgets primitives (no `#ifdef`, no platform-specific code) — identical on Windows / macOS / Linux.
- `SetPlayVolumeTo(int)` is a new central helper that updates `playVolume`, `AudioManager::SetGlobalVolume`, the toolbar slider, and the matching Audio menu radio. The 5 existing menu volume handlers now just call this.
- Speed step handlers (`OnAuiToolBarSpeedDownClick` / `OnAuiToolBarSpeedUpClick`) walk a static `kPresetSpeeds[8]` array and call the existing `SetPlaySpeedTo(float)` — same path the Audio menu's `SetPlaySpeed(wxCommandEvent&)` ends up at.
- `TOOLBAR_SAVE_VERSION` bumped `"0003"` → `"0005"` so stale cached perspectives don't hide the new pane on existing installs.

## Test plan

- [x] macOS Debug: Audio toolbar appears on its own row at top of window, with Vol slider + `-` 1.0x `+` controls
- [x] Drag volume slider → audio level changes live; menu's Loud/Medium/Quiet/Very Quiet/Silent radio updates
- [x] Pick menu volume preset → slider snaps to matching value
- [x] Click `+` from 1.0x → walks to 1.5x, 2x, 3x, 4x (stops at 4x)
- [x] Click `-` from 1.0x → walks to 3/4x, 1/2x, 1/4x (stops at 1/4x)
- [x] Pick speed from Audio menu → toolbar label updates
- [x] Pane can be dragged, floated, re-docked via gripper
- [x] Volume persists across launches (existing behavior, unchanged)
- [ ] Windows: untouched code paths, pure wxWidgets — should be identical
- [ ] Linux: same

## Known pre-existing issue (out of scope)

Changing playback rate across the 1.0x boundary on macOS produces an audible stop/restart from the `AVAudioEngine`'s `AVAudioUnitTimePitch` node being inserted/removed by `ensureTimePitch()` in `macOS/src-apple-core/media/AVAudioEngineOutputBridge.mm`. This affects both the new toolbar `+`/`-` and the existing Audio menu identically (any 1.0 ↔ non-1.0 transition triggers an `[engine stop]` + `[engine startAndReturnError:]`). Going to be filed as a separate issue / PR against the macOS submodule.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,16 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -enh (Neil)                 Play toolbar: add an inline Volume slider (0-100) and a Speed dropdown (1/4x .. 4x)
+                                next to Play/Stop so users no longer have to drill into the Audio menu's discrete
+                                radio presets to change playback volume or speed during a session. Both controls
+                                are bidirectionally synced with the existing Audio menu items (slider snaps to /
+                                unchecks the Loud / Medium / Quiet / Very Quiet / Silent radio as appropriate;
+                                dropdown reflects the current playback rate). Volume still persists across launches
+                                via xLightsPlayVolume; speed still resets to 1.0x each launch. Pure wxWidgets
+                                (wxSlider + wxChoice), no platform-specific code — same UI on Windows / macOS /
+                                Linux. Bumped TOOLBAR_SAVE_VERSION so cached toolbar perspectives don't clip the
+                                new controls.
     -enh (Neil)                 Custom Model Wiring view: header overlay now includes a "Total Nodes: N" line below
                                 Rotation so you can see the model's pixel count at a glance without counting.
     -bug (scott)                Fix HTDemucs ONNX model download failing: the 12-second total curl timeout set for
--- a/README.txt
+++ b/README.txt
@@ -11,16 +11,17 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
-    -enh (Neil)                 Play toolbar: add an inline Volume slider (0-100) and a Speed dropdown (1/4x .. 4x)
-                                next to Play/Stop so users no longer have to drill into the Audio menu's discrete
-                                radio presets to change playback volume or speed during a session. Both controls
-                                are bidirectionally synced with the existing Audio menu items (slider snaps to /
-                                unchecks the Loud / Medium / Quiet / Very Quiet / Silent radio as appropriate;
-                                dropdown reflects the current playback rate). Volume still persists across launches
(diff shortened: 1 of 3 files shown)


heffneil:
Image showing the toolbar
Screenshot 2026-05-13 at 9 52 16 AM

heffneil:
Addressed all three Copilot threads in 0d9234b71:

- **`ID_AUITOOLBAR_SPEED_BUTTON` → `ID_AUITOOLBAR_SPEED_LABEL`** (xLightsMain.h:723, .cpp:193,818) — matches the actual widget (wxStaticText) and the member name `_playSpeedLabel`.
- **En-dash → ASCII hyphen** for the Slower button label (xLightsMain.cpp:813) — matches the existing `"-"` convention from `GenerateCustomModelDialog.cpp:2792`.
- **README rewrite** — release note now describes the actual shipped UI (dockable Audio toolbar with Volume slider + Speed `-`/`+` buttons), not the earlier iteration that mentioned a Play-toolbar Speed dropdown.

dkulp:
Also "OK" with this, but also needs ability to hide it (like the AC toolbar).   This is not something I expect many people to need/want so having it cluttering the interface by default is not desirable.

Issue #6353, opened 2026-05-13
macOS: AVAudioEngine stops/restarts audio whenever playback rate crosses 1.0x


heffneil:
## Summary

On macOS, changing the sequencer playback rate to or from **1.0x** produces an audible stop/restart of the audio (engine `stop` + `startAndReturnError:`). Walking between non-1.0x rates (e.g. 1.5x → 2x → 3x → 4x) is smooth; transitioning through 1.0x is not.

## Reproduce

1. Load a sequence with audio, press Play
2. Audio menu → Play 1.5x Speed → **audible stop/restart**
3. Audio menu → Play 2x Speed → smooth
4. Audio menu → Play 1.0x Speed → **audible stop/restart**
5. Audio menu → Play 3x Speed → **audible stop/restart** (because we just crossed 1.0 again)

## Root cause

`macOS/src-apple-core/media/AVAudioEngineOutputBridge.mm` — the `ensureTimePitch(bool need)` method (~line 113-154) inserts or removes the `AVAudioUnitTimePitch` node from the audio graph based on whether the new rate is exactly 1.0. The implementation:

```objc
void ensureTimePitch(bool need) {
    if (need == usingTimePitch) return;
    if (!engine) return;

    bool wasRunning = engine.isRunning;
    if (wasRunning) [engine stop];     // <-- audible stop

    // disconnect all player nodes
    // attach / detach timePitch
    // reconnect all player nodes

    if (wasRunning) {
        [engine startAndReturnError:&error];  // <-- audible restart
    }
}
```

`SetRate` calls `ensureTimePitch(rate != 1.0f)` on every rate change, so any transition through 1.0 forces an `[engine stop]` + `[engine startAndReturnError:]` cycle.

## Proposed fix

Keep the `AVAudioUnitTimePitch` node **permanently** in the audio graph (attach during `init()` regardless of starting rate), and use its inherited `AVAudioUnit.bypass` flag at rate==1.0 to avoid the time-stretching algorithm's latency:

```objc
// In init() — always insert timePitch:
[engine attachNode:timePitch];
[engine connect:timePitch to:mixer format:fmt];
usingTimePitch = true;

// In SetRate() — just update the rate and bypass flag:
void SetRate(OutputHandle* h, float rate) {
    h->playbackRate = rate;
    h->timePitch.rate = rate;
    h->timePitch.bypass = (rate == 1.0f);
}
```

[...]

PR #6355, opened 2026-05-13
House Preview: add Volume slider + Speed -/+ controls in the bottom bar

heffneil:
## Summary

Adds an always-visible row below the existing House Preview transport bar (Play/Pause/Stop/scrubber/time) with:

- **Volume slider** (0-100)
- **Speed `-` / `+` buttons** walking the 8 existing preset speeds (1/4x .. 4x)
- **Current speed label** between the buttons (e.g. `1.0x`, `2.0x`)

```
[Play] [Pause] [Stop] [<<] [<<10] [>>10]    [─────●──── scrubber ────]    00:00.000
Vol: [────●──────]   [-] 1.0x [+]
```

## Why

Quick volume / speed adjustment without leaving the House Preview pane or drilling into the Audio menu. The House Preview is where users watch playback during a sequence, so the controls sit right there at the bottom of the surface they're already looking at — same pattern as media-player transport rows.

## Alternative to PR #6352

There are two reasonable places to put these controls:

- **#6352**: a separate dockable **Audio Tool Bar** at the top of the window, alongside Play / Edit / View.
- **This PR**: an inline row in the **House Preview** bottom strip.

I built both and #6352 was already opened first. **Maintainers should pick whichever placement they prefer** — both share the same backend helpers and only differ in widget placement. The two would conflict on the helper declarations (`SetPlayVolumeTo` / `GetPlayVolume` on `xLightsFrame`); whichever lands second will drop the duplicate during rebase (annotated in the code with NOTE comments).

Trade-offs:

|  | #6352 toolbar | This PR (House Preview row) |
|---|---|---|
| Always visible regardless of pane state | ✅ (top of window) | ✅ (in the preview pane) |
| Visible in other previews (Layout, Sequencer Model Preview) | ❌ separate toolbar, lives at frame level | ❌ House-Preview-only |
| Real estate cost | Adds a toolbar row at top | Adds a strip at bottom of preview |
| Discoverability | Lives in the toolbar bar with other transport-like controls | Lives next to the thing it controls |
| Floating House Preview pane gets the controls | ❌ (toolbar stays in main frame) | ✅ (row floats with the pane) |

My personal lean: **this PR** for placement and **#6352**'s code path for the toolbar approach. They're complementary; you could ship both. But maintainers know best.

## Implementation

- `_mediaPanel` — new child `wxPanel` of `HousePreviewPanel`, sibling of `Panel1`. Added as a third row in `FlexGridSizer1` below `Panel1`. Leaves the wxsmith-managed `Panel1` layout entirely untouched.
- Speed buttons walk a static `kSpeeds[8]` array and call `xLightsFrame::SetPlaySpeedTo()` — the same code path the Audio menu's radio items use, so behavior matches the menu exactly (including the pre-existing 1.0x-boundary engine restart documented in **issue #6353** — separate problem, separate fix).
- Volume slider calls `xLightsFrame::SetPlayVolumeTo()` — same helper added by #6352.
- Pure wxWidgets primitives (`wxPanel` + `wxSlider` + `wxButton` + `wxStaticText`). No platform-specific code, no hover overlay (early prototype used `EVT_ENTER_WINDOW` on the canvas — unreliable when the canvas is grabbing 3D rotate/drag events; ditched in favor of this persistent row).

## Test plan

- [x] macOS Debug: Volume slider visible at bottom of House Preview in Sequencer tab
- [x] macOS Debug: dragging slider live-updates `AudioManager` global volume
- [x] macOS Debug: `+` walks 1.0x → 1.5x → 2x → 3x → 4x (stops at 4x); `-` walks the other direction
- [x] macOS Debug: Audio-menu speed pick reflects in the toolbar label
- [x] macOS Debug: visible and functional in 3D view (the original reason we abandoned the hover-overlay approach)
- [ ] Windows / Linux: pure wxWidgets, should be identical

## Known carry-over (not fixed by this PR)

Speed transitions through 1.0x produce an audible engine restart from `AVAudioEngineOutputBridge.mm::ensureTimePitch()`. Pre-existing in master, affects both this PR and #6352 identically. Tracked in **#6353**.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,13 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.09  May ??, 2026
+    -enh (Neil)                 House Preview: add an always-visible row below the existing transport bar with a
+                                volume slider (0-100) and speed -/+ buttons (walks the 8 preset speeds 1/4x .. 4x
+                                with a current-value label between them). Persistent, reliable in 2D and 3D views,
+                                no popup or hover-overlay needed. Speed buttons go through SetPlaySpeedTo() — the
+                                same path the Audio menu uses. Pure wxWidgets, no platform-specific code.
+                                Alternative direction to PR #6352 (which adds a dockable Audio toolbar at the top);
+                                maintainers can pick whichever placement they prefer.
     -bug (dkulp)                Fix Apple Intelligence crash inside FoundationModels.respond() on macOS 26.
                                 Reorder LD_RUNPATH_SEARCH_PATHS so /usr/lib/swift is searched before the .app's
                                 bundled libswift_Concurrency.dylib. With the old order, the back-deployed
--- a/src-ui-wx/layout/HousePreviewPanel.cpp
+++ b/src-ui-wx/layout/HousePreviewPanel.cpp
@@ -21,6 +21,7 @@
 #include "render/SequenceFile.h"
 #include "UtilFunctions.h"
 #include "sequencer/MainSequencer.h"
+#include "media/AudioManager.h"
 
 //(*IdInit(HousePreviewPanel)
 const long HousePreviewPanel::ID_BITMAPBUTTON1 = wxNewId();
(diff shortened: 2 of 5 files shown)


dkulp:
I'm "OK" with this approach, but the controls must hide/show with the current media playback controls on that panel.

heffneil:
I can add it to the menu.  I have another PR that has the menus over the house preview.

derwin12:
You got a screenshot of this?

heffneil:
Screenshot 2026-05-20 at 4 40 09 PM
Screenshot 2026-05-20 at 4 39 39 PM

heffneil:
Screenshot 2026-05-20 at 4 41 27 PM

dkulp:
As I said, since 90% of the people would have no use for this, this wastes valuable screen real estate.  It needs to be toggled with the play controls that are already on that panel (and likely could live to the right of those controls on the same line).

heffneil:
takes up more room on the top bar as buttons.  I think its better on the house preview.  We can remove playboack speed its dumb but volume should be on the house preview in my opinion.  my screenshots were in response to daryls request. I gave up on all my PR's

derwin12:
Neil, perhaps you are not aware. There is a preference that turns on or off the playback buttons. This needs to be paired with that.

heffneil:
https://github.com/xLightsSequencer/xLights/pull/6352

Issue #5843, opened 2026-02-21
Import multiple models


heffneil:
**Is your feature request related to a problem? Please describe.**
It would be GREAT if we could select multiple .xmodel files to import at one time.

**Describe the solution you'd like**
Select multiple files to import the models at one time rather than having to select each one file by file.

**Describe alternatives you've considered**
Now you have to go through the process of finding multiple files and importing them one by one.

cybercop23:
Please ensure they support xLights and follow the process to get their models listed in the catalog. With your multi-insert from the catalog, this would not be an issue

heffneil:
I agree - they all should.  For the smaller prop makers I don't think its affordable.  Others I think just aren't supportive.

dkulp:
This isn't something I would want to make easier... If vendors want their users to have an easier time, they should support xLights.   If vendors don't want to support xLights, xLights users should not support them.

heffneil:
fair enough - people do import models they worked on in other places but I respect that position.

PR #6365, opened 2026-05-14
Allow multi model export/import from the same xmodel file.

cybercop23:
Useful to have both a custom tree model and a native stat import together, or the WiredWatts octopus.
Or even multiple moving heads with the Pan/Tilt shadow models as a single import


heffneil:
https://github.com/xLightsSequencer/xLights/issues/5843

This project is a complete circus now

PR #6384, opened 2026-05-19
Keep individual start channel on model import

cybercop23:
So you can import multiple moving heads with their corresponding single-line for Pan, Tilt, Color, Dim etc.


AGFazio:
YES>  I Need this.. Push it please!!!!

derwin12:
What if it points to models that you dont have?

cybercop23:
> What if it points to models that you dont have?

Should be no harm, since they will just get the ! as they do now. Plus this is more for MH and the single lines so they should all be there.

Issue #6405, opened 2026-05-23
Moving Head model builder


AGFazio:
A tool for xlights that creates a multi model file for easy import and setup of moving heads using the Advanced Moving Head model.

What it does
Pick a moving-head fixture (or build a new one), set the number of heads and the DMX start channel, and the tool generates a ready-to-import xLights .xmodel file with:

One DmxMovingHeadAdv model per physical head
"String Group" sub-models you can drive together from the layout view (Dimmer, Shutter, Pan, Tilt, etc.)
A Model Group containing all heads (when you have more than one)
Properly mapped channel positions, pan/tilt motor parameters, color wheel slots, and head orientation
The output matches xLights' native .xmodel format — import it, place your heads in the layout, and you're done.

Also adding a new option in the tools menu that links to the builder.  https://agfazio.github.io/Moving-Heads/xmodel-builder.html

AGFazio:
PR 6416

PR #6416, opened 2026-05-25
#6405 Tools - Moving Head Builder

AGFazio:
Add a Moving Head Adv model builder to the Tool Menu.


computergeek1507:
I don't think xLights should have links to other unrelated websites, it will cause gatekeeping issues

AGFazio:
It can be hosted anywhere you want to. Put it on xlights Github. Google drive.

Issue #6279, opened 2026-05-01
Shift + Drag in 3D


jbs00210:
When I hold shift and drag in 3d, it is selecting models in opposite sides of my layout. 

MAC OS

Xlights Ver 2026.07.1

also replicated on my PC as well with version 2026.07

it appears that it only happens when you are zoomed in on the models

https://github.com/user-attachments/assets/c8d97c3b-bc81-4392-8534-d4dda41e0517

cybercop23:
PR: https://github.com/xLightsSequencer/xLights/pull/6354

cybercop23:
Next version.

PR #6420, opened 2026-05-25
Layout: fix 3D ring-drag group rotation (delta always zero)

heffneil:
## Summary

In the Layout tab, multi-selecting models (tree or preview Cmd-click) and dragging the **3D rotation ring** on the primary should rotate the whole selection as a rigid group around the primary's pivot. It was only rotating the primary; the rest stayed put. This PR fixes that.

`Bulk Edit Rotate X/Y/Z` (right-click menu) is unaffected — that path is intentionally per-model.

## Root cause

The per-frame delta is computed in `OnPreviewMouseMove3D` (`LayoutPanel.cpp:5547`):

```cpp
glm::vec3 new_worldrotate = sloc.GetRotationAngles();
glm::vec3 rotate_offset   = new_worldrotate - last_worldrotate;
```

`GetRotationAngles()` reads `ModelScreenLocation::angles`. That field is **initialized to `(0,0,0)` and never written anywhere** in the codebase (verified via grep). The legacy drag path used to write it; the new `BoxedRotateSession` writes to `rotate_quat` / `rotatex/y/z` via `Rotate()`, not to `angles`.

So `new_worldrotate - last_worldrotate = (0,0,0) - (0,0,0) = (0,0,0)` every frame. The primary still rotated visibly because `BoxedRotateSession::Update` applies its delta directly to the primary's quaternion. The follower path (in the multi-select branch) depends on reading the primary's frame-to-frame delta back out — which is what was broken.

## Diagnostic proof

Added temporary logging during testing — every drag tick produced:

```
RING-DRAG rotated: 'Star of Stars Wreath - Half'  around (161.27,-2.12,35.47)  by (-0.00,0.00,0.00)
RING-DRAG summary: multiSel-models-count=4  rotated=3  rotate_offset=(-0.00,0.00,0.00)
```

So:
- Multi-select branch fires ✓
- 3 follower models found ✓
- `RotateAboutPoint` is called on each ✓
- ...with a zero-degree delta ✗

## Fix

Replace `GetRotationAngles()` (broken — reads the never-written `angles`) with `GetRotation()` (which returns `glm::vec3(rotatex, rotatey, rotatez)` — the live values `BoxedRotateSession::Update` actually mutates via `Rotate()` calls).

Same replacement at all 4 sites — the 3 drag-start baselines (`ProcessLeftMouseClick3D`, 2× in `OnPreviewMotion3D`) and the per-frame read in the `AxisRing` branch — so both ends of the subtraction use the same source.

## Why other branches weren't affected

- **`AxisCube` (resize)** reads `GetScaleMatrix()` — independent of the broken `angles` accessor. Group resize worked.
- **`AxisArrow` (translate)** reads `GetCenterPosition()` — also independent. Group drag worked.
- Only the rotate branch went through `GetRotationAngles()`.

This is why the user could group-resize and group-translate but couldn't group-rotate.

## Regression history

Silently regressed during the legacy-to-new-drag-session refactor — likely one of:
- `92ae2d594` Port the SpaceMouse stuff over to new drag session API's, allows removing the remaining legacy stuff
- `e49e5a8f9` Start cleaning up the handle hit test/drag stuff to be able to use it cleanly with iPad and finger instead of mouse
- `dc05d1832` Fix move/resize/rotate handles on 3d Objects not working
- `e1c0104a3` Fix (most) 3D lasso multiselect #6279 (#6354)
- `164cf69ac` Some leftover cleanup of the legacy drag/mouse apis

None of those touched `GetRotationAngles()` directly — but cumulatively they replaced the legacy code that *was* writing `angles` with the new session that doesn't.

## Test plan

- [x] macOS Debug 3D: select multiple models (tree-multi-select + preview Cmd-click both tested), drag rotation ring on primary → whole group rotates around primary's pivot as a rigid unit
- [x] Single-model rotation unchanged
- [x] Group resize still works (regression check)
- [x] Group translate still works (regression check)
- [x] Bulk Edit Rotate X/Y/Z (right-click menu) unchanged
- [ ] Windows / Linux: pure C++ change, no platform-specific code
- [ ] `OnPreviewMotion3D` SpaceMouse path: also uses the same drag-start helper that I fixed, so should now work correctly with multi-select too — wasn't tested directly

## Files touched

| File | Change |
|---|---|
| `src-ui-wx/layout/LayoutPanel.cpp` | 4 sites: `GetRotationAngles()` → `GetRotation()` |
| `README.txt` | release note |

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -14,12 +14,13 @@ XLIGHTS/NUTCRACKER RELEASE NOTES
     -bug (heffneil)              Layout 3D rotation ring with multi-selection now actually rotates the whole
                                  selection as a rigid group around the primary-selected model again. Regressed
                                  silently during the legacy-to-new-drag-session refactor: the per-frame delta
-                                 was being computed from GetRotationAngles(), but the `angles` field that
-                                 backs it is never written so it was always (0,0,0). Switch to GetRotation()
-                                 which reads the live rotatex/y/z that BoxedRotateSession actually updates
-                                 during the drag. Same fix at the 3 drag-start sites so the baseline matches.
-                                 Bulk Edit Rotate X/Y/Z (right-click menu) is unaffected — that path is
-                                 intentionally per-model.
+                                 was being computed from ModelScreenLocation::GetRotationAngles(), but the
+                                 `angles` field that backed it was never written by anyone — always returned
+                                 (0,0,0). Fixed at the accessor level so the trap is gone for all callers
+                                 (two iPad twist-rotate sites had the same latent bug). The `angles` field is
+                                 deleted; GetRotationAngles() now reads the live rotatex/y/z that
+                                 BoxedRotateSession actually updates during the drag. Bulk Edit Rotate X/Y/Z
+                                 (right-click menu) is unaffected — that path is intentionally per-model.
     -enh (derwin12)              Add "Per Model Default" layout option to model groups so effects using the
                                  Default render style automatically render per-model on that group (#4125).
     -enh (heffneil)              Add Bulk Edit Rotate X / Y / Z to the Layout tab right-click menu.
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,15 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.10  May ??, 2026
+    -bug (heffneil)              Layout 3D rotation ring with multi-selection now actually rotates the whole
+                                 selection as a rigid group around the primary-selected model again. Regressed
+                                 silently during the legacy-to-new-drag-session refactor: the per-frame delta
+                                 was being computed from GetRotationAngles(), but the `angles` field that
+                                 backs it is never written so it was always (0,0,0). Switch to GetRotation()
+                                 which reads the live rotatex/y/z that BoxedRotateSession actually updates
+                                 during the drag. Same fix at the 3 drag-start sites so the baseline matches.
+                                 Bulk Edit Rotate X/Y/Z (right-click menu) is unaffected — that path is
+                                 intentionally per-model.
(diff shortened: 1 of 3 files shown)


heffneil:
Addressed the Copilot thread in 9e087d851 — the better fix per the review.

- `ModelScreenLocation::GetRotationAngles()` now reads the live `rotatex/y/z` (same as `GetRotation()`) instead of the never-written `angles` field. Trap is gone for every caller, not just the four I patched in LayoutPanel.cpp.
- Deleted the unused `angles` member (only the broken accessor referenced it).
- The two iPad twist-rotate call sites (`src-iPad/Metal/XLMetalBridge.mm:2711` and `:2839`, both saving the rotation before a twist gesture) had the same latent bug and are now fixed implicitly.
- Comment block at the `AxisRing` call site removed since the trap is gone.

heffneil:
Closing this in favor of @derwin12's #6421, which is a better fix.

Same root cause (the never-written `ModelScreenLocation::angles` field that `GetRotationAngles()` was returning), but his PR fixes it more thoroughly:

- **Adds a `DragSession::GetRotationInfo()` virtual** that exposes the drag session's own raw accumulated degrees + actual rotation axis — bypassing the lossy quat→Euler round-trip that my fix was still relying on (which can produce wrong-sign deltas across certain gimbal positions).
- **Covers the 2D rotate path** in `OnPreviewMouseMove` and `Boxed2DRotateSession` — my patch only touched the 3D code path.
- **Covers `TwoPointRotateSession`** (single-line/string models) — my patch didn't address those.
- Adds the same source-level `GetRotationAngles()` accessor fix I made (identical line).

For any reviewer who wants the diagnostic proof of the root cause, the `RING-DRAG` spdlog lines in this PR's description confirm `rotate_offset = (0,0,0)` every frame because `angles` is the never-written field — that holds for derwin12's fix too.

Going with the more thorough one.

derwin12:
Careful .. dont include macos sudmodule updates in your PRs.

heffneil:
noted in my md file. Thanks for the heads up!

Issue #6418, opened 2026-05-25
Pivot point for rotating multiple objects is broken .09


derwin12:
When selecting multiple objects, and rotating on the 1st object, rotate only rotates the one and not the entire selection.

derwin12:
In the next update

PR #6421, opened 2026-05-26
Fix Pivot point for rotating multiple objects  #6418

derwin12:
image

### DIFF
--- a/src-core/models/BoxedScreenLocation.cpp
+++ b/src-core/models/BoxedScreenLocation.cpp
@@ -997,6 +997,12 @@ public:
 
     handles::Id GetHandleId() const override { return _handleId; }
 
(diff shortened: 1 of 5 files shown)


heffneil:
FYI — I had also tracked this down to the same root cause yesterday in #6420 (now closed in favor of this PR). Your fix is the better one: exposing the drag session's raw accumulated angle via `GetRotationInfo()` is architecturally cleaner than the accessor patch I made, and you covered 2D + TwoPoint sessions which mine missed.

The closed PR has the `RING-DRAG` diagnostic spdlog lines I captured during local repro — `rotate_offset = (0,0,0)` every frame because `ModelScreenLocation::angles` is never written. Leaving it here in case any reviewer wants the bisection trail.

PR #6447, opened 2026-05-31
Preserve relative Start Channel even if Individual Start Channels is not checked

cybercop23:
As example, the 10 MH YPS Spider model, the MHs are relative to a 216-node string.


dkulp:
@copilot resolve the merge conflicts in this pull request

Issue #5506, opened 2025-08-27
Enhancement To Custom Model Input Grid


dragonzpneuma:
When creating a custom model and adding a background image to the grid, it would be beneficial to be able to move the image up/down/left/right to align the image better with the cells.  For example, as you can see from the attached pic, it would be beneficial to nudge the background image to align the dots better with the cells.
Thank You!!!!

<img width="1717" height="1320" alt="Image" src="https://github.com/user-attachments/assets/bb104bed-a1b6-4e6f-a22b-50be928771f9" />

derwin12:
In the mean time, add or subtract a couple from width and or height.

cybercop23:
#6451

derwin12:
In the next update

PR #6451, opened 2026-06-01
Add the ability to move image in Model Data (#5506)

cybercop23:
Once a background image is loaded, you can move it by 0.5 cells using the arrows:
image


cybercop23:
Added reset button per CoPilot's suggestion - nice

derwin12:
It doesnt save the settings for next time? The offset should be used for the  image if used to display on the layout.

cybercop23:
That was a purpose not to save the offset, since it was just to make clicking nodes easier to line up. Using it as a background should still be relatively okay, unless you shift it way off. I can look at adding that in, but I didn't think it was necessary. Lets discuss on zoom.

PR #6468, opened 2026-06-02
Layout tree: show padlock indicator next to locked models

heffneil:
## Summary

- Locked models in the Layout model tree on the left now have a small padlock glyph appended to their name, so you can see at a glance which models are protected from drags/resizes/property edits without having to click each one and check the Locked property.
- Skips ModelGroup and SubModel rows since they don't carry a meaningful per-instance lock on the layout.
- Toggling **Locked** in the property grid now adds `WORK_RELOAD_MODELLIST` to the ASAP work mask (for Boxed/TwoPoint/Poly screen locations) so the tree picks up the new state immediately instead of waiting for an unrelated rebuild.

## Test plan

- [x] Build (Release, macOS) — clean.
- [ ] Lock a model from the property grid → padlock appears next to its name in the tree the moment the property change settles.
- [ ] Unlock → padlock disappears.
- [ ] Lock a model inside a group → padlock shows on both the top-level row and the group-child row.
- [ ] Verify groups and submodels never show a padlock.


heffneil:
Screenshot 2026-06-02 at 5 46 07 PM
Screenshot 2026-06-02 at 5 46 18 PM

heffneil:
Thanks @copilot — going through these:

**Fixed**

- **Raw UTF-8 → wxString on Windows** (3 comments on this): Real bug. The implicit `std::string` → `wxString` conversion uses `wxConvLocal` (CP1252 on Windows) which would mojibake the padlock. Wrapped the `AppendItem` label with `wxString::FromUTF8(...)` so the bytes are always interpreted as UTF-8. Pushed in efea86ef3.

**Declined (with reasoning)**

- **Use `u8"\U0001F512"` instead of `\x` escapes**: in C++20 `u8"..."` is `const char8_t*` and doesn't implicitly convert to `const char*` — would require a cast at every use. The `\x` byte escapes are 100% portable at the source level; the only portability risk was the destination encoding, which the `FromUTF8` change fixes.
- **Add `[Locked]` text fallback for accessibility**: the indicator is supplementary — locked models already get a red bounding box in the 3D preview and the property grid shows a `Locked` checkbox. Doubling up the text label would clutter the tree for every locked row. Open to revisiting if a user actually reports a screen-reader gap.
- **`WORK_RELOAD_MODELLIST` / `ReloadModelList()` is expensive**: matches the pattern already used throughout `LayoutPanel` (e.g. `BulkEditActive`, `BulkEditDimmingCurves`, rename, replace, etc. all do the same full reload). Targeted single-row refresh would be a worthwhile project on its own but is out of scope for a 3-line indicator fix.
- **Factor the duplicated `Locked` work-mask**: agreed it's duplicated across Boxed/TwoPoint/Poly handlers, but that's pre-existing code shape — every property in `ScreenLocationPropertyHelper.cpp` is repeated three times the same way. A refactor would touch every handler, not just `Locked`. Better as a separate cleanup PR.

cybercop23:
I would want this behind an option since all my models are locked, and won't want to see locks for every model in the list.

heffneil:
Noted. I’m not changing it but noted.

focusedonsound:
This is a great addition.  

Would be great to have the option at the top to lock / unlock all,  while also being able to individually lock / unlock.  Almost like at the top of a column.

I would also note that putting this behind an option in the preferences menu defeats the purpose.  Many users who come into the zoom room often overlook this due to the fact that outside of the red box there is no other indicator to understand that the model is locked in place.

cybercop23:
You can already do that. Select All - and right click lock or better yet set keybindings - a (select all) - l/U (lock/unlock).
As we discussed, this could have its own column so the user can choose where to put it/show it.

heffneil:
it can not have its own column based on claude:
The tree-control's first column is special (expand arrow + icon + text — you can't insert before it). The cleanest way to get a "consistent" lock position is to move the 🔒 from a suffix to a prefix on the name, so it always sits immediately before the model name text instead of trailing it.

Alex I appreciate your perspective however, I didn't want to add even more code it possible to hide the lock and it didn't seem necessary.  I appreciate you might lock most of your models but I suspect you are outside the norm.  This lock hardly takes up any space but without direction - are you the only voice we need to develop around?

cybercop23:
Neil, you can add as many columns as you need, don't believe everything Claude says. It needs to make sense for the majority of people. If you take your inspiration from other tools, do they have the lock at the beginning of the column or the end?
How do you know how many people lock their layout and how many lock some, many or all?

dkulp:
I should point out that the iPad app has some icons there so this would bring it in closer to the iPad app....  Right now the iPad app puts a link icon for anything linked to the base folder.   

Screenshot 2026-06-03 at 1 25 16 PM

heffneil:
I pay attention to the people who come in the zoom room.  I see their screens.  Most don't lock anything, many don't even know about it, I have yet to see any other person who has locked a majority of their models.  Its not a guess its observation.

if you don't like the feature alex flush it I don't care. Your point is clear and this will be my last contribution. There are 5 guys who are in the club and you guys decide what goes in and out.  Since it doesn't wet your whistle it doesn't make it in.  I provide new perspective and certainly less expertise in the product.  You could almost say I am more like the average user than the aforementioned 5.  But you guys keep doing you.

cybercop23:
I didn't say I didn't like it, just its position. Just as Dan indicated the link to base in iPad, that's at the end, not in front. I'm sure you can get your intern to add another column and place it in that column. There's already functionality that allows the user to sort the columns and place them where they want. Make use of what's already there and add to it.

derwin12:
Is it just a unicode character? can it be suffixed to the name?

deplanche:
Instead of a lock icon, could you just change the color of the font or the background for locked models?  Similar to what is done for the actual model, when it changes to red when locked.

heffneil:
Closing this in favor of @cybercop23's approach: rather than appending a padlock glyph to the model name in the tree's first column, the lock state should be its own sortable column in the model tree, with the indicator at the end — consistent with how the iPad app places the base-folder link icon (per @dkulp's note). That lets users show/hide/position it via the existing column functionality instead of it always trailing the name.

Thanks @cybercop23, @dkulp, @focusedonsound, @derwin12, @deplanche for the feedback. Closing the suffix-glyph implementation; the column-based version can be picked up separately.

Issue #6478, opened 2026-06-04
Add Drag/Drop Effects in sequencer as the iPad has. 


AGFazio:
Add drag and drop to the sequencer like in the iPad version. 

● The branch adds ghost drag-to-move for effects in the sequencer grid — instead of the old behavior where clicking an effect body and dragging would move it in real time (mutating position on every
  mouse tick), the new behavior is non-destructive:

  - Drag: the original effect dims in place, a ghost rectangle shows the target position. The ghost turns red and the cursor switches to a no-entry symbol when the target slot is occupied.
  - Drop: on release, the move is applied atomically — same-row drops shift the times, cross-row drops copy the effect to the new layer and delete the original.
  - Cancel: Escape during drag discards the operation with no undo entry.
  - Multi-select: clicking an already-selected effect in a group preserves the whole selection; all selected effects drag together. Releasing without moving collapses back to single selection
  (standard DAW behavior).
  - Snap: Ctrl inverts the global snap-to-timing-marks setting, matching existing edge-resize behavior.
  - Full undo support throughout.

  Edge resizing (left/right edge drag) is completely unchanged.

derwin12:
Need more details :-)

AGFazio:
Adding ability to let you grab and drop effects across timeline and models not just up and down with arrow keys. Also allows dragging of multiple effects once selected or lassoed.

AGFazio:
> Need more details :-)

In first post.

AGFazio:
https://github.com/user-attachments/assets/4503ae19-eff2-4319-983f-55fc918e5fe4

derwin12:
In the next update..

PR #6479, opened 2026-06-04
#6478 - Add ghost drag-to-move for sequencer effects

AGFazio:
## Summary

Replaces the old real-time effect body drag with a non-destructive ghost drag-to-move:

- **Originals dim in place** during drag; a **ghost rectangle** shows the target position
- **Collision detection**: ghost turns red and cursor switches to no-entry when the target slot is occupied — drop is blocked
- **Same-row drop**: shifts the effect in time atomically on release
- **Cross-row drop**: copies the effect to the new model/layer row and deletes the original
- **Multi-select group drag**: clicking a selected effect in a multi-selection preserves the group; all effects drag together with relative row/time offsets preserved; releasing without dragging collapses to single selection (standard DAW behaviour)
- **Snap-to-timing**: Ctrl inverts the global snap-to-timing-marks setting during drag, matching existing edge-resize behaviour
- **Escape cancels** the drag with no undo entry
- Full undo support throughout
- Edge resize (left/right handles) is completely unchanged

## Test plan

- [ ] Drag a single effect left/right within the same row — confirm it lands at the ghost position
- [ ] Drag an effect to a different model row — confirm it moves (not duplicated, not deleted)
- [ ] Drag into an occupied slot — confirm red ghost + no-entry cursor; effect stays on release
- [ ] Select multiple effects, drag as a group — confirm all move with relative offsets preserved
- [ ] Press Escape mid-drag — confirm effect stays at original position and undo stack is clean
- [ ] Ctrl+drag with a timing row selected — confirm snap-to-timing-marks toggles
- [ ] Undo after a move — confirm effect returns to original position and row
- [ ] Edge resize (drag left/right handle) still works as before

### DIFF
--- a/src-ui-wx/sequencer/EffectsGrid.cpp
+++ b/src-ui-wx/sequencer/EffectsGrid.cpp
@@ -560,6 +560,12 @@ void EffectsGrid::rightClick(wxMouseEvent& event) {
 void EffectsGrid::keyReleased(wxKeyEvent& event) {
 }
 void EffectsGrid::keyPressed(wxKeyEvent& event) {
+    if (event.GetKeyCode() == WXK_ESCAPE && mEffectMoveDragging) {
+        CancelMouseOperations();
+        mSequenceElements->get_undo_mgr().SetCaptureUndo(false);
+        mSequenceElements->get_undo_mgr().RemoveUnusedMarkers();
+        Draw();
+    }
 }
 
 void EffectsGrid::sendRenderDirtyEvent() {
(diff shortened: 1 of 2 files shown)


derwin12:
Has this been tested on 2d and 3d?

derwin12:
Got a video?

AGFazio:
Not yet. Will do more when I get home. So far seems ok, I am having the intern clean it up. I'll let you know.

AGFazio:
https://github.com/user-attachments/assets/4909e4cb-fd91-4663-8872-dc033675dab5

PR #6485, opened 2026-06-05
Crashes from the logs

derwin12:
Fix crash on hardware with OpenGL 1.1 (GDI Generic) when shaders fail to initialise 
Fix crash in PolyLine model when two adjacent control points are coincident (zero-length segment) 
Fix crash in Waveform render when media is closed during a pending paint


derwin12:
Closed .. not confident in any of these...

PR #6486, opened 2026-06-05
Add Effect Wheel for quick access to keybingings - sequencer - effects

cybercop23:
Quick way to add effects in an empty spot by double-clicking

https://github.com/user-attachments/assets/eedccf36-960f-4361-adea-7398699d7049


derwin12:
I think this might be better filled with presets rather than keybindings... why not just do the O key for the on for example.

cybercop23:
Would it be too much to have an option for both? I thought presets too, but IDK how the image would look. With this as keybindings, I can recoup my keys and make the Effect keys more "complex" (shift/crtl) and get back my single keys.

Issue #4688, opened 2024-07-31
Adjustment Effect no longer working since 2024.11


derwin12:
It appears the canvas change may have impacted this .. 

Prior
![image](https://github.com/user-attachments/assets/f5a7dcb5-8a23-421e-873d-421a10a7ac44)

After .11
![image](https://github.com/user-attachments/assets/6eeb2478-a33c-4e20-ae02-77338f18b909)

Just selecting some values to make it apparent..
![image](https://github.com/user-attachments/assets/9c421a29-99db-4e2d-a690-8a6a905dac57)
![image](https://github.com/user-attachments/assets/52e734b1-ad75-4a02-9e82-913ac51c5f38)

derwin12:
![image](https://github.com/user-attachments/assets/0c02ef4f-2cd0-46e7-a485-ae1108c0da60)

computergeek1507:
Probably the canvas mode changes

derwin12:
Dan committed the fix..

PR #6489, opened 2026-06-05
Add cycles to custom value curves (#4688)

cybercop23:
(#4688) add an easy way to "replicate" a custom curve.


derwin12:
Intern needs to review to copilot stuff -- lots of good things in there..
I think this also needs an adjust effect coded to handle older sequences and set the cycle to 1.

derwin12:
Cant push -- its got an inadvertent macos update.

cybercop23:
My guy took care of it. Should be okay now.

PR #6494, opened 2026-06-05
Add FadeIn & FadeOut handles

cybercop23:
ala iPad. 

https://github.com/user-attachments/assets/51d3e864-6da6-4cf4-a3ce-2afbe7df12a3


o0charlie0o:
I have a similar implementation to this locally which I never put a PR in for but basically it implements a "smart tool" behind a modifier key that lets you do what you have in that video as well as sparkles and brightness. Maybe we could combine the approaches?

cybercop23:
> I have a similar implementation to this locally which I never put a PR in for but basically it implements a "smart tool" behind a modifier key that lets you do what you have in that video as well as sparkles and brightness. Maybe we could combine the approaches?

Hey Charlie. Sure, anything to make the UI nicer and easier to use. I'm on zoom most of the time/days. Hop in and we can review, or draft a PR or open a discussion, and we can colab.

Issue #6488, opened 2026-06-05
Add cycles to custom value curves


cybercop23:
Add the ability to "replicate" the curve shape x cycles

cybercop23:
<img width="792" height="295" alt="Image" src="https://github.com/user-attachments/assets/191e6203-950a-4af2-b282-3027f719cc0d" />

cybercop23:
In the next version.

derwin12:
In the next update.

PR #6495, opened 2026-06-05
Add cycles to custom value curves (#6488)

cybercop23:
Easy way to replicate across multiple timing marks (#6488)
image


cybercop23:
HELP!!! Sorry. I had both my interns review this before the new resubmit and assured me it would be fine. Clearly, it is not. Compiles and runs fine on Mac.

Issue #5584, opened 2025-10-12
xlights pushes large channel numbers to Falcon V3 controllers that it can't support


darylc:
**Describe the bug**
xlights pushes large channel numbers to Falcon V3 controllers that it can't support when in DDP mode.

We have seen this several times in the zoom room this year.

I suggest that falcon.xcontroller is modified to add:
< DDPStartsAtOne/ >
to:
< AbstractVariant Name="BaseFalconV3Settings" Base="Falcon:BaseFalconSettings" > 

so that xlights uploads valid channel numbers.

There is a related bug #5581 to fix One Based DDP on Falcon v3 controllers for Serial ports.

derwin12:
I dug out my v3 controller and I tried to duplicate this and couldnt - perhaps the OP didnt have the latest firmware. There is a trigger condition if the status page is still at 1 and the outputs page is some 40K number. Ie the spread cant be that big. A normal push should have these set correctly. (outputs page also has a serial device so it doesnt start at the start channel here in the screengrab)

I know we have seen this in the past .. but perhaps  it just a config thing ..

<img width="1047" height="106" alt="Image" src="https://github.com/user-attachments/assets/6a3da301-4620-407c-a52d-f4cef5a4dfe3" />

<img width="1034" height="277" alt="Image" src="https://github.com/user-attachments/assets/0a995444-c7bb-446b-9b7c-e650b5bac0a0" />

darylc:
@derwin12 This is still an issue, I was running the latest firmware.  Try channel numbers in the 100's of thousands, 40k is an example of a working channel number.

This is a genuine problem and seems like a simple fix.  Happy to pull out a controller and demonstrate to you in zoom sometime if you like?

derwin12:
I got mine online and here to test.. I just pasted 672,000 as a channel number. So I guess i need to know where it fails.

derwin12:
The key is to make sure the start channel on the first page is the start of the channel range - which should be the case if pushed from xlights.

darylc:
@derwin12 Thanks for looking at this, I really appreciate it.  I pulled out a f16v3 firmware 2.59, factory reset it.  Flipped one of my problem controllers from E1.31 to DDP in xlights, hit upload.

Pixel ports look OK, but DMX ports look wrong.

<img width="1696" height="668" alt="Image" src="https://github.com/user-attachments/assets/d1046462-3c2a-436c-8a8b-eb85ae903b34" />

[...]

PR #6501, opened 2026-06-06
Fix Falcon V3 controllers receiving DDP channel numbers beyond hardware maximum (#5584)

derwin12:
The serial ports page in Falcon V3 is limited in how high a value you can enter. So warn and block if the channel is too high and you have a serial port configured.


computergeek1507:
Why not require keep start channels to be disabled

derwin12:
> Why not require keep start channels to be disabled

I thought that seemed a bit heavy handed .. with everyone currently using the sw as is, why make them have to change their configs now? I also was not 100% certain that is the correct fix - as there was some talk about keep channels only being for fpp/wled devices.

Issue #6509, opened 2026-06-07
Sort Models/Groups in Sequencer view


AGFazio:
Sort models and groups in the sequencer view. Same sort as in Edit Display Elements.

AGFazio:
PR 6510

PR #6510, opened 2026-06-07
#6509-Sort-Sequencer-Display-Elements

AGFazio:
Sort capability for the sequencer view without having to go to edit display elements. Same sort functions as in Edit Display Elements.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,8 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.11  June ??, 2026
+    -enh (AGFazio)               Add "Sort Display Elements" submenu to sequencer row heading right-click menu,
+                                 with the same sort options as Edit Display Elements.
     -enh (cybercop23)            Import Effects: stacked mappings — dropping/double-clicking an available model
                                  onto an already-mapped destination shows Replace / Add Additional prompt;
                                  Add the ability to load multiple xmap files with a prompt of what to do if maps exist.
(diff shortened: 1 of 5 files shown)


computergeek1507:
Not a huge fan of this idea, their is already a ton of right click menu options in the row heading and this functionality already exists in the Edit Display Elements

derwin12:
> Not a huge fan of this idea, their is already a ton of right click menu options in the row heading and this functionality already exists in the Edit Display Elements

I guess this is valid .. folks dont need to sort all that often and should just as easily do it in the display elements section.

cybercop23:
Isn't this the same as the new reorder by dragging, just another way of doing the same thing? The flow is you do the sort and then drag around if you still need slight changes.

derwin12:
I think the case here is that it doesn't provide anything new, simply a duplication of what is available elsewhere. The drag n drop provides a quick way to tweak a current sequence.  Sorts are done once when you create the view and unlikely to need to do that while sequencing. Probably should be an ongoing guide, dont just add duplication of existing functions?

deplanche:
Another thing to consider is if this is able to sort the Master View, it changes the way a sequence is rendered.  Right now, you have to open a different window to be able to change that.  If it's a drag and drop or a right click and sort, people might not realize the implications of that when taking those actions.  I can already hear people coming in for help saying they aren't sure why things look different on their sequence - they didn't do anything, except change the order of the list to be alphabetical so they could find a model hiding in the list.  Yes, they can do the same in the Edit Display Elements, but that action makes it seem like something different that has more consequences.

computergeek1507:
I do think the "concept" of how we edit view should be reevaluated, it is a frequent pain point with people trying to add new models to old sequences. But I agree with Jon, it needs to be "intentional" like a edit mode enabled with a tool bar or something else.

PR #6521, opened 2026-06-09
Bulk Edit Rotate: skip models whose type doesn't support that axis

heffneil:
## Summary

Follow-up to #6184. Bulk Edit → Rotate X / Y / Z applied the rotation unconditionally to every unlocked selected model, but several model types don't actually honor every axis:

- **BoxedScreenLocation** (Matrix, Custom, Image, Cube, Sphere, Circle, Star, ModelGroup, etc.) — supports X, Y, Z.
- **ThreePointScreenLocation** (Arches, CandyCane, Icicles) — only X has a visible effect.
- **TwoPoint / PolyPoint / MultiPoint** (SingleLine, PolyLine, ChannelBlock, etc.) — orientation comes from endpoints; calling `SetRotateX/Y/Z` stores a value with no visible effect.

So a user could bulk-rotate a mixed selection and watch some models silently ignore the action.

## Fix

`BulkEditRotateAxis` now classifies each selected (unlocked) model into "supported on this axis" or "unsupported", and:

- If **no** selected model supports the axis, the value-entry dialog is skipped and an info box lists the model names with the reason.
- If **some** models support and some don't, the supported ones are rotated normally and a follow-up info box lists the skipped names, one per line, capped at 20 with "… and N more" for huge selections.
- If **all** selected models support the axis, behavior is unchanged (no extra dialog).

Includes are already pulled in transitively (the same `dynamic_cast<ModelWithScreenLocation<BoxedScreenLocation>*>` pattern is already used at `LayoutPanel.cpp:1384`). wx-platform-neutral — works on macOS, Windows, Linux.

## Test plan

- [x] macOS Release build (Xcode 26.5)
- [ ] Select a mix (e.g. Matrix + Arch + SingleLine), Bulk Edit → Rotate Y — Matrix rotates; follow-up dialog lists Arch and SingleLine as skipped
- [ ] Select only Arches + Lines, Bulk Edit → Rotate Y — no value dialog appears; info box lists all of them as skipped
- [ ] Select Matrix + Arch, Bulk Edit → Rotate X — both rotate; no extra dialog (Arch supports X)
- [ ] Select only Matrix models, Bulk Edit → Rotate Z — unchanged behavior, no extra dialog
- [ ] Lock one model in a mixed selection — locked model is silently skipped as before, no double-count in the "of N" total
- [ ] Select 25 unsupported models — skipped list shows 20 names plus "… and 5 more"

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -2406,19 +2406,68 @@ bool ModelSetHasLockedMember(ModelManager& mgr, ModelSet* s) {
     }
     return false;
 }
+// Which axes a model's screen-location class meaningfully rotates around.
+// BoxedScreenLocation supports all three. ThreePointScreenLocation (Arches,
+// CandyCane, Icicles) only exposes X. Two/Poly/Multi/Terrain point screen
+// locations derive their orientation from their anchor points - calling
+// SetRotateX/Y/Z stores a value but has no visible effect.
+bool ModelSupportsRotationAxis(Model* model, char axis) {
+    if (model == nullptr) {
+        return false;
+    }
+    if (dynamic_cast<ModelWithScreenLocation<BoxedScreenLocation>*>(model) != nullptr) {
+        return true;
+    }
+    if (dynamic_cast<ModelWithScreenLocation<ThreePointScreenLocation>*>(model) != nullptr) {
+        return axis == 'X';
+    }
+    return false;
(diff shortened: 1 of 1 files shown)


heffneil:
Some screenshots of the process and alerts now.

Screenshot 2026-06-09 at 3 51 09 PM
Screenshot 2026-06-09 at 3 51 51 PM
Screenshot 2026-06-09 at 3 51 55 PM

PR #6537, opened 2026-06-11
Adds auto-scroll to dragging of effects and/or elements list.

AGFazio:
Sequencer: auto-scroll during drag operations

Row reorder (RowHeading): When dragging a model/group row to reorder, the list now auto-scrolls up or down when the cursor approaches the top or bottom edge of the visible area. Previously it was impossible to drag a row past the visible boundary.

Effect drag-to-move (EffectsGrid): Auto-scroll in all four directions (up, down, left, right) when dragging an effect and the cursor nears any edge of the grid. Vertical mouse drag moves the model rows; horizontal mouse drag moves the timeline position.

### DIFF
--- a/src-ui-wx/sequencer/EffectsGrid.cpp
+++ b/src-ui-wx/sequencer/EffectsGrid.cpp
@@ -8044,7 +8044,6 @@ void EffectsGrid::ResetEffectMoveDragState() {
 void EffectsGrid::OnScrollTimer(wxTimerEvent&)
 {
     MainSequencer* ms = static_cast<MainSequencer*>(mParent);
-    bool scrolled = false;
 
     if (mScrollDir != 0) {
         int cur = mSequenceElements->GetFirstVisibleModelRow();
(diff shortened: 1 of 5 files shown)


cybercop23:
The drag&drop only affects the "sequence" view, not the actual Master View itself, so there's no need to save layout.
Also it looks like any open sequence would mark rgbeffects dirty by using CurrentSeqXmlFile!=nullptr.

AGFazio:
> The drag&drop only affects the "sequence" view, not the actual Master View itself, so there's no need to save layout. Also it looks like any open sequence would mark rgbeffects dirty by using CurrentSeqXmlFile!=nullptr.

I was going by what currently happens when an model or group is moved within Edit Display Elements.  If you change the order of a view or add remove models/groups it marks it dirty and would require a Save. If we want the function to be just for the current sequence that is open, I can remove the mark dirty flagging.

derwin12:
I think a separate PR that can be done to wrap the dirty flag proc in multiple places...

```
        if (currentView != MASTER_VIEW) {
            MarkViewsChanged();
        }
```

AGFazio:
> I think a separate PR that can be done to wrap the dirty flag proc in multiple places...
> 
> ```
>         if (currentView != MASTER_VIEW) {
>             MarkViewsChanged();
>         }
> ```

Done. Removed mark-dirty.

AGFazio:
Between Claude and Copilot this is a mess. Will redo and resubmit.

PR #6560, opened 2026-06-16
VideoReader: fix sws_scale crash with CUDA/NVDEC hardware decoded video

AGFazio:
## Summary

- When a sequence containing Video effects is loaded, render threads open the video via `FFmpegVideoReader` using `h264_cuvid` (NVDEC on NVIDIA GPUs). The `h264_cuvid` decoder produces frames whose `format` field reports `AV_PIX_FMT_NV12` while the actual pixel data lives in CUDA device memory. The old hw-frame routing gate used `_srcFrame->format == __hw_pix_fmt` (`NV12 == CUDA`) which evaluated false, so `f` was set to `_srcFrame` (the GPU-backed frame). `sws_getContext` accepted NV12 as a valid software format, created a context, then `sws_scale` read from a CUDA device address as if it were CPU memory → access violation in `swscale-7.dll`.
- Fix: use `_srcFrame->hw_frames_ctx != nullptr` as the authoritative indicator that a frame is hardware-backed (regardless of reported pixel format), routing it through `av_hwframe_transfer_data` to obtain a real CPU copy before passing to `sws_scale`.
- Also guard against `av_hwframe_transfer_data` returning failure (abandon hw decode for that reader), and add a final safety check rejecting any frame still carrying `AV_PIX_FMT_FLAG_HWACCEL` before the `sws_scale` call.

## Test plan

- [ ] Load a sequence with Video effects using an MP4 on a machine with an NVIDIA GPU (NVDEC/h264_cuvid path) — previously crashed immediately on load
- [ ] Confirm `sws_scale` is called with `fmt=nv12` CPU data (visible in debug log) and no access violation occurs
- [ ] Verify video effects render correctly during playback
- [ ] Test on a machine without hardware acceleration (software decode path) to confirm no regression

### DIFF
--- a/src-core/media/FFmpegVideoReader.cpp
+++ b/src-core/media/FFmpegVideoReader.cpp
@@ -768,9 +768,6 @@ bool FFmpegVideoReader::readFrame(int timestampMS) {
                     }
                 }
                 if (f != nullptr && _swsCtx != nullptr) {
-                    spdlog::debug("VideoReader: sws_scale fmt={} {}x{} linesize={} -> fmt={} {}x{} linesize={}",
-                        av_get_pix_fmt_name((AVPixelFormat)f->format), f->width, f->height, f->linesize[0],
-                        av_get_pix_fmt_name(_pixelFmt), _width, _height, _dstFrame2->linesize[0]);
                     sws_scale(_swsCtx, f->data, f->linesize, 0,
                         f->height, _dstFrame2->data,
                         _dstFrame2->linesize);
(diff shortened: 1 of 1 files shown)


AGFazio:
**Re: stale _dstFrame2 on transfer failure** — Good point in principle. When `av_hwframe_transfer_data` fails we immediately set `_abandonHardwareDecode = true` and clear `_swsCtx`, so subsequent frames fall through to software decode. For the single dropped frame during that one-time state transition, leaving `_dstFrame2` holding the previous frame is acceptable — one held frame is preferable to a caller-API change. The next frame decodes cleanly on the CPU.

**Re: flush() on hot paths** — These `flush()` calls are on error/state-transition paths only (transfer failure, hw-frame safety-net guard), not in the per-frame normal decode path. They execute at most once per video file open, so there is no per-frame I/O overhead.

**Re: av_get_pix_fmt_name nullptr** — Valid. Fixed in 039cc93 by storing the result and using a `?: unknown` fallback before formatting.

Issue #6545, opened 2026-06-13
Make keybindings work across all shows


AGFazio:
Make keybindings be available across all show folders. Move from show folder into xlights install.

derwin12:
I dont think windows will like this...

derwin12:
The purpose of a showfolder is to contain everything for that show folder. May need to think this one over guys...

cybercop23:
Not into the install folder, but in the preferences. It's just a bunch of strings so same place as preferences.

focusedonsound:
@derwin12 @cybercop23 — before I put time into this, I want to check on something that I don't think actually got resolved in this thread.

@cybercop23's reply addressed the install-folder/Windows-permissions objection (Preferences instead of the install dir solves that). But @derwin12's other point — "the purpose of a show folder is to contain everything for that show folder... may need to think this one over" — never got a follow-up. That's a real design principle in this app (show folders are meant to be self-contained/portable), so I don't want to just quietly override it by moving keybindings to a global preference.

**Proposal: make it opt-in instead of a hard switch.** Keep today's behavior (keybindings stored per show folder) as the default — nothing changes for anyone who doesn't touch this. Add a preference, something like "Use these keybindings for every show" (off by default). When it's off, behavior is identical to today. When it's on, the keybindings editor reads/writes a copy in Preferences instead of the show folder's `xlights_keybindings.xml`, and that copy follows the user across every show they open. That way the self-contained-show-folder principle stays intact for anyone who cares about it, and the people who actually want one binding set everywhere (the original ask) get it without anyone else's behavior changing.

One more thing worth flagging: I noticed #6672 relocates the Key Bindings editor into a Preferences page as part of a larger dialog rewrite that's still being discussed. I'd scope this to just the storage layer (where the bindings live) and leave the editor's UI/entry point exactly where it is today, so this doesn't collide with that PR regardless of how it resolves.

Does the opt-in approach address the self-containment concern, or is there a reason even an opt-in global option is something we'd rather not have?

cybercop23:
There's already a PR that puts it in % AppData% like other options and also adds it to the startup and restore options to address that concern.

cybercop23:
In 2026.14. Change Preferences -> Other to choose the location.

PR #6571, opened 2026-06-18
Make KeyBindings generic not show folder specific (#6545)

cybercop23:
#6545 - now use the same location as preferences/options.


derwin12:
I think this is a step backwards ... what if you want specific halloween settings vs christmas settings. Or you host multiple shows. Wonder if there is a better solution here..

cybercop23:
I didn't think you'll want different key bindings for different shows, defeats the purpose. I can add the same logic as options, if there's one in the show folder use that over instead.

cybercop23:
new cone coming

PR #6573, opened 2026-06-18
Fix Last Input/Output Upload timestamp

jessica12ryan:
## Summary

This PR should be merged before the next release as it fixes an issue that was brought on with PR #6533.
Previously, the timestamps were stored in an AppData file for xLights. When a controller name changed or deleted, or a show folder changed or deleted, those stale timestamps would stay in the config. This PR will use a file within the show folder called xlights_local.json which will move with the show folder, so if the show folder is renamed or moved, timestamps will stay.

It also handles the renaming and deletion of controllers so that stale entries will not persist. This PR is required to prevent xLights from writing a bunch of data to an AppData file which may become stale and clog the configuration files.

This PR also adds support for iPad.

### Core (`src-core/`)

- **`src-core/utils/UtilFunctions.h`** — Replaced `fnv1a()` + `MakeControllerTimestampKey()` (global config, show-scoped via name-hash) with three helpers using per-show `xlights_local.json`:
  - `GetControllerTimestampPath(showDir)` — returns path or empty if no show directory
  - `ReadControllerTimestamp(showDir, prefix, ctrlName)` — reads timestamp string
  - `WriteControllerTimestamp(showDir, prefix, ctrlName, ts, validCtrls)` — writes and prunes stale entries
  Pruning: every write receives the full set of current controller names; any key matching `prefix_<name>` whose `<name>` isn't in the set is deleted.

- **`src-core/outputs/Controller.h`** — Added `[[nodiscard]]` to `SetExtraProperty`.

### Desktop UI (`src-ui-wx/`)

- **`src-ui-wx/app-shell/TabSetup.cpp`** — Property grid reads timestamps via `ReadControllerTimestamp` with pre-save `GetXLightsConfig()` fallback. Both `UploadInputToController` and `UploadOutputToController` write via `WriteControllerTimestamp` + live property-grid update. No `NetworkChange()` call — uploading never marks the show dirty.

- **`src-ui-wx/controllers/FPPConnectDialog.cpp`** — Both FPP and non-FPP branches use `WriteControllerTimestamp`. Renamed `controllerFailed` → `controllerCancelled` to match existing `cancelled` convention. `.ToStdString()` removed from `Controller::GetName()` (`GetName()` returns `std::string`, not `wxString`).

### iPad UI (`src-iPad/`)

- **`src-iPad/Bridge/XLSequenceDocument.mm`** — `runUpload:` writes timestamps on success. `controllerDetailForName:` reads timestamps into `lastInputUpload`/`lastOutputUpload` dict keys for SwiftUI.

- **`src-iPad/App/LayoutEditorView.swift`** — Extracts timestamps from the detail dict, passes them to `LayoutEditorControllerDetailView`, renders them after the models section via `TimestampRow` (new private struct).

### Documentation

- **`plans/ipad-parity/07-setup-controllers-upload.md`** — Row 56 updated from `ipad-missing` to `parity`.

## Key Design Decisions

- **Per-show JSON file** over global config: `xlights_local.json` lives inside each show directory. Portable across computers; deleted when the show folder is deleted; no cross-show key collision.
- **Flat key-value format**: `{"LastInputUpload_ControllerName": "2026-06-17 14:30:00", ...}`
- **Stale-entry pruning on every write**: Handles both rename and delete in a single pass.
- **Pre-save fallback**: When `showDirectory` is empty, timestamps fall back to `GetXLightsConfig()` with simple `"LastInputUpload_<name>"` keys.
- **No dirty flag**: `SetExtraProperty` is used only for pre-save fallback; the JSON file is the primary store — no `NetworkChange()`, no save prompt.
- **iPad parity achieved**: Same C++ helpers called from ObjC++ bridge; SwiftUI `TimestampRow` renders the result.


cybercop23:
That's the big design decision. Imagine having a timestamp on the Outputs page of the controller that shows you the last time channel data was updated. It doesn't care where it came from - it is just the last time it was updated. This should be the same.

If you switch to Christmas and upload, then go back to Halloween and upload (intentionally or not), it will show you it is *up to date* for Christmas, but it isn't. That's why it needs to be global, not show-specific, unless we scrap this whole thing and just "ping" the controller and compare its config with the current xLights config and flag any differences.

We want this to be a single version of the truth. Currently, FPP Connect keeps all ctrls and sequences ever uploaded into an ever-growing list. I wouldn't be concerned with that.

computergeek1507:
I was OK with the value stored in the settings/local computer(not great for people who use multiple computers) but like Alex said storing it in the show folder is not an acceptable solution for people show have multiple show folders.

jessica12ryan:
The issue that I was trying to solve is that every controller for every show folder is going to appear in the local settings file, and if you delete a show folder or move it to another system, all those stale entries will remain and never be deleted unless you delete xLights local data which is not typical.

So currently, if I create a show folder called "Test" and a controller called "Test123", and than I delete it, that will forever stay inside the local settings file with the last upload timestamp unless a new folder and controller are created with the same names and location. xLights never has any idea those folders were deleted. By storing it in the show folder, it moves and deletes with the show folder. I've also made changes to ensure renaming and deleting controllers will properly sync.

I have two different show folders for Christmas and Halloween, different props, different setup.... When I go to Halloween, I'm seeing the last time my Halloween outputs were uploaded, and same for Christmas. If I open it on my windows machine instead of my Mac, I still get the same timestamps and the information persists.

I honestly think that this is great for people with multiple show folders. When I'm working on Christmas, I'm not concerned when the last time I uploaded Halloween was, I'm more concerned about did I upload Christmas.... If the date is last year or 6 months ago, I know I didnt upload recently, whereas if it followed the controller it would show me "Last Upload - October 2025", so now im thinking my christmas outputs were uploaded in October because I'm in the christmas show folder, on the controller that has my christmas props, showing it was uploaded in October when that is not the case for this show folder or this controller. xLights controller properties load by show folder, so if the timestamps are going into the controller properties, it should be specific to that show folder.

Another issue you would run into without being show folder specific, is that if you load an imported show folder/sequence, and the controller on that show folder matches the same name, you're going to see that outputs were uploaded with incorrect information for that controller. By having it in the show folder, if you send that to someone else for testing, they'll know exactly when it was uploaded last, and could even solve the problem by opening it up on their machine and realizing that outputs were never uploaded for that controller.

This also opens functionality on iPad as well.

I can make changes as we see fit, but there is currently a stale data issue that should be resolved.

cybercop23:
I didn't look at the code in close detail, but how do you identify a controller? It should be name/ip, so the same one across shows folders or a base showfolder, would only be there once - if done right.

derwin12:
I concur with others .. the settings file is the place to be here. What is the real consequence of a couple of extra entries in the file - miniscule.  Folks dont normally rename and drop/recreate controllers all that often..

Issue #6528, opened 2026-06-10
Import effects - time limit


HansonElectronics:
When doing an Import Effects with xlights it would be nice if there was a checkbox to limit the imported effects to the duration of the sequence into which it's being imported.
A further tweak would be to only delete effects that begin after the duration of the sequence being imported into.
In some cases it can be handy to have access to effects that don't fit into the timeline so a hard cutoff would preferably be optional rather than default.

Xlights 2026.10
Windows 11 on a PC

cybercop23:
Effects past the end of the sequence will not play. They are there, but ignored. You can lasso multiple effects to delete them as needed.
What is the use case of importing something that's longer (or shorter) than what your current sequence?

computergeek1507:
Sometimes if the effects are too far over the time you can't see them in the hashed area to lasso. Like if it's a 30 sec sequence and you import a 3min sequence. I can see it's usefulness.

HansonElectronics:
I have a number of sequences with nice effects for a number of props that I use to import the effects from as a template. These sequences are typically made up to around 3.5 minutes. Some of the songs that I import the effects into are as short as 1:20. 
I found that sometimes I can't zoom out far enough to be able to lasso effects that are past the songs finish time by more than a few seconds. When zooming out as soon as all the song duration is reached I can't zoom out any further. If there is a method to zoom out even further that would be handy to know.

cybercop23:
Sounds like a more generic change would be to allow zooming bound to the last effect vs duration.
Your time limit clip may work if you know/remember the duration of the source sequence; even so, you may not be able to get to all effects.

dkulp:
Likely a tool to trim effects to sequence length as well.   Just loop through all the layers and remove extra effects and adjust the length of the last one if it exceeds.

THAT said, adjusting the length of effects is not trivial.  If an effect has a "cycles" type of thing, making the effect shorter can make the effect go really really fast.    Not sure we want to attempt to compensate for that or not when adjusting the length.

derwin12:
If you import a sequence and there are effects that far off from your audio -- there are likely other issues :-) Like you used the wrong song. :-)

derwin12:
..and keep in mind I and perhaps many others .. create an animation and import the vendor sequence before acquiring the audio. So the 30s is there until I see the length of the song. Note too that effects can be slightly longer than the song as the song length may vary. (These effects are noted in a check sequence)

AGFazio:
[...]

PR #6597, opened 2026-06-23
Allow timeline to zoom out to see any effects outside the duration (#6528)

derwin12:
If there are effects misplaced outside the duration, they are not accessible. It wont remove them automatically as there may be a reason one has them out there (placeholders? a work area - who knows)


derwin12:
Pushed too much...

Issue #6611, opened 2026-06-25
Rendering of sub-models


wease944:
Upgraded from 2025.08-2026.11
1) EFL Pumpkin; for the JakOLantern----normally put a fire effect on the jkoltrnFireEffect and a face effect (jkoltrnFaceeffectA) on jkoltrnFaceEffect.....creates a nice burning jack o latern, now the entire prop shows with fire

2) usually use liquid on the EFL_Ghost outline and leave the render style to default, now it must be changed to per preview to show more than 4 pixels

Happy to discuss in Zoom room
steve-pa

derwin12:
Create a demo seq and package the sequence..and upload.

wease944:
<img width="955" height="447" alt="Image" src="h

[jkoLtrn.zip](https://github.com/user-attachments/files/29385165/jkoLtrn.zip)

ttps://github.com/user-attachments/assets/a356f1ba-7baa-43f4-9a91-baf980f24f87" /> 

I was unable to duplicate the Ghost outline issue, but you can see in the picture what the jkOLtrn looked like with export from 2025.8

derwin12:
The image didnt get uploaded.

derwin12:
Broken between .09 and .10

derwin12:
So there was a change where faces and states on submodels would use the submodel numbering .. that broke these faces.

wease944:
Thank you

derwin12:
You saw Jacky's wizardry - so the new method is working as expected and he will update the model so the past trickery is no longer needed.

PR #6615, opened 2026-06-27
Fix for #6611 - Face Effect on submodels was using the wrong nodes *DONT MERGE*

derwin12:
This was a breaking change - the face definitions are still using the full model numbering. It corrects the reported issue but perhaps @cybercop23 wants to confirm it still works with his states and faces change.


derwin12:
Not very confident in this one...

PR #6670, opened 2026-07-08
Add live filter to the Choose Model(s) checklist dialog

heffneil:
## What

Two usability improvements to the **Choose Model(s)** dialog shown by **Copy Layers/Submodels → to Models** (`CheckboxSelectDialog`):

### 1. Donor model shown in the header
The dialog now shows which model you are copying **from** at the top:

> Copying layers/submodels from: **&lt;model name&gt;**

("Copying layers/submodels from:" in normal weight, the donor model name in **bold**.) Previously the dialog gave no indication of the source model once it opened.

### 2. Live filter box
A **filter box** above the checklist so long model lists can be narrowed as you type:

- Debounced (200 ms) `wxSearchCtrl` with a cancel button and "Filter" placeholder.
- Match is **case-insensitive**, whitespace-tokenized **AND** (e.g. `arch left` matches "Left Arch Mega").
- Check state is held in a persistent set that is the source of truth, so:
  - toggles survive filter rebuilds, and
  - a model checked, then hidden by a narrower filter, is **still returned** by `GetSelectedItems()`.
- Select All / Deselect All / (De)select Highlighted right-click actions and OK all reconcile through that set.

Also fixes the Ok/Cancel buttons being clipped at the bottom — the sizer hints are now recomputed after the header/filter rows are inserted (the generated code had sized the dialog before those rows existed).

## Scope / parity

Desktop-only: this is a wxWidgets dialog (`src-ui-wx/shared/dialogs/`) with no iPad counterpart — the Copy-Layers-to-Models flow is desktop-only. No `src-core/` changes.

## Notes

- Edits are entirely **outside** the wxSmith `//(* … *)` guards, so no `.wxs` change is required.
- The new header is an optional `CheckboxSelectDialog` parameter (normal prefix + bold value); only the Copy-Layers caller passes it.
- Builds clean on macOS (Release, arm64).

### DIFF
--- a/src-ui-wx/sequencer/EffectsGrid.cpp
+++ b/src-ui-wx/sequencer/EffectsGrid.cpp
@@ -8702,7 +8702,8 @@ void EffectsGrid::CopyModelEffectsToModels(int row_number) {
         return;
     }
 
-    CheckboxSelectDialog dlg((wxWindow*)mParent, "Choose Model(s)", choices);
+    CheckboxSelectDialog dlg((wxWindow*)mParent, "Choose Model(s)", choices, wxArrayString(),
+                             _("Copying layers/submodels from: "), wxString(source_name));
     dlg.SetMinSize(wxSize(400, -1));
     dlg.SetSize(wxSize(400, dlg.GetSize().GetHeight()));
     if (dlg.ShowModal() != wxID_OK)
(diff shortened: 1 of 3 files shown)


heffneil:
Screenshot 2026-07-08 at 11 24 25 AM
Screenshot 2026-07-08 at 11 24 34 AM

heffneil:
Filter king here: we can't expect people to sort through lists.  Picking a letter to get to the right name - isn't really usable.  Also it doesn't filter words within the string.  If I have an EFL Showstopper it may be named showstopper or EFL but I wouldn't know for sure until I try E or S and drive down the alphabetized list.

PR #6673, opened 2026-07-08
headless: non-interactive font mapper (+ Windows build fixes)

derwin12:


derwin12:
Hopefully folks are okay with this ... it allows me to compile and I dont get missing fonts for a couple of sequences in headless render.

PR #6685, opened 2026-07-13
Add effect and xLights UI plugin architecture, with a -sm safe-mode switch

focusedonsound:
## Summary

Implements the plugin architecture originally proposed in
[discussion #6368](https://github.com/xLightsSequencer/xLights/discussions/6368).
Adds a plugin architecture so third-party DLLs/dylibs/`.so`s can extend xLights at
runtime without being merged into the main build:

- **Effect plugins** register new `RenderableEffect` types, appearing in the
  Effects toolbar and Effects panel exactly like a built-in effect.
- **xLights UI plugins** add dockable AUI panes and `Tools > Plugins` menu
  entries through an abstract widget facade — plugin DLLs never link wxWidgets
  directly, avoiding a second, non-communicating copy of wx's statically-linked
  runtime inside the host process.
- A new `-sm`/`--safemode` startup switch skips loading both plugin kinds, for
  isolating "is this a plugin problem or a core problem" when troubleshooting.
  This was **@jessica12ryan**'s suggestion in that discussion — a safe-mode
  launch flag so support/zoom helpers can ask "have you tried safe mode?" to
  quickly tell a plugin-caused issue from a real core bug, instead of digging
  for hours first.

Ships with three example plugins (Voronoi Cells effect, a minimal sample
xLights plugin, and an Effects Toolbar Manager) and a Preferences > Plugins
panel for managing what's loaded.

## Screenshots

**Preferences > Plugins > Effect Plugins** — Voronoi Cells loaded from `effect_plugins/`:

Effect Plugins tab

**Preferences > Plugins > xLights Plugins** — Effects Toolbar Manager loaded from `xlights_plugins/`:

xLights Plugins tab

**Effects Toolbar Manager plugin in action** — a dockable dialog built entirely through the widget facade (`IPluginPanel`/`IPluginListBox`), no direct wx linkage:

Manage Effects Toolbar

## How it works

- **Effect plugins** (`src-core/effects/effectPlugin.h`): a plugin DLL exports
  `xlCreateEffectPlugin`/`xlDestroyEffectPlugin`; `EffectManager::loadEffectPlugins`
  scans an `effect_plugins/` folder next to the executable and registers each
  one starting at `eff_LASTEFFECT`.
- **xLights UI plugins** (`src-ui-wx/plugins/`): a plugin DLL implements
  `IXLightsPlugin`; `XLightsPluginManager`/`XLightsPluginHost` load DLLs from an
  `xlights_plugins/` folder *after* `CreateSequencer()` (so `m_mgr`/the Tools
  menu already exist) and hand each plugin a curated widget facade
  (`IPluginPanel`/`IPluginListBox`/etc.) to build its UI against.
- **Safe mode**: `-sm`/`--safemode` sets `_safeMode` on `xLightsFrame`, which
  short-circuits both `loadEffectPlugins()` and `XLightsPluginManager::loadPlugins()`.
  The title bar shows `[SAFE MODE - plugins not loaded]`, applied consistently
  across startup, sequence open, save-as, and close via a shared
  `SafeModeTitleSuffix()` helper (this needed 4 call sites, not 1 — the title
  gets reset on every file-lifecycle event).

## Implementation

- New shared `PluginLoader<T, Context>` template
  (`src-core/utils/PluginLoader.h`) wraps `dlopen`/`LoadLibrary` +
  create/destroy export lookup once; both `EffectManager` (new) and
  `ServiceManager`'s AI plugin loading (refactored onto the template, -131/+~30
  net in `ServiceManager.cpp`) now share it instead of each hand-rolling
  platform `#ifdef`s.
- **Plugin-loading order fix**: effect-plugin discovery now runs *before*
  `AddEffectToolbarButtons()` and `effectPanelManager.Init()` in the
  `xLightsFrame` constructor. Previously the toolbar/effects-panel buttons were
  built straight from the 56 built-in effects with no rebuild afterward, so a
  plugin effect (Voronoi Cells) silently never got a button anywhere — this is
  what surfaced the whole investigation.
- `CoreExport.h`/`pluginExport.h` — shared cross-platform symbol-export macros
  for `src-core/` plugin ABIs.
- `EffectsGrid`/`EffectIconPanel`/`EffectIconCache`/`BitmapCache` changes
  support plugin-effect icons through the same render/cache paths built-ins use.
- `plans/ipad-parity/` (themes 04, 12, 13 + `00-overview.md`) updated: all
  three plugin-loading surfaces (AI/effect/xLights-UI) are marked infeasible on
  iPad — App Store policy forbids loading externally supplied executable code
  at runtime, and the xLights-UI-plugin widget facade additionally has no wx to
  wrap on iPad even setting code-loading aside.

## Files changed

- `src-core/utils/PluginLoader.h`, `CoreExport.h`, `pluginExport.h` — new
  shared plugin-loading infrastructure
- `src-core/effects/effectPlugin.h`, `EffectManager.{h,cpp}` — effect-plugin ABI
  + loading
- `src-core/ai/ServiceManager.{h,cpp}`, `aiPlugin.h` — refactored onto the
  shared loader
- `src-ui-wx/plugins/` (new) — `XLightsPluginManager`, `XLightsPluginHost`,
  `xLightsPlugin.h` (the `IXLightsPlugin`/widget-facade API)
- `src-ui-wx/preferences/PluginsSettingsPanel.{h,cpp}` (new) — Preferences >
  Plugins
- `src-ui-wx/xLightsMain.{h,cpp}`, `xLightsApp.cpp` — safe mode wiring,
  plugin-load ordering fix, `RebuildEffectsToolbar`, title-bar suffix
- `xLights/VoronoiCellsPlugin/`, `SampleXLightsPlugin/`,
  `EffectsToolbarManagerPlugin/` (new) — example plugins + `.vcxproj` targets
- `xLights/xLights.cbp`, `Xlights.vcxproj(.filters)`, `xLights.sln` — new
  project/target registration
- `plans/ipad-parity/00-overview.md`, `04-effects-and-panels.md`,
  `13-tools-diagnostics-help.md` — parity scorecard updates
- `README.txt` — release note

## Test plan

- [x] Windows Debug/x64 build green (`xLights.sln`)
- [x] Launched with `-sm`: confirmed plugins are not loaded and the
      `[SAFE MODE]` title marker persists across sequence open/save-as/close
- [x] Confirmed Voronoi Cells plugin effect icon now appears in the Effects
      toolbar and Effects panel without `-sm`
- [ ] macOS Debug build + `xLights-iPadLib` build (touches `src-core/`) — not
      yet run, no macOS environment available in this session
- [ ] Linux build


dkulp:
Developers have discussed this and this is something we are NOT interested in adding into xLights.   Anything that would cause xsq files to not render completely and properly on EVERYONES xLights instances is not something we want.  If there is some new effect or something that is really cool? Great!  Submit it to include with xLights so everyone can benefit.

focusedonsound:
Appreciate the quick response, but I want to push back a bit.

I opened discussion #6368 in May specifically asking for feedback before I put real time into it, including volunteering to build the initial framework myself. It sat for months with essentially no engagement from the core team — no "we wouldn't accept this," no "here's the concern," nothing. I took that as tacit permission to go build it.

I then spent real time implementing it — effect plugins, a UI plugin framework, a safe-mode switch (which was directly a response to feedback in that same discussion, from @jessica12ryan), example plugins, docs. The PR was open and out of draft for about a a minute before it was closed.

The .xsq portability concern is legitimate and I'd have engaged with it seriously if it had come up in the discussion, or even as a review comment instead of a close. A few things I'd genuinely like to know:

Was there ever a design that would have addressed this — e.g. the sequence recording which plugin effects it depends on, and xLights warning (not silently rendering wrong) when a required plugin is missing — or is a plugin architecture a non-starter regardless of implementation?

- If it's the latter, I'd rather have heard that in May than after building it.

I'm not trying to relitigate the decision if the answer is genuinely "no plugin architecture, period" — I just don't think a multi-month-old discussion with an open call for feedback, followed by a close within a minute on the implementation, is a fair way to treat a contributor's time.

computergeek1507:
I wouldn't assume no answer is a "go for it". Plugins are not usable on any platform besides windows, which maybe not be possible in a certain versions either.

focusedonsound:
@computergeek1507 — to close the loop on the cross-platform question: you're right that I hadn't verified this on an actual macOS build. I checked the project config directly — Hardened Runtime is enabled (`ENABLE_HARDENED_RUNTIME = YES`) and `xLights.entitlements` doesn't include `com.apple.security.cs.disable-library-validation`. That means Library Validation is on by default, and a third-party plugin dylib signed with a different Team ID would likely fail to `dlopen` on a real notarized build. That's a legitimate gap I should have called out instead of leaving the macOS row unchecked in the test plan.

Worth noting: this isn't specific to this PR. `PluginLoader<T,Context>` is a refactor of the exact `dlopen`/entitlements setup `ServiceManager` already uses today for AI service plugins — so this same question already exists for a feature that's shipped. I'm happy to submit a small, separate fix for that entitlement regardless of what happens with the rest of this, since it's worth resolving on its own.

@dkulp — on the bigger question: I want to understand exactly what the objection is targeting, because I think there might be a narrower version of this worth considering.

Your concern was specifically about `.xsq` files not rendering identically across everyone's xLights install. That's a real risk for **effect plugins** — a sequence using a plugin-provided effect genuinely can't render the same way on an install without that plugin. I don't have a way to fully engineer that risk away; the closest I can think of is xLights refusing to render / warning loudly when a sequence references an effect it doesn't have, instead of silently rendering wrong, but that's a mitigation, not an elimination of the risk.

That said, the **UI/UX plugin half is a separate system that never touches rendering or `.xsq` output at all** — it only adds dockable panels and Tools-menu entries (things like the Effects Toolbar Manager example, which just reorders/filters which built-in effect icons show up). A sequence file is identical regardless of which UI plugins someone has installed. It genuinely can't cause the failure mode you described.

So: would a rendering-safe, UI-only plugin system — no new effect types, nothing that changes what a sequence renders, just a way to add tool panels and menu entries — be something the team would consider? If so, I'd rather resubmit narrowly scoped to just that (plus the safe-mode switch, which is orthogonal either way) than assume the door's closed on everything. If the answer is no on the concept regardless of rendering safety, that's useful to know too — I'd rather hear that now than build it and find out again.

dkulp:
No one except the actual xLights developers should have been creating "Discussions".   The main reason is that none of the developers are actually following the discussions ore getting notifications about the discussions.   It was setup so if WE wanted to extend a discussion out to other and thus should not have been set to allow anyone to create them.   That's now fixed.   So basically, no one responded because no one knew the discussion was even there.   I would have filed a note completely against it back then if I had known it was there.    

Concretely:  I'm 100% against any generic plugin  anything at this point.  Without a super concrete reason for needing to implement something as a plugin as opposed to just putting it in xLights directly, I will completely push back on it.   I'm also against the plugin stuff that was added for the AI stuff.  The concrete reason there was to avoid shipping giant dll's, which is almost reasonable, but even then if you compare the size of xLights to any modern application, we could grow 10 fold and still be less the most apps.    But in any case, the plugins stuff for AI is also NOT documented and NOT part of the releases notes at all so  we haven't really considered it a shipping feature.    Even Scott's commit specifically says "Experimental ServiceManager plugin interface for AI services".     *EXPERIMENTAL*

Specifically for your use case, if you wanted to add a Preferences panel that allows customizing the toolbars, I'm OK with that.   No plugin needed.   That is genuinely a missing feature.  Could even start with JUST the effects toolbar which is now too long to even fit on smaller screens without setting the icon size way down.   

Beyond that, if you wanted to come up with some sort of json description of a "tool" -> "api" mapping or something that we can load from the show folder or settings locations and wire up new tools or panels, I would consider that.   At this point, I'm not considering anything that would load or require a .dll/.so/.dylib without some really serious need.  It opens things up for way too much possible abuse, risk, etc...    That said, any such dll/so/dylib would likely have to be GPL anyway, so again, no reason to not just submit it to xLights assuming it's actually a good idea.

focusedonsound:
That's a genuinely helpful answer, thanks for the detail.

The Discussions explanation clears up the part that was actually bothering me — I read the silence as "nobody objects," when it sounds like nobody even saw it. Good that it's fixed so that doesn't happen again to someone else.

On plugins: understood, and I won't push on it further. The abuse-surface/risk argument for blocking arbitrary native code loading applies regardless of what the code does, so I get why UI-only doesn't change the calculus.

I'll take you up on the toolbar customization panel instead — starting with just the Effects toolbar, since that's the concrete pain point you called out. I already have working logic for choosing/reordering which effects show up (built as part of the plugin example), so it's a small lift to reimplement it as a native Preferences panel calling straight into `EffectManager`/the toolbar rebuild, no plugin involved. I'll open a separate, narrowly-scoped PR for that.

Closing this one out.

derwin12:
Is there really a need to have a preference to remove like 4 effects when so many folks dont even use effects and just map.  And they are alphabetical and that helps to keep consistent. Not sure a big pending ask for this either.

cybercop23:
One issue I have at times is effects running off the screen since I have all my toolbars all on the same line.
I was thinking of something similar, a custom toolbar where you can choose from any of the toolbars and mix your own, and have them in the order you want.

derwin12:
Take a look at ac toolbar. You can put some less popular ones on a toolbar that can be moved or not shown

PR #6686, opened 2026-07-14
Add Preferences > Toolbars: customize the Effects toolbar

focusedonsound:
## Summary

Adds Preferences > Toolbars: lets users hide/reorder entries in the
always-visible Effects toolbar, persisted per user.

This is a narrower re-scope of #6685 (the plugin architecture proposal),
which was closed over its `.xsq`-portability and native-code-loading risk.
This PR drops all of that — no plugin, no DLL/dylib/so loading of any kind —
and implements the specific concrete feature @dkulp said he'd be OK with:
"if you wanted to add a Preferences panel that allows customizing the
toolbars, I'm OK with that... Could even start with JUST the effects
toolbar which is now too long to even fit on smaller screens without
setting the icon size way down."

## Screenshots

**Preferences dialog** — new "Toolbars" entry alongside the existing pages:

Preferences dialog

**Toolbars page** — the Effects toolbar checklist + reorder control (top of list):

Toolbars page, top

**Toolbars page, scrolled** — same list further down, showing hidden effects (Glediator, Moving Head, Servo) staying listed but unchecked:

Toolbars page, scrolled

## How it works

- Preferences > Toolbars shows every registered effect in a `wxRearrangeCtrl`
  (checkbox list + Up/Down reorder buttons) — hidden effects stay listed,
  unchecked, so they can be re-enabled later.
- OK applies the new order/visibility to the running Effects toolbar
  immediately, no restart needed.
- The layout persists across restarts the same way `mIconSize` and other
  toolbar-adjacent preferences already do: written once in
  `~xLightsFrame()`, not on every keystroke.
- Hiding an effect only affects the toolbar's button set — the effect stays
  fully available via the Effects panel and right-click "change effect
  type." Nothing about rendering, `EffectManager`, or `.xsq` content changes.

## Implementation

- `src-ui-wx/preferences/ToolbarLayout.h/.cpp` (new) — two free functions,
  `LoadToolbarLayout`/`SaveToolbarLayout`, persisting an ordered
  `(name, visible)` list as one JSON-array config string via
  `XLightsConfigAdapter`. Generic over a config key + candidate name list
  (not Effects-specific in shape) so a second customizable toolbar could
  reuse it later without a redesign — nothing else calls it yet, and the
  app's other toolbars are hand-built from heterogeneous, stateful items
  rather than a uniform named list the way `EffectManager` gives us for
  effects, so generalizing further isn't attempted here.
- `xLightsMain.h/.cpp` — `xLightsFrame::RebuildEffectsToolbar()` replaces the
  old unconditional `AddEffectToolbarButtons()`, filtering/ordering by the
  saved layout instead of always showing all effects. Loaded at startup,
  saved in `~xLightsFrame()` alongside the existing `config->Write(...)`
  block for `mIconSize` etc.
- `src-ui-wx/preferences/ToolbarsSettingsPanel.h/.cpp` (new) — the
  Preferences page itself. Hand-written (not wxSmith), same reasoning as the
  existing `PluginsSettingsPanel`: a dynamically populated list, not a fixed
  settings form.
- `xLightsPreferences.cpp` — registers the new "Toolbars" page alongside the
  existing ones.
- `plans/ipad-parity/04-effects-and-panels.md` + `00-overview.md` — new gap
  row: iPad's `EffectPaletteView` has no equivalent order/visibility
  control. Not addressed here — `src-ui-wx/` (and therefore this feature)
  isn't part of the iPad build at all, and closing that gap would be a
  separate SwiftUI effort.

## Files changed

- `src-ui-wx/preferences/ToolbarLayout.h`, `ToolbarLayout.cpp` (new)
- `src-ui-wx/preferences/ToolbarsSettingsPanel.h`, `ToolbarsSettingsPanel.cpp` (new)
- `src-ui-wx/preferences/xLightsPreferences.cpp` — registers the new page
- `src-ui-wx/xLightsMain.h`, `xLightsMain.cpp` — `RebuildEffectsToolbar()`,
  layout load/save wiring
- `xLights/Xlights.vcxproj(.filters)`, `xLights/xLights.cbp` — new file
  registration (Windows + Linux `.cbp`; macOS Xcode auto-discovers files
  under `src-ui-wx/`, no project edit needed there)
- `plans/ipad-parity/00-overview.md`, `04-effects-and-panels.md` — parity
  scorecard update
- `README.txt` — release note

## Test plan

- [x] Windows Debug/x64 build green
- [x] Preferences > Toolbars lists every effect, checked state matches the
      current toolbar (including previously-hidden ones, shown unchecked)
- [x] Uncheck/reorder, OK — toolbar updates immediately, no restart needed
- [x] Close and relaunch xLights — layout persists (save/load round trip,
      not just the live-apply path)
- [x] A hidden effect is still fully usable via the Effects panel and
      right-click "change effect type"
- [ ] macOS Debug build — not run, no macOS environment available this
      session. No Windows-specific APIs are used; the one new control,
      `wxRearrangeCtrl`, only depends on `wxUSE_CHECKLISTBOX` (confirmed on
      for Windows in `wx/chkconf.h`; no reason to expect it's disabled
      elsewhere), but this is unverified rather than assumed.
- [ ] Linux build


cybercop23:
How about adding an "overflow" menu "..." at the end that will show all of the unselected ones? That way, they will always be accessible(just an additional click) for those who want them but don't have the screen size for them.

dkulp:
Bah... there is always View -> Windows -> Effect Dropper if you need/want access to all of them.

cybercop23:
Sure... I can hear @derwin12 saying... not everyone has a 4k screen.

focusedonsound:
The vast majority of people may not use dmx, or servos, or other effects which was the initial reason for this.  If adding an additional ... Or sub menu option to the effects menu is an additional expansion to this PR than I think this is a different discussion point for a future PR 

The general idea is to allow users to customize the top tool bars to their workflow and maximize screen space.  Adding the ability to hide / reorder the other tool bars should be considered in addition to this as it adds to the overall flexibility.

computergeek1507:
Should add checkbox columns for Alex's wheels and if they are randomable, I don't understand how to configure the wheel with key bindings.

dkulp:
Personally, I think the wheel should be a separate page (likely not part of this PR) that gives you 16 options (or however many wheel segments) with each being a drop down where you can select an individual effect OR select an Effect Preset....  But that's me.   A quicker way into the highly used presets without affecting my existing key bindings would be beneficial.   That said, that's likely a completely different discussion than what this PR is for.

cybercop23:
> Should add checkbox columns for Alex's wheels and if they are randomable, I don't understand how to configure the wheel with key bindings.

@computergeek1507  It takes the first 16 keybindings effects. - you knew that. :)

derwin12:
Could we just paste an effect in the wheel config..which is pretty much what wa preset is..I guess to that might not be intuitive

deplanche:
Would it be possible to add a right click over the effects icons in the menu bar to bring up this menu?  Sort of like many programs have the editable ribbons or menu bars, where you can pick and choose what you see.

Would also be nice if it had a reset to put them all back on and in the default order (mostly alphabetical, except for the Off and On).

focusedonsound:
@deplanche Both of these are in now:

- **Right-click the Effects toolbar** to jump straight to Preferences > Toolbars (on Windows/Linux; on macOS it opens Preferences, just not pre-selected to that page — the native `wxPreferencesEditor` has no page-select API, so that one extra click remains there).
- **"Reset to Defaults"** button in the Toolbars page restores every effect, visible, in the default order.

Both apply live (no OK/restart needed), same as the rest of the panel.

PR #6689, opened 2026-07-14
Media panel: always offer Bulk Find, not just when files are missing

### LINKED ISSUE #6658: Always offer to bulk edit files in the media panel
We currently only offer the bulk edit if there are files not found.. but folks are trying to repoint them to a new directory  even if they are currently in  a valid directory. 2025 vs 2026 for example.  The current workaround is to rename 2025 so that it explicitly fails the file lookup.

focusedonsound:
## Summary

Fixes #6658. The Manage Media panel (Sequence Settings > Media) only offers
"Bulk Find `<Type>`s..." when something is already broken, and only searches
for/relinks broken entries. Redirecting a whole set of already-working files
to a new folder (e.g. last year's show copied forward to this year) required
first renaming something to force a "not found" state — the workaround
described in the issue.

## How it works

- "Bulk Find `<Type>`s..." is now offered whenever there's more than one item
  of a media type, regardless of whether any of them are currently broken.
- The search/relink pass now runs over *every* item of that type (matched by
  filename against the folder you pick), not just broken ones — so a valid
  file gets repointed too if a same-named file is found in the new folder.
- "Re-select `<Type>`..." (the single-item fix action) is unchanged — still
  only offered for a broken item.

## Implementation

- `src-ui-wx/media/ManageMediaPanel.cpp`:
  - `OnTreeContextMenu`: decoupled the "Bulk Find" menu item from the
    right-clicked item's broken state; it's now gated only on `> 1` items of
    that type existing (image count via `GetImagePaths().size()`, other
    types via a count over `GetAllMediaPaths()`).
  - `OnBulkFindImages` / `BulkFindMediaByType`: the path-collection step no
    longer filters to `!entry->IsOk()` — collects every path of the type, so
    a filename match in the target folder relinks it whether or not the
    current path still resolves. Folder-picker prompt text updated
    ("Select folder to bulk-find images in") since it's no longer
    missing-only.
- `plans/ipad-parity/01-file-lifecycle.md` + `00-overview.md`: the
  corresponding iPad `MediaManagerSheet` "Bulk Find Missing…" action is still
  `.disabled(brokenCount == 0)` and broken-only, so this is a new
  `ipad-weaker` gap (was `parity` since both had the same limitation before).
  Recorded with `file:line` evidence; not addressed here.

## Test plan

- [x] Windows Debug/x64 build green
- [x] Walked through it live with a real sequence: right-clicked a
      **currently-valid** image with more than one image in the sequence —
      "Bulk Find Images..." now appears (previously it wouldn't, since
      nothing was broken)
- [x] Confirmed a genuinely broken item still offers "Re-select..."
      unchanged
- [ ] macOS / Linux build — not run, no environment available this session;
      no platform-specific code touched


derwin12:
Does  this try to remap everything in the media panel or just the one that is selected or have multiple items selected?

focusedonsound:
Everything of that media type in the sequence — not just the selected item, and there's no multi-select scoping. Right-clicking any image (or shader/video/etc.) just determines *which type's* "Bulk Find..." shows up; once you pick a folder, it tries to match every image (or every item of that type) in the sequence against filenames in that folder, and relinks whichever ones it finds a match for. Anything it doesn't find a match for is left alone and reported as "not found" in the summary.

PR #6727, opened 2026-07-21
Add models/controllers and xlDo API metadata

cybercop23:
This allows agents to manage show folder items, create and edit models.

Can you add five 3-layer Stars to my show layout (xLights_Testing/Playground), with the layers being 50 inside, 60 middle, and 70 outside?
Please convert the matrix in my current show folder to a custom matrix so I can have individual star channels and set them to 250 per port, except the last one at 200.
Please create a group of the two matrices, called 'All Matrices'

image

And the crazy part is that it used this info to add a DMX model and test the last fix by actually creating a sequence and validating channel data values.

### DIFF
--- a/resources/controllersmetadata/Ethernet.json
+++ b/resources/controllersmetadata/Ethernet.json
@@ -1,0 +1,93 @@
+{
+    "schemaVersion": 1,
+    "objectType": "controller",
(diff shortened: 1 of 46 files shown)


cybercop23:
These could be pulled into an MCP that can fully manage the show folder. The idea is that we can add Bulbby as a way for folks to just instruct it what to do.

Issue #6720, opened 2026-07-21
Save Sequence then Close Sequence still thinks the sequence is dirty


derwin12:
In master/nightly ..

Could now be a timing thing .. 

Save a sequence then when it completes go to close the sequence. You often get prompted to save it again.

PR #6730, opened 2026-07-21
Possible Fix for Dirty Save (#6720) Dan/Scott please review

derwin12:
I have been getting a save (again) when I close after doing a save button. This adds a wait/delay ..

Does this seem like a good thing - making folks wait in the ui for the save to finish?

### DIFF
--- a/src-ui-wx/app-shell/TabSequence.cpp
+++ b/src-ui-wx/app-shell/TabSequence.cpp
@@ -1732,6 +1732,16 @@ void xLightsFrame::SaveSequence()
             mSavedChangeCount = _sequenceElements.GetChangeCount();
             mLastAutosaveCount = mSavedChangeCount;
         } );
+
+        // RenderGridToSeqData is async (finishes via the callback above once the
+        // render engine drains). Callers that just want "SaveSequence() returned
+        // means it's saved" (Ctrl+S, the toolbar Save button, CloseSequence's own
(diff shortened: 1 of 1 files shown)


src-ui-wx/app-shell/TabSequence.cpp
```diff
@@ -1732,6 +1732,16 @@ void xLightsFrame::SaveSequence()
             mSavedChangeCount = _sequenceElements.GetChangeCount();
             mLastAutosaveCount = mSavedChangeCount;
         } );
+
+        // RenderGridToSeqData is async (finishes via the callback above once the
+        // render engine drains). Callers that just want "SaveSequence() returned
+        // means it's saved" (Ctrl+S, the toolbar Save button, CloseSequence's own
+        // save-and-close path) would otherwise race the callback and see
+        // mSavedChangeCount still stale, spuriously re-prompting to save on close.
```
dkulp, line 1740: The main comment I have is that EnableSequenceControls(false) prior to this should have disabled all the controls until the render is done.  Thus, I'm not sure how a subsequent "CloseSequence" or whatever would be triggered in a state where mSavedChangeCount is incorrect.

derwin12: ok .. I will try to keep a close eye out for the triggering condition.

PR #6733, opened 2026-07-22
Vendor model catalog: hierarchy-preserving live filter (replaces #6256)

heffneil:
Replaces #6256 per @dkulp's review there: squashed history, description matches the code, every review item addressed.

**Update per review below: the Vendor → Category → Model hierarchy is fully retained.** The earlier flat-list revision was reworked after vendor feedback that the categorical display is a hard requirement.

### 1. Live filter on the existing search control

The existing `TextCtrl_Search` (a `wxSearchCtrl` since #6267) is **repurposed in place** and moved above the tree — one control, both behaviors:

- **Typing filters live** (debounced 200 ms): whitespace-tokenized, case-insensitive AND across vendor / category path / model / wiring names — `pumpkin boscoyo` works.
- **Matches stay inside their category structure**: branches with no matching models disappear, surviving branches auto-expand (a bounded set — the unbounded whole-catalog expansion that hung Win32 in #6256's early iterations never happens; unfiltered builds expand nothing, same as master).
- **Enter** applies immediately; **cancel (✕)** restores the full tree.
- **Magnifier button** keeps the step-through **next-match** navigation, now within the filtered tree (under an active filter it steps through the result leaves, since a raw substring test can't match a multi-token query). #6256 had hidden this control, orphaning the feature.
- Matching runs against **lower-cased per-category-path and per-model text built once at load** — no per-keystroke `GetPath()` recursion, no map rebuilds, no string concatenation per model per rebuild.

### 2. Browsing is unchanged, with two small fixes

With no filter typed the tree is identical to master — vendors, categories, sub-categories, multi-wiring model nodes with wiring children, empty categories pruned. Two additions:

- Models are **sorted case-insensitively** within each category (previously raw XML order).
- Models with **no resolvable category** — previously unreachable through the category walk — are listed directly under their vendor.

### 3. `.wxs` back in sync

`VendorModelDialog.wxs` had drifted from the generated code twice (#6267's wxSearchCtrl conversion and #6073's `wxTR_MULTIPLE`). It now matches: `wxSearchCtrl` above the tree with all four handlers declared, `Button_Search`/`FlexGridSizer9` removed, tree style corrected.

### 4. Review items from #6256

- All `spdlog::info` tracing removed.
- No runtime `Hide()` of the search control (option (a) as preferred).
- Stale `wxTreeItemId` use: `_lastSearchItem` is dropped before every rebuild and selection-change events during a rebuild are ignored (`_treeRebuilding` + RAII scope) — no reads of deleted items.
- The `GetFocusedItem()` fallback from #6073 is kept.
- No function-local `static bool` guards added.
- `_initialBuild` clears unconditionally, so filter rebuilds never yank scroll position.
- `LoadTree` errors out on zero vendors **before** building the tree.
- `wxTE_PROCESS_ENTER` now has a matching `EVT_TEXT_ENTER` handler.
- Focus is only restored to the filter box if it had focus when the rebuild started.
- Suppress-list config read hoisted to once per rebuild; vendors with zero models stay visible (their contact/website panel remains reachable).
- Detail panels are cleared after a rebuild so they never show an item no longer in the tree.
- Timer is a by-value member, stopped on close.

## Testing

macOS Release (arm64) build, manual: unfiltered tree identical to master (categories, wiring children, vendor/category/model panels); filter narrows live for `efl`, `gilbert`, `tree`, `arch`, `pumpkin boscoyo`; cancel restores; Enter applies instantly; next-match steps through filtered results; multi-select insert still works; suppressed and model-less vendors reachable.

Windows testing welcome — unfiltered builds expand nothing (as master), and filtered expansion touches only the surviving branches.


dkulp:
Just had a chat with a couple of vendors...  Removing the categorical display is a non-starter.      I'm not sure if anything of this can be salvaged while maintaining the original tree structure or not, but that is a hard requirement.

heffneil:
The usability in the current version isn’t great. The categories remain, but the search filters everything out and only shows the models that apply. If the primary purpose of the catalog is to serve as an advertisement or storefront for vendors, then I can understand the current approach. It’s just cumbersome from an end-user perspective.

If the goal is to build a catalog primarily for browsing and purchasing, that’s a neat idea. But from a utility and usability standpoint, I think implementing this functionality would be a significant improvement.

heffneil:
@dkulp reworked as requested — the Vendor → Category → Model hierarchy is fully retained (pushed in 83f129dab, title/description updated). With no filter typed the tree is identical to master: categories, sub-categories, multi-wiring model nodes with wiring children, empty categories pruned. The live filter now narrows *within* that structure — matching models stay under their categories, empty branches drop away, and only the surviving branches auto-expand (bounded, so the Win32 whole-catalog-expansion hazard never applies; unfiltered builds expand nothing). Two small browsing improvements ride along: models sort A–Z within each category, and models with no resolvable category — previously unreachable — appear directly under their vendor. Everything from your #6256 review remains addressed.

PR #6736, opened 2026-07-23
Fix Servo 3D preview: pivot offsets, mesh linkage, and Zero Behavior accuracy

pgianotto:
## Summary
- Fix Servo 3D pivot offsets: dynamically-added servos lost Z axis and used wrong scale
- Fix Servo 3D mesh linkage to resolve remapped Servo Linkage, not mesh index
- Fix mesh-linkage composition order regression and wire in the merged GetChannelValue
- Make servo Zero Behavior preview accurate using pixel alpha, not value==0 (with an interim layout/sequencer-preview approach tried and reverted in favor of this)
- Extend the fix to DmxSkull's servo preview and dedupe channel reads

Rebased cleanly onto current master (2026.13) with no conflicts; confirmed a full Release/x64 build succeeds with these changes applied.

## Test plan
- [ ] Verify Servo 3D models with dynamically-added servos preview correct pivot/Z-axis behavior
- [ ] Verify Servo 3D mesh linkage follows remapped Servo Linkage rather than raw mesh index
- [ ] Verify Zero Behavior preview matches DMX channel value (via pixel alpha) for both DmxServo3D and DmxSkull models
- [ ] Regression-check existing Servo/DMX effects still render correctly in layout and sequencer previews

### DIFF
--- a/plans/ipad-parity/06-layout-models-preview.md
+++ b/plans/ipad-parity/06-layout-models-preview.md
@@ -155,7 +155,7 @@
 | Mesh object file picker (.obj/.3ds/.stl/.ply) | dialog | ✅ | ✅ | parity | P2 | medium | feasible | iPad mesh-file picker w/ UTType filter. |
 | Terrain object height-map / editing | dialog | ✅ | 🟡 | ipad-missing | P3 | hard | feasible | iPad terrain edit target flow; full height-map paint scope narrower than desktop TerrainObjectPropertyAdapter. |
 | DMX MovingHead / MovingHeadAdv create + props | menu/panel | ✅ | ✅ | parity | P2 | hard | feasible | iPad AddModelSheet + descriptor (fixture/mode), MovingHeadFixtureRowView. |
-| DMX Servo / Servo3D create + props | menu/panel | ✅ | ✅ | parity | P2 | hard | feasible | iPad add types + descriptor. |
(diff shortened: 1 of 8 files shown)


dkulp:
Thanks for digging into this....

I'm taking two of the three changes:

- is_2d on dynamically-added servos — clear bug, nice catch. SetNumServos/InitModel disagreed with CreateServo, so servos added by raising the servo count lost Pivot Offset Z and stored offsets at 1/100 scale.
- Mesh linkage via servo_links — agreed this is more correct, and it's a no-op when servo linkage is left at its default, so the blast radius is small.

I'm dropping the Zero Behavior / pixel-alpha commits, though. The alpha signal isn't available at that point in the pipeline: GetChannelValue reads Nodes[], and nodes only store uint8_t c[3] — no alpha. DMX models force StringType = "Single Color White", and NodeClassWhite::GetColor calls the 3-arg xlColor::Set, which hardcodes alpha = 255. So driven is always true, the Min/Max/Center branches never execute, and the only live effect is removing the channel_value == 0 → lastValue hold — which makes servos snap to min_limit whenever an effect ends.

The underlying idea is still worth having, just keyed off channel_value == 0 (which is what the physical PWM controller reacts to) with "Hold" still mapping to lastValue. That keeps today's behavior for everyone who hasn't explicitly picked Min/Max/Center. Happy to look at a follow-up PR for that if you're up for it.   

I'm going to close this PR as the usable parts have been merged.  Subsequent updates/changes can be in a new PR. Thanks!

PR #6746, opened 2026-07-24
Making square pixel appearance face the camera correctly in 3d

kylegrymonprez:
I happened to see this issue on my own layout in 3d - I had a 1 node single strand model that when in 3d, would rotate with the 3d model. 

I had the intern address the issue with the square appearance much, similar to circles. Now in 3d, it respects the rotation of the camera instead of continuing to rotate to face the camera the whole time.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -11,6 +11,11 @@ Issue Tracker is found here: www.github.com/xLightsSequencer/xLights/issues
 XLIGHTS/NUTCRACKER RELEASE NOTES
 ---------------------------------
 2026.15  August ??, 2026
+    -bug (dkulp)                 Shader: a shader that accumulates into gl_FragColor without first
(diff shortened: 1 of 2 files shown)


dkulp:
I'm honestly not sure this is what we want....   The entire point of "Square" is it's the simplest and fastest option for the graphics cards to render.   We sent points and tell the graphics card to just render them as points.   There isn't any complex math in and of the shaders, no blending needed, etc....    This PR turns a single point into 6 point (6x the data) and involves way more math and operations during the shaders.

kylegrymonprez:
i get that - so you're saying if i want it to look 'good' in 3d, i need to change it to a circle and the preference is to keep the square a non-3d object? Thats fine, in which case I'd recommend we say that somewhere that it's not intended for 3d space - maybe even calling is a 2d square.

PR #6751, opened 2026-07-25
Filterable model pickers for the Export-to-Other-Models dialogs

heffneil:
## Summary
Routes the four "Export → Other Models" model pickers through the existing filterable `CheckboxSelectDialog` instead of the unfiltered `wxMultiChoiceDialog`, so their model lists can be searched/filtered:

- **Export Faces/States/SubModels** — layout model right-click (`LayoutPanel.cpp`)
- **Export Face Definitions** — Faces panel (`ModelFacesPanel.cpp`)
- **Export States** — States panel (`ModelStatesPanel.cpp`)
- **Export SubModels** — SubModels panel (`SubModelsPanel.cpp`)

`CheckboxSelectDialog` already provides the search/filter box (e.g. the EffectsGrid "Choose Model(s)" copy-layers dialog uses it). These four export pickers were the remaining ones still on the unfiltered `wxMultiChoiceDialog` — on a show with many props the list was hard to navigate. This is a small, behavior-preserving swap: same selections in, same models exported.

## Notes
- No change to `CheckboxSelectDialog` itself — it is used as-is.
- `GetSelectedItems()` returns the checked model names (guaranteed to be in the passed list), replacing the old index-based `GetSelections()` / `choices.at(idx)`; the dirty / change-count behavior at each site is unchanged.

## Test Plan
- [x] macOS Debug build succeeds.
- [ ] Right-click a model → **Export Faces/States/SubModels** → the picker shows a **Filter** box; typing narrows the list; checked models persist across filtering; **OK** exports to exactly the checked models. Same for the Faces / States / SubModels panel "Export … to Other Models" menu items.

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -50,6 +50,7 @@
 #include "layout/LayoutPanel.h"
 #include "layout/ModelPreview.h"
 #include "xLightsMain.h"
+#include "shared/dialogs/CheckboxSelectDialog.h"
 #include "xLightsApp.h"
 #include "settings/XLightsConfigAdapter.h"
 #include "model/ChannelLayoutDialog.h"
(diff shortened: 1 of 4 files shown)


computergeek1507:
Is the selection dialog list already alphabetical sorted?

heffneil:
we have applied this to similar dialogs.  Alphabetize doesn't work

cybercop23:
Yeah, but he has all his models start with the vendor name, hard to get to the right candy cane or flake or tree

computergeek1507:
Maybe have Claude rename the models to be sensible

heffneil:
Not my layout.  You guys are resistant to my suggestions but you gotta pay attention to the users.

Screenshot 2026-07-27 at 4 35 11 PM

heffneil:
Would be nice to incorporate this one guys as well as the new one from yesterday regarding group filtering. Regardless of props names filtering groups is a useful tool

derwin12:
Please include screen shots...

PR #6756, opened 2026-07-27
Submodel Dimming Curves

### LINKED ISSUE #6704: Dimming Curve on Submodels
Feature request for Dimming Curve on Submodels. 

Would like to be able to set a permanent brightness dimming for submodels to create xmodel dependant shadows on props. This would allow users to have shadows set on the prop for 3D look. 

I currently either use a State Effect set to brightness and dim the Brightness in the color box. This can be done now once during each sequence. I also place all SHADOWS in a group at the bottom of the Master View and apply the permodel default to the group to get the effect. I also use an on effect set to Brightness blending mode. This works as well for a submodel. 

This feature would allow a permanent setting that wouldn't require the user to set these groups up for this specific purpose and to set the effects necessary for the result. Having the dimming curve could allow a fixed solution. Thanks for the consideration.

AGFazio:
### Overview

This is a genuinely slick addition to the modeling toolkit: submodels can now carry their own permanent brightness dimming, independent of whatever effect happens to be painting their channels. Up to now, dimming a section of a prop meant fighting with effect settings or living with a lighting mismatch — now you just set a brightness value on the submodel once, and it stays dimmed, forever, no matter what runs on top of it. This is exactly the kind of quality-of-life feature that makes props look dramatically more realistic with almost zero ongoing effort from the sequencer.

### What it does

- **Set-and-forget dimming.** Right-click a submodel → *Set Dimming Curve* → pick a brightness (−100 to 100). The dim is baked into the node data, so it applies at render regardless of effect, and across model/strand/group views.
- **Good for depth/shadow effects.** Negative values are ideal for faking recessed or shadowed sections of a prop (window mullions, roofline overhangs) for a more 3D look.
- **Overlap-safe.** When submodels share nodes, the strongest dim wins — no conflicts.
- **Two enable switches, non-destructive.** Per-model and per-sequence toggles let you turn dimming off without losing saved brightness values.
- **Preview fidelity.** Layout preview correctly reverses the curve to match render output; fixed a bug where preview selection got stuck on dimmed submodels.
- **List UI.** Submodel list shows brightness inline (e.g. `Window (-50)`) so curves are visible at a glance.
- **Full-stack parity.** Lives in the shared core (`Model`, `ModelManager`, `SequenceElements`/`SequenceFile`), so desktop, iPad, and headless render agree by construction. iPad got matching UI (brightness slider + enable toggles); parity scorecard updated.

Fixes #6704


AGFazio:
Thanks for the review — both points were valid, fixed in 799bb47a9: added an equality guard in `SequenceElements::SetSupportsSubModelDimming` to avoid redundant restamping, and added the missing `parent->ApplySubModelDimmingToNodes()` call after `replaceSubModelsOnModel:` rebuilds submodels on iPad.

computergeek1507:
As per chat, this will probably not happen

AGFazio:
Addressed the second review pass in 4c128003b: dropped the O(n^2) padding scan in `SubModelDisplayText` (it didn't reliably align in a proportional font anyway — just appending ` (-50)` now), moved dimming-curve preview suppression to `OnActivate()`/`OnDeactivate()` so it's scoped to the SubModels tab instead of the whole dialog's lifetime, and fixed `sequenceSupportsSubModelDimming` to return NO (not YES) when no sequence is loaded, matching `sequenceSupportsModelBlending`.

AGFazio:
Addressed the third review pass in e39c47eb4 (after 1cd7cde42): made `ModelManager::_subModelDimmingEnabled` a `std::atomic<bool>` since it's written from the UI/sequence-settings path but read unlocked from `Model::ApplySubModelDimmingToNodes`, and fixed overlap resolution so the strongest effect wins (strongest dim among negatives, strongest brighten among positives, dim wins on mixed sign) instead of always picking the smallest raw value.

AGFazio:
Addressed the fourth review pass in 6c9b90459: ValidateWindow() no longer force-resets the per-model checkbox to true when no submodel currently has dimming (was silently overwriting an explicit disable on save), and the property-grid "(disabled)" label now also checks the sequence-level master switch, not just the per-model one.

AGFazio:
Addressed the fifth review pass in 4c1018f4b:
- Fixed the use-after-free `Model::ApplySubModelDimmingToNodes()` deleted `DimmingCurve` objects on rebuild while render buffer node clones (via `InitRenderBufferNodes`) held raw pointers to them, since the clone's copy ctor propagates `nodeDimmingCurve`. Changed it to `std::shared_ptr<DimmingCurve>` throughout (`NodeBaseClass::nodeDimmingCurve`, `Model::_subModelNodeDimmingCurves`) so a live clone keeps its curve alive regardless of when the owning model re-stamps.
- Wired the per-model "Enable Dimming Curves" checkbox into `SubModelsPanel::HasChanges()" so toggling it alone is no longer invisible to the dialog's dirty-check.

AGFazio:
Fixed the clamp issue in 8b942cc10 (clamp the stored brightness to -100..100 before using it as the dialog's initial value, in case a sequence was hand-edited with an out-of-range value).

On the shared_ptr overhead comment: leaving this as-is rather than reverting to a raw/non-owning pointer. The per-node cost is smaller than it looks - a default-constructed `shared_ptr` (the common case: most nodes never get a dimming curve) is two null pointers with no control-block allocation and no atomic traffic, and for nodes that do get one, the refcount operations only happen at clone-construction (render buffer build) and restamping (an editor-time operation), not on the per-frame GetColors/SetColors hot path, which only dereferences/compares the pointer. Reverting to a non-owning index/pointer would reintroduce the exact use-after-free this same review flagged two passes ago - an index into `_subModelNodeDimmingCurves` is only as safe as the vector never being cleared while a clone still references it, which is precisely what breaks today. Happy to revisit if profiling ever shows this mattering in practice.

dkulp:
Closing this...  As discussed previously, this will NOT be merged and every co-pilot review is costing me actual money.

heffneil:
I don't see any public details why this was rejected.  As a contributor, it would be good to have more insight into why PR's wouldn't fit with the overall vision.

AGFazio:
@heffneil  It was discussed. Just not as comments here.

PR #6758, opened 2026-07-27
Add zoom and fit controls to Node Layout

mpl1337:
## Summary

- add a 10–300% zoom slider and percentage display to the Node Layout dialog
- support Ctrl+mouse-wheel zoom while leaving normal wheel scrolling unchanged
- limit zooming to the node grid so the model information remains readable
- add a Fit action that selects the largest zoom level that shows the complete layout without scrolling
- keep Print and Open in Browser output unchanged

## Motivation

Large custom models can extend beyond the visible Node Layout area. The existing static HTML table requires scrolling and provides no way to see the complete node arrangement at once.

## Implementation

The dialog uses a display-only copy of the generated HTML. Node-table cell contents are assigned their own HTML font-size slot, allowing the grid to be scaled independently from the model metadata. wxWidgets reparses and lays out the table after each font-size change. The wxSmith source is updated alongside the generated dialog declarations and initialization.

The Fit action uses the rendered virtual size and the HTML window's client size to find the largest zoom percentage that fits both dimensions.

## User impact

Users can inspect large node layouts as a whole, zoom in for individual node labels, and switch between zooming and normal scrolling without affecting printed or browser-rendered layouts.

## iPad parity

The iPad `NodeLayoutSheet` already provides zoom and pan for the node-grid visualization, so this closes a desktop usability gap without creating a new iPad gap. The existing theme-06 parity row is updated with the desktop controls.

## Validation

- Windows `Release|x64` build with MSVC v145: passed with 0 errors
- wxSmith XML parsed successfully
- `git diff --check`: passed
- built application launched successfully on Windows


mpl1337:
image

dkulp:
Merged, but also applied some performance fixups and minor cleanups.

PR #6763, opened 2026-07-27
Layout: open Wiring View and Node Layout as modeless windows

heffneil:
## What
Right-click a model in the Layout panel → **Wiring View** or **Node Layout**. Both previously opened **modally**, blocking the rest of xLights until you closed them. This PR makes both **modeless**, so you can keep working in xLights (select models, pan the preview, wire things up) with the reference window open behind you — and open several at once (e.g. Wiring View on two different props side by side).

## How
`ShowWiring()` and `ShowNodeLayout()` (`src-ui-wx/layout/LayoutPanel.cpp`) now heap-allocate the dialog, parent it to the top-level frame (so it stacks like a normal window and can go behind the main window rather than floating over the layout panel), and `Show()` it modeless instead of `ShowModal()`. Each dialog destroys itself on close.

**Lifetime safety:** neither dialog retains a `Model*`.
- `WiringDialog::SetData()` snapshots everything it needs (points, rows/cols, name) up front.
- `ChannelLayoutDialog` only holds the HTML string handed to `SetHtmlSource()`.

So both are safe to outlive the selected model.

**Close handling:** `WiringDialog` has no buttons — the title-bar close fires `wxEVT_CLOSE_WINDOW` → `Destroy()`. `ChannelLayoutDialog` has an OK button; on a modeless dialog wxDialog's default OK only *hides* the window (never firing the close event), so its OK is routed through `Close()` to actually destroy it.

The other, modal `WiringDialog` use inside the Custom Model editor is unchanged.

## Testing
Built (macOS Debug) and exercised manually: both windows open modeless, xLights stays fully usable behind them, multiple windows coexist, and both close cleanly via the title-bar ✕ and (for Node Layout) the OK button.

## iPad parity
Desktop-only Layout-panel model-inspection dialogs; the iPad app has no Wiring View / Node Layout surface. This is a modality change to existing desktop-only dialogs — no iPad counterpart to update.

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -8051,9 +8051,16 @@ void LayoutPanel::ShowWiring()
 {
     Model* md = dynamic_cast<Model*>(selectedBaseObject);
     if (md == nullptr || md->GetDisplayAs() == DisplayAsType::ModelGroup || md->GetDisplayAs() == DisplayAsType::SubModel) return;
-    WiringDialog dlg(this, md->GetName());
-    dlg.SetData(md);
-    dlg.ShowModal();
+    // Show modeless so the user can keep working in xLights with the wiring
+    // window open behind it. WiringDialog::SetData snapshots everything it needs
+    // (points, rows/cols, name) and keeps no Model* pointer, so it is safe to
+    // outlive the model. Parent to the top-level frame so it stacks like a
+    // normal window (can go behind the main window) rather than floating on the
+    // layout panel. Heap-allocated; it destroys itself on close.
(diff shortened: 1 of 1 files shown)


deplanche:
Have you tested what happens if you change something in the model (add nodes, change number of ports, etc), when this is open?  Will it refresh, close, or just stay with the old layout that is now incorrect... or worse yet, will it crash things?

heffneil:
I am sure it will not update.

heffneil:
This leaves wiring in front so you can do other stuff while leaving the wiring view open (and node layout)

Screenshot 2026-08-16 at 10 02 30 AM
Screenshot 2026-08-16 at 10 02 25 AM

PR #6770, opened 2026-07-28
Test dialog (Tools > Test): live name filter on every tab + crash fixes

heffneil:
## What

Adds a **live name filter** to every tab of the **Tools → Test** (Pixel Test) dialog, so a long model/controller list can be narrowed by typing instead of scrolling.

- A search box sits above each of the **Outputs**, **Model Groups**, **Models** and **Controllers** trees.
  - Typing narrows the tree to items whose name — or a descendant's — matches (case-insensitive).
  - A matching item keeps its whole subtree; container branches (controller → port → model, group → model) are kept when a model under them matches, and surviving branches auto-expand so nested matches are visible.
  - Debounced 250 ms.
  - **Selections are preserved** across filtering, because they live in the `ChannelTracker` (channel ranges), not in the tree items — the rebuilt items re-derive their check state.
- The **Model** tab uses a `wxChoice` dropdown, not a tree; its filter narrows the dropdown's entries against the full model list, keeps the current selection when it still matches, and only re-renders the preview when the selection actually changes.

The filters and the Model-tab box are added in non-wxSmith code (reaching the generated sizer via the control), so no `.wxs` / generated-code changes were needed.

## Crash / leak fixes (pre-existing, surfaced while building this)

Exercising the dialog on a Debug build tripped three latent bugs, all fixed here:

1. **Format-string SIGTRAP** — `wxString::Format("%ld", …)` was given `int` args (`CPR_PortTestItem` port label; port-cycle status line). The wx Debug format validator aborts on the `%ld`/`int` mismatch → crash on **opening** the dialog / running a port test. Cast the args to `long`.
2. **Double-free on tree teardown** — `wxTreeListModelNode`'s destructor deletes each node's client data, while `ModelTestItem`/`ModelGroupTestItem` *also* delete their submodel/member children. The old code relied on delete-time "don't delete these" exceptions; rebuilding a filtered tree (delete + repopulate) double-freed. Replaced with an explicit `ReleaseDualOwnership` pass that hands the two double-owned relationships to the tree, so `DeleteAllItems`/`DeleteItem` frees each object exactly once. **Also fixes a latent double-free on dialog close.**
3. **`UDController` leak** — `CPR_ControllerTestItem` now owns and frees the `UDController` built for it (previously leaked one per controller per open; would have multiplied under per-keystroke rebuilds).

## Testing

- Built Debug (macOS) and exercised all four tree tabs + the Model dropdown filter; lists narrow as expected and selections survive.
- Opened/closed the dialog repeatedly (previously a latent double-free on close) — stable.

## iPad parity

Desktop-only (`src-ui-wx/setup/`). The pixel Test dialog has no iPad counterpart, so this is a capture-in-plans item (Tools/13 backlog), not an immediate port.

### DIFF
--- a/src-ui-wx/setup/PixelTestDialog.cpp
+++ b/src-ui-wx/setup/PixelTestDialog.cpp
@@ -17,7 +17,6 @@
 #include <wx/dataview.h>
 #include <wx/confbase.h>
 #include <wx/numdlg.h>
-#include <wx/utils.h>
 
 //(*InternalHeaders(PixelTestDialog)
 #include <wx/intl.h>
@@ -656,6 +655,10 @@ public:
             _subModels.pop_front();
         }
     }
+    // Give up ownership of the submodel items without deleting them. Used when
+    // the tree they live in is torn down: on the Models tab each submodel is a
+    // tree node that the tree will free, so we must not also free it here.
+    void ReleaseSubModels() { _subModels.clear(); }
     ModelTestItem(const std::string& name, const std::string& modelSuffix, ModelManager& modelManager, bool channelsAvailable, int nodes = -1, long startChannel = -1, long endChannel = -1, long nodeOffset = 0) :
         TestItemBase()
     {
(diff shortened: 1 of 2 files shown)


derwin12:
Screen shots?

heffneil:
Screenshot 2026-07-28 at 9 28 37 AM
Screenshot 2026-07-28 at 9 28 26 AM
Screenshot 2026-07-28 at 9 28 17 AM
Screenshot 2026-07-28 at 9 28 11 AM
Screenshot 2026-07-28 at 9 28 05 AM
Screenshot 2026-07-28 at 9 27 59 AM
Screenshot 2026-07-28 at 9 27 45 AM
Order may not be accurate.

PR #6771, opened 2026-07-28
Replace Model(s): add a Model Groups option (keep / replace / merge)

heffneil:
## What

Adds a **Model Groups** option to **Replace Model(s) With This Model**, controlling what happens to each replaced model's group memberships:

| Choice | Behavior |
|---|---|
| **Do not change** *(default)* | The clone takes the target's name, so it keeps the target's own groups. |
| **Replace destination's direct groups with source's** | The replaced model ends up in exactly the source's direct groups. |
| **Merge source's groups into destination's** | The replaced model keeps its groups and gains the source's. |

## Shared core + desktop **and** iPad (parity)

Reconciliation lives in a shared, wx-free helper — `ModelManager::ReconcileReplacedModelGroups` (src-core) — so the desktop dialog and the **iPad** `ReplaceModelSheet` drive identical behavior. The desktop dialog gets a radio; the iPad `replaceModels` bridge gets a `groupMode` arg + a matching picker. Parity plan row (`06-layout-models-preview.md`) updated to reflect the option on both sides.

## Correctness: submodel-aware

The helper works on **direct** membership including submodel entries: `"Source/Strand1"` is remapped prefix-safe to `"Target/Strand1"`, and in Replace mode a group's existing `"Target/*"` entries are cleared first so stale submodel references don't survive when submodels aren't kept. Base-folder groups are never touched; the group list is scanned once per batch.

## Builds

- macOS desktop (Debug): ✅
- `xLights-iPadLib` (arm64/iOS): ✅

## Release note

`README.txt` (2026.15) will get, on merge:
```
    -enh (heffneil)              Replace Model: new option to keep, replace, or merge each replaced
                                 model's group memberships with the source model's (desktop + iPad)
```


heffneil:
image

dkulp:
Issues

1. The iPad-parity claim in the PR body is factually wrong (blocking per AGENTS.md §3)

▎ "The bulk Replace-Model dialog has no iPad counterpart yet"

It does. plans/ipad-parity/06-layout-models-preview.md:111 marks this feature ✅ parity, backed by ReplaceModelSheet (src-iPad/App/LayoutEditorView.swift:13663) and -replaceModels:withSource:keepStartChannel:keepSubmodels:keepSizePosition:forDocument: (src-iPad/Metal/XLMetalBridge.mm:1570) — the same three options this dialog has.

So this is State A or an explicit State B, not "no counterpart": either add the group mode to the sheet + bridge, or flip row 111 to 🟡 ipad-weaker with file:line evidence for both sides and bump 00-overview.md counts. Right now the PR changes 2 files and touches neither plan.

2. Missing README.txt entry (AGENTS.md §10)

A user-visible desktop enhancement needs a line in the 2026.15 section. Precedent exists for this exact feature — README.txt:640 (-enh (heffneil) Replace Model: new dialog…) and :537. Suggested:

    -enh (heffneil)              Replace Model: new option to keep, replace, or merge each replaced
                                 model's group memberships with the source model's

3. Submodel group entries are ignored — inconsistent with the shipped paste flow

DirectlyContainsModel(name) is an exact match on modelNames, so a group whose membership is "Source/Strand1" (not "Source") yields srcIn == false and the target is never added; symmetrically, a group containing only "Target/Strand1" is never cleaned up in Replace mode. The existing "add model to the same group(s) as the original" paste path deliberately handles this with StartsWith(groupModels[i], source_model_name) + Replace(...) (LayoutPanel.cpp:10470-10475).

This matters more than it looks because when Keep each target's submodels is off, the clone carries the source's submodels — so surviving "Target/OldSub" group entries now reference submodels that no longer exist. Worth at least mirroring the paste flow's StartsWith mapping for the add side.

4. "Replace" only means direct membership — the label overpromises

"Replace destination's groups with source's groups" reads as exactly, but a target reachable through a subgroup stays reachable, and base-folder groups are (correctly) untouched. Consider "Replace destination's direct groups…" or a tooltip noting base-folder and nested groups are left alone.

Nits

- srcIn is loop-invariant — hoist it. Compute the source's group set once before the replacedNames loop instead of re-scanning every group's modelNames per target. With N targets × G groups this is N×G linear scans plus one recursive ResetModels() per AddModel (ModelGroup.cpp:763 recurses into nested groups); fine for typical batches, wasteful for a 200-target × 150-group layout.
- Dead guard: if (name == srcName) continue; can't fire — candidates already exclude sourceModel by pointer. Harmless as defense, but the comment implies it's load-bearing.
- Redundant type test: GetDisplayAs() == DisplayAsType::ModelGroup followed by dynamic_cast — pick one (the codebase generally does the enum check + static_cast, e.g. ModelManager.cpp:2151).
- _() inconsistency: the new strings are the only translated ones in ReplaceModelDialog.h — every other string in the file is raw. Either drop _() for consistency or add #include <wx/intl.h> so the macro isn't relying on a transitive include that differs per platform.
- Hardcoded wxSize(460, 640): the radiobox is now the widest control in a fixed, non-DIP-scaled dialog. At 150–200% Windows DPI the labels can clip. SetSizerAndFit(top) (or top->SetSizeHints(this)) would size it from content and is a one-liner.
- Style: GroupHandling() is read after the loop while the other three options are read up front — pulling it up with copyStartCh/mergeSubs/copySizePos reads better and makes the dialog-lifetime dependency obvious.

heffneil:
Thanks for the thorough review, Dan — all addressed in the force-push:

1. **iPad parity (blocking)** — you're right, I had it backwards. Rather than downgrade the row, I ported the option to iPad (State A): the group-mode reconciliation now lives in a shared wx-free helper `ModelManager::ReconcileReplacedModelGroups`, called by both the desktop dialog and the iPad `ReplaceModelSheet` (new `groupMode` arg on the `replaceModels:` bridge + a picker in the sheet). Plan row 111 updated to reflect the option on both sides; still ✅ parity. `xLights-iPadLib` builds.

2. **README.txt** — release-note line prepared (your suggested wording) and will be added to the 2026.15 section on merge.

3. **Submodel entries** — fixed. The shared helper matches direct entries *and* `Source/*` submodel entries, remapping prefix-safe (`Source/Strand1` → `Target/Strand1`). In Replace mode it clears the group's existing `Target/*` entries first, so the stale-submodel case you flagged (clone carries the source's submodels when "keep submodels" is off) no longer leaves dangling `Target/OldSub` references.

4. **Label overpromises** — now "Replace destination's **direct** groups with source's groups", plus a tooltip noting nested (subgroup) and base-folder groups are left alone.

**Nits:** source-owned set is now computed once per group for the whole batch (single scan); dead `name == srcName` guard removed; `dynamic_cast` → enum check + `static_cast`; explicit `#include <wx/intl.h>`; `SetSizerAndFit` (with a DIP-scaled min list size) instead of the hardcoded `wxSize`; and `GroupHandling()` is now read up front with the other options.

Issue #6762, opened 2026-07-27
Camera View Points


tgrinnall:
when you click on view then perspective it would be nice to have a import and maybe even export option to be able to import/export different views from/to other show folders. also when import/export views you should be able to select what views you want to import/export. as for exporting views you should be able to choose what show folders you would like to export each view too.

derwin12:
Just to be clear here Tim .. got a screenshot of what you are referring to?

computergeek1507:
I think you already do this with the right click menu in the layout tab to Import Views/Models/Groups

tgrinnall:
per scott
<img width="318" height="552" alt="Image" src="https://github.com/user-attachments/assets/a8d01b80-c487-4dd2-8705-7fc12685c218" />

tgrinnall:
a button that you could click under auto save to import a camera view from another show folders rgb effect file. also a button to export to another rgbeffect file that you could select the files from a drop down menu. Alex and I had talked and he thinks where scott is talking that it could be added there instead. that is where you can right click somewhere on the screen and go to import previews models and groups and add views to that list. 

<img width="545" height="380" alt="Image" src="https://github.com/user-attachments/assets/7335ee2d-25fd-4121-852f-8fbef3374d1f" />

cybercop23:
@tgrinnall please update title to ViewPoints.

Also, Tools -> Perspectives should change to ViewPoints

computergeek1507:
3d viewpoints and perspectives are not the same,

tgrinnall:
Scott its listed in the perspectives list under view. so maybe perspectives should be changed to viewpoints

cybercop23:
> 3d viewpoints and perspectives are not the same,

But there are two perspectives, that's the point. Tools menu refers to 3dviewpoints  and Persectives inside Windows is screen layout, no?

computergeek1507:
Persectives should only be the window screen layout

tgrinnall:
there isn't one under the tools menu. they are both under the view menu. witch has perspectives and windows as the options to choose from. the perspectives one gives you all the view points. the windows one gives you another perspective click option with gives you the chooses for the layout.

tgrinnall:
we got things confused and got it all worked out. Alex is going to put in the enhancement by adding views to the import previews groups models to list all the views from the selected show folder file.

AGFazio:
In next release

PR #6772, opened 2026-07-28
Add Viewpoints imports #6762

cybercop23:
Add the ability to also import Viewpoints #6762 

image


dkulp:
Issues

1. No duplicate-name check on import (main issue)

ImportCameraFromNode() (ViewpointMgr.cpp:194) unconditionally push_backs. Every other path into these vectors guards first — ModelPreview.cpp:124 and iPad XLMetalBridge.mm:652 both require IsNameUnique(name, is3d) and reject with "ViewPoint name is a duplicate or blank".

Consequences:
- Running the import twice (or importing a show that shares viewpoint names) silently doubles the list, with no way to tell the entries apart in the right-click Viewpoints menu.
- Name-based lookups take the first match, so the imported duplicate is unreachable: GetNamedCamera3D() (used for the per-effect camera setting via BufferPanel) and ModelPreview.cpp:172-182 (BUTTON_V1/FRONT/… viewpoint buttons) both break on first hit.

Suggested shape:

bool ViewpointMgr::ImportCameraFromNode(pugi::xml_node node)
{
    auto new_camera = CreateCameraFromNode(node);
    if (!IsNameUnique(new_camera->name, new_camera->is_3d)) return false;
    ...
}

…and have ImportModelsFromRGBEffects() count the rejects and report them (or offer rename-on-collision, mirroring how model import handles existing names). A "Deselect Viewpoints Already In Layout" option — the analogue of the existing Deselect Models Already In Layout — would fit naturally too.

2. The Viewpoints row is identified by its label text

GetPreviews(), GetModelsInPreview(), and GetViewpoints() all key off GetItemText(it) == "Viewpoints". A source show with a layout group literally named "Viewpoints" produces two identically-labelled root rows and three misbehaviours:

- GetPreviews() skips it → the preview is never created in the destination.
- GetModelsInPreview("Viewpoints") returns empty at the new early-out (ImportPreviewsModelsDialog.cpp:170) → that preview's checked models are silently dropped.
- GetViewpoints() walks both rows, so a checked model under that layout group is passed to ImportCameraFromNode(<model> node) → CreateCameraFromNode happily builds an all-zero 2D camera named after the model. Silent junk in the viewpoint list.

Also note the label is user-visible and untranslated ("Viewpoints" rather than _("Viewpoints")), so the comparisons would break under localization anyway.

Fix: stash the row as a member (wxTreeListItem _viewpointsRoot) set in PopulateTree() and compare items, or attach an impTreeItemData of kind Viewpoint to the root row and test the data pointer. Same applies to the _checkedPreviews set, which is also keyed by label string.

3. Display names skip UnXmlSafe

AddViewpoints() reads c.attribute("name").as_string() raw, but CreateCameraFromNode() applies UnXmlSafe(). Camera names are written with XmlSafe() and re-escaped by pugixml on save, so a viewpoint named Front & Center renders in the tree as Front &amp; Center while the imported camera is correctly Front & Center. One-line fix: wrap the label/name in UnXmlSafe(...).

4. Latent: !IsModelGroup() still means "is a model"

SelectAllModel() was correctly tightened to GetKind() == ImpItemKind::Model, but LayoutPanel.cpp:11388 and :11408 still branch on IsModelGroup() / !IsModelGroup(). Viewpoint items can't currently reach ImportModelsFromPreview() — but only because of the label-string filters in issue 2. Worth converting those two sites to explicit GetKind() checks so the guarantee doesn't rest on that.

---
Nits

- PopulateTree()'s hasCamera pre-scan duplicates the scan inside AddViewpoints(). Since childless rows are already deleted under a filter, you could append unconditionally and extend that cleanup to run always.
- GetViewpoints() doesn't break after finding the row.
- Tree column header is still "Previews & Models" and the filter placeholder still _("Filter models"), though both now cover viewpoints. The dialog title and menu item were updated; these two weren't.
- The previews vector now also holds the Viewpoints row — the name reads a bit off (the comment was updated, the variable wasn't).
- Checking the bare Viewpoints parent row with no children enables OK but imports nothing. Matches existing preview-row behavior, so arguably fine.

cybercop23:
Handle duplicate names with the usual -#  and address the above items.

image

PR #6781, opened 2026-07-28
3d - clicking on the orange center handle will not open states dialog

kylegrymonprez:
Problem: Clicking on the center orange handle would sometimes bring up the states/faces/submodel dialog. After @cybercop23 fixed deselecting in [1fd791](https://github.com/xLightsSequencer/xLights/commit/1fd791d5f5c005d8538b7378b7c813857e261393), it was still quirky. This is my (interns) attempt to address the UX issue.

Now, no matter how many times you click on the orange cube to change the handles, it will change the handles. Double clicking anywhere else on the model will bring up the dialog, and clicking off the model will deselect.


derwin12:
Can we just remove the double click on the layout opening up model properties?? submodels/faces/states are not that important to be opened from the layout - use the properties section.

derwin12:
We have so much AI slop now to avoid misclicking something that could just easily be removed.

kylegrymonprez:
@derwin12 Thats not a me decision - someone else decided that was the double click functionality. Open to discussion but that was the design pattern someone added in a recent version.

PR #6794, opened 2026-07-29
Add FPPConnect to Play toolbar

cybercop23:


computergeek1507:
Why, it's already other places

cybercop23:
> Why, it's already other places

no need to move from the sequencer tab to layout just for that or menu. single click

computergeek1507:
You should probably not FPP connect from the sequencer, I would encourage users to close sequences before uploading them

cybercop23:
We never do that. All they have to do is render and save. We never told them to close a sequence before uploading.

PR #6802, opened 2026-07-30
Autoset StartChannel for the "shadowee" when Shadow For is set

cybercop23:
When users have to define a shadow model, they must also set the "sequencing/shadowee" model correctly. This will automatically do that for them once the shadow model has its Shadow Model For: set.
It also hides the "shadowee" from the available models to be placed on the controller so it can't accidentally be placed on a port and break the start channel link. The warning message still remains, indicating that a model is shadowed by the model placed on the controller.


derwin12:
How do we handle the shadows for MH-Dimmer etc? I can never remember where to put shadow model for and dont always populate it. .. but will this work or will it override the channel number back to 1?

cybercop23:
We can't set them for MH, so we don't. Long discussion last night on Zoom on how to support this since there are multiple that reference the same model. In the case of MH, we do the opposite. We put the MH on the ctrl, but have these shadows for sequencing not be placed on the ctrl. So the shadow would have to be placed on the MHs themselves

derwin12:
i have them on mh so that check sequence doesnt yell at me.

derwin12:
just trying to think of the other edge cases .. like shadowing for weird trees and controllers. I guess if they dont set the shadow model for then it wont impact them. .. what about those users that have that bug where it populated that field throughout their display. that could be fun.

cybercop23:
It will set it for the initial config, and you can adjust it as needed if you have a tree/prop split among controllers (long discussion on a way to address that in the future too).
Yes, we also thught of the bug of phantom shadow setting, but a) we haven't see that in recent version, so maybe .4 "fixed it" and b) we have the bulk edit to address that.

cybercop23:
Added a check to not set it if already set by setting up the MH first using Andy's tool.
MH1-4 are shadows for the string models, and they don't show up in Visualizer so the user can't place them on a port and break the strt channel link

image

PR #6818, opened 2026-08-02
Allow up to 4 "pages" of effects for EffectsWheel

cybercop23:
Apparently, some people do use Effects in the Key Bindings. Allow up to 4 pages of 18 effects. Only shows "pages" for whatever many effects there are.
image


computergeek1507:
Gives real confidence that it is well tested when it doesn't compile...

cybercop23:
Be nice. I removed a comment and inadvertently removed the header declaration.

PR #6672, opened 2026-07-08
Modernize Preferences dialog + embed Key Bindings (treebook nav, descriptions, modeless)

heffneil:
## Modernize the Preferences dialog + fold in Key Bindings

Overhauls the desktop Preferences UI and moves keyboard-shortcut editing into it as a single page.

### Preferences
- **Left navigation** rebuilt as a macOS-Settings-style treebook with colourful per-page icons; pages are sorted **alphabetically**.
- **Modeless** — the dialog can stay open while you keep working in Layout/Sequencer (Show() instead of ShowModal(); OK applies + runs the post-change work, Cancel/close dismiss; reuse an open instance by window name).
- **Enforced minimum size** so pages (e.g. the Random Effects lists) can't be narrowed into clipping — `wxTreebook` doesn't propagate page minimums.
- **Descriptions** — every setting on **Other**, **Backup** and **Video** now has a plain-language description line; descriptions hide alongside any per-platform-hidden control so no orphaned text is left behind.
- **Consistent bold section headers** (plain text, not `wxStaticBox` captions, which macOS renders in a smaller font).
- **Video page** — codec / bitrate / hardware-decode split out of "Other" into their own page.
- **AI page** font matches the rest of the dialog (it renders in a `wxPropertyGrid`).
- **Random Effects** — two-list shuttle that no longer clips effect names (rebuilds list contents after layout so macOS re-measures the column).

### Key Bindings
- Now a **page inside Preferences**; the **File-menu "Key Bindings" item is removed**.
- Friendly action names, a single **Shortcut** column, per-binding descriptions, a live filter, a **Category** filter (**Effects / Wheel of Effects**, Presets, Apply Settings, Commands), **alphabetical** Action column, and editing via a popup editor.
- A note explains that key-bound effects appear on the Wheel of Effects when double-clicking the sequencer grid.

### Sequencer
- New bindable action **Copy Layers/SubModels to Models** (previously only a right-click row-heading command), dispatched to a new `EffectsGrid::CopyModelLayersToModelsForSelection()`.

### iPad parity
Desktop-only Preferences/UI work (`src-ui-wx/`). The iPad app has its own settings surfaces and no equivalent Preferences dialog; the shared `src-core/` behavior is unchanged. No iPad code changes are required — captured as a note here per the parity policy.

### Notes
- Branch was cut on June 28 and is behind current `master`; may need a rebase before merge.
- Release-note text is staged locally for manual paste into `README.txt` (not modified here).

[...]

PR #6836, opened 2026-08-03
Key Bindings: modeless editor with inline editing and Wheel-of-Effects positions

heffneil:
## What

Reworks the **Key Bindings** editor in place (**File → Key bindings**), keeping it where users already expect it.

This is a focused revival of the keybindings half of the closed #6672 — **without** the Preferences-treebook redesign that PR was closed over (too many conflicts / needed more design discussion). No Preferences pages are touched here.

### Editor
- **Modeless** — stays open while you keep working in Layout/Sequencer; reopening raises the existing window instead of stacking duplicates.
- **Friendly action names** + a plain-language **description** for every binding.
- **Category / Scope / text filters.** The Scope dropdown is built from the data, so it only offers scopes that actually occur in the selected category (no more picking a combination that yields an empty list).
- **Inline editing, no popup** — the key is captured by pressing it, with **Control / Alt / Shift / Raw Control** checkboxes and a compact effect / preset / setting control.
- **Save / Cancel** — edits are staged on a working copy of the `KeyBindingMap` and only written to the live map (and saved) on **Save**, so a mistake can be backed out.

### Effects / Wheel of Effects
- A **Position** column showing each effect's wheel **clock slot**, plus a **live wheel preview**.
- Reorder by **drag** (with a native drop indicator) or **Move Up / Move Down**.
- Both mirror `EffectsGrid`'s actual rule — the first **18** *enabled*, *Sequencer-scoped* effect bindings — so an effect that is keyless (**"needs a key"**, since `SetKey("")` disables a binding) or past the cut-off (**"not on wheel"**) is called out rather than implying a slot it will never occupy.

## Demo

See the demo video in the comments below.

## Notes

- The list uses `wxDataViewListCtrl` for native drag-reorder and fast scrolling. A trailing spacer column absorbs the control's last-column stretch, so the key columns keep their widths.
- New sources are registered in `xLights.cbp` and `Xlights.vcxproj`(`.filters`); `CMakeLists.txt` already globs `src-ui-wx/preferences` and `src-ui-wx/app-shell`.
- Mac-only modifier glyphs (⌘/⌥/⌃/⇧) stay behind the existing `#ifdef __WXOSX__`; Windows/Linux render `Ctrl+/Alt+/Shift+`.

## Testing

- macOS Debug build clean; exercised filters, inline key capture, modifier checkboxes, Save vs Cancel, add/remove, drag + Move Up/Down reordering, and the wheel/Position behaviour around the 18-slot limit.

## iPad parity

Desktop-only (`src-ui-wx/`). Keyboard shortcuts have no iPad counterpart (no hardware-keyboard binding editor there), so this is a capture-in-plans item rather than a port.


heffneil:
Demo video below. I had to lower the resolution to get it under 10mb.  Let me know if anything is unclear.

https://github.com/user-attachments/assets/a9abf49d-4a41-4b9e-84d7-c527b01394e5

dkulp:
Personally, I want to completely decouple the "Wheel" from the keybindings.    The wheel should have its own customizer and it should not be dependent on anything key related.

cybercop23:
Should we port the Effects from KeyBindings and use them in this "fashion"? Otherwise, we'll also need Effect settings for the wheel since "some" do actually use that, which is why I expanded the use to 4 "pages".

heffneil:
ok thats good feedback - not sure how that works I guess Wheel and then drag and drop or something or sort position?

heffneil:
Closing this rather than letting it sit — the feedback here is clear and this PR is the wrong shape for it.

@dkulp's point stands: the Wheel should have its own customizer and shouldn't be coupled to key bindings at all. And @cybercop23 is right that a proper Wheel editor needs effect *settings*, not just effect names, which is more than this PR models.

Bundling the two is what made this hard to review — the modeless key-binding editor and the Wheel-of-Effects work are independent features sharing a dialog only for historical reasons. Rather than argue the bundle, I'll split it:

- the modeless editor, inline key editing and enriched binding descriptions can come back as a focused PR that touches only key bindings
- the Wheel deserves its own customizer, decoupled as suggested, with settings support designed in from the start

The branch is preserved, so nothing is lost. Thanks both for the review — it saved this from becoming a much bigger mess later.

PR #6866, opened 2026-08-06
Auto-close toolbar gaps and reset toolbars without a restart

cybercop23:
When moving toolbars around or rearranging them, some of them could go off the screen and needed to restart xLights to get it back. No need to restart and now the gap between toolbars auto-closes once the toolbar is placed.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -13,6 +13,8 @@ XLIGHTS/NUTCRACKER RELEASE NOTES
 2026.16  August ??, 2026
     -change (cybercop23)         Add warning prompt when opening a sequence outside the show directory. Save requires
                                  saving inside the show directory - Save As was already ensuring that path.
+    -enh (cybercop23)            Toolbars automatically close gaps between docked items when moved and update
+                                 visibility live when switching tabs.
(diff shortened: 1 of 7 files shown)


dkulp:
What if people actually WANT the gaps for visibility sake?  Or have a toolbar somewhat right aligned on their screen?

cybercop23:
Sure, that's a possibility, but honestly, I've never seen anybody on Zoom with "custom" spacing.
I can add a selection to the Toolbar menu, or a preference to "close the gap", or wait and see if anybody actually cares about the spacing. Or put a FB poll get feedback. Which of the options do you think makes the most sense?

dkulp:
Personally, I think they should be treated just like dialogs... Leave them where the user stuck them with the one exception being if it ends up off screen, then pull them in.

That said, if they could support left/right  and top/bottom anchor points, that would be awesome, but that's likely a wxAUI set of  changes.    Basically, have the ability to anchor some toolbar to the left most (like today) others get anchored to the right.  Same for top/bottom so the effect toolbar could be anchored to the bottom right immediate under the sequencer grid (or top right for immediately above it).

cybercop23:
Already can. Is this what you mean?
image

cybercop23:
image

PR #6867, opened 2026-08-06
Effect symbols: fix render-thread race and multi-second edit freezes

o0charlie0o:
Follow-up to #6859, covering three further problems with effects linked to a reusable `EffectSymbol`. All three are on master today.

## 1. Data race corrupting `EffectSymbolManager` (crash)

`EffectSymbolManager` has **no locking at all**, but `_linkedEffects` is mutated from two threads:

- **Render workers** — `RenderJob::FinishRender` → `Element::CleanupAfterRender` → `~Effect` → `UnregisterLinkedEffect`
- **Main thread** — paste → `LinkToSymbol` → `RegisterLinkedEffect`

Two threads restructuring one `std::multimap` is undefined behaviour. The red-black tree's invariants break and it aborts inside `__tree_remove`:

```
#3  std::__tree_remove<std::__tree_node_base<void*>*>
#4  std::__tree<...Effect*...>::__remove_node_pointer
#6  std::multimap<std::string, Effect*>::erase
#7  EffectSymbolManager::UnregisterLinkedEffect   EffectSymbolManager.cpp:207
#8  Effect::~Effect                               Effect.cpp:262
#11 EffectLayer::CleanupAfterRender               EffectLayer.cpp:71
#14 RenderJob::FinishRender                       RenderEngine.cpp:3145
#17 JobPoolWorker::ProcessJob                     JobPool.cpp:394
```

Reproduced by copy/pasting linked effects while a render was draining.

Every method now takes a `recursive_mutex`. Recursive is required rather than incidental — `RegisterLinkedEffect` calls both `SymbolExists` and `UnregisterLinkedEffect`, and `DeleteSymbol` calls `NotifySymbolChanged`; a plain mutex would self-deadlock immediately.

## 2. Render temporaries claiming symbol ownership

`RenderEngine.cpp` creates throwaway effects via `new Effect(*ef)`. The copy constructor copied `_linkedSymbolId` even though the copy is never passed to `RegisterLinkedEffect` — so `~Effect` mutated the shared multimap from a worker thread for an entry it never owned, and would unregister a **live** effect whenever the allocator reused that address. It also let `IncrementChangeCount` fan symbol changes out off the main thread.

Render reads `mSettings`/`mPaletteMap` and never the id, and every other reader (`SequenceFile`, `TimeLine`, `EffectsGrid`) is main-thread, so the copy no longer carries it.

## 3. Multi-second UI freezes when editing a linked effect

**Moving or resizing.** `SetStartTimeMS`/`SetEndTimeMS` call `IncrementChangeCount`, which fanned out to every linked sibling. A symbol captures an effect's *settings*, not its timeline position, so this work was not merely slow but wrong. A single drag ran it twice (start and end), doing a full settings + palette re-parse per sibling and queueing a render job each. Both setters now hold `ScopedSymbolPropagationSuppressor`. The layer's own dirty-marking is unaffected, so the moved effect still re-renders.

**Settings changes.** These legitimately need to propagate, but the render fan-out queued one job per *effect* rather than per *model*. Measured on a real sequence:

| symbol | linked effects | models |
|---|---:|---:|
| `sym_2` | 707 | 5 |
| `sym_4` | 99 | 2 |
| `sym_1` | 97 | 1 |
| `sym_3` | 91 | 1 |
| `sym_5` | 13 | 1 |

986 render jobs across 11 model-slots, each paying a full `RenderJob` + `PixelBuffer::InitBuffer` setup — and that setup dominates the profile. Coalescing per model into one widened range takes it to **986 → 11**, which is what `RenderDirtyModels` already does for ordinary edits.

I measured the obvious alternative first: merging adjacent time ranges gives **986 → 986**, no reduction at all, because the effects don't overlap in time. Per-model coalescing was the option that actually pays.

## Verification

Builds clean on macOS for both `x86_64` and `arm64`. Headless render of the reproducing sequence completes and is byte-identical run to run:

```
IDENTICAL: 10560 frames x 44448 channels match exactly
```

The two freeze fixes were confirmed by hand in the GUI — dragging and render-style changes on a 700-link symbol are both responsive now.

One note for reviewers: the race fix is verified by construction rather than by a reproducing test — a race needing a specific interleaving won't reproduce on demand. What is certain is that the mechanism in the trace above is now impossible: all mutation of `_linkedEffects` is serialised under one lock, and render temporaries no longer touch the map at all.

### DIFF
--- a/src-core/render/Effect.cpp
+++ b/src-core/render/Effect.cpp
@@ -29,6 +29,8 @@
 #include <filesystem>
 #include <chrono>
 #include <unordered_map>
+#include <map>
+#include <algorithm>
 #include <regex>
 
 #include <log.h>
@@ -193,7 +195,15 @@ Effect::Effect(const Effect& ef)
     mPaletteMap = ef.mPaletteMap;
     mColors = ef.mColors;
     mCC = ef.mCC;
-    _linkedSymbolId = ef._linkedSymbolId;
+
+    // Deliberately NOT copying _linkedSymbolId. This copy is a short-lived
+    // render temporary (RenderEngine's `new Effect(*ef)`) that is never passed
+    // to RegisterLinkedEffect, so it owns no entry in the symbol manager. If it
+    // carried the id, ~Effect would call UnregisterLinkedEffect from a render
+    // worker thread — mutating the shared multimap for an entry it never owned,
+    // and matching a live effect outright whenever the allocator reused the
+    // address. It would also let IncrementChangeCount fan symbol changes out
+    // off the main thread. Render reads mSettings/mPaletteMap, never the id.
 }
 
 Effect::Effect(EffectManager* effectManager, EffectLayer* parent,int id, const std::string & name, const std::string &settings, const std::string &palette, int startTimeMS, int endTimeMS, int Selected, bool Protected, bool importing) :
@@ -370,10 +380,18 @@ std::string Effect::GetDescription() const
         return GetSetting("X_Effect_Description");
 }
 
+// A symbol captures an effect's *settings*, not where it sits on the timeline,
+// so moving or resizing an effect must not fan out to its linked siblings.
+// Without the suppressor a single drag ran CopyFromEffect plus a full
+// settings+palette re-parse on every sibling and posted one render event each,
+// twice over (start and end) — seconds of frozen UI per drag on a symbol with
+// a few hundred links. The layer's own dirty-marking still happens, so this
+// effect re-renders normally.
 void Effect::SetStartTimeMS(int startTimeMS)
 {
     assert(!IsLocked());
 
+    ScopedSymbolPropagationSuppressor noFanOut;
     if (startTimeMS > mStartTime) {
         IncrementChangeCount();
         mStartTime = startTimeMS;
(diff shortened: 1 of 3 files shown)


derwin12:
Are you stress testing or is this a real scenario?

o0charlie0o:
@derwin12 Real scenario — this came out of a show sequence I was actually editing, not a synthetic stress test.

It's a 4:24 song with 2320 effects. The symbols are drum hits — `Hat`, `Clap`, `Kick Yellow`, `Kick Green`, `Horn` — each one linked so I can tune the look in a single place. `Hat` alone is 707 linked effects across 5 models (four matrices plus one more), simply because a hi-hat pattern over four minutes is that many hits. So the link counts come from using symbols the way they're meant to be used, not from pushing them.

How each problem turned up in ordinary editing:

- **Crash** — copy/pasting a block of linked effects while a render was still draining. That's the stack trace in the description.
- **Freezes** — roughly 30 seconds of beachball when changing the render style on a linked effect, and again when dragging one to a new spot.

The 986-render-job figure is measured from that file rather than extrapolated, and the 986 → 986 result for time-range merging is why I went with per-model coalescing instead.

Worth noting the scaling: the drag path fanned out on *every* `SetStartTimeMS`/`SetEndTimeMS`, so it got worse the more you reuse a symbol — which is exactly backwards from what the feature is for. Even a symbol with a few dozen links would have felt it, just less dramatically.

Happy to attach the sequence if it'd help you reproduce.

derwin12:
Nah .. this is your code .. leaving it to you :-)

PR #6896, opened 2026-08-10
Fix layout undo destroying every Model Set in the show

heffneil:
## The bug

Undoing a layout change silently deleted **every Model Set in the show**. The next save made it permanent.

`LayoutPanel::DoUndo` for an `"All"` entry restores models by calling `ModelManager::LoadModels` directly, and `ModelManager::clear()` takes the set manager down with it. The undo snapshot only ever serialized `<models>` and `<modelGroups>`, so there was nothing to restore the Sets from — and `xLightsFrame::LoadModels`, the only code that reloads them, is on the show-open path that undo bypasses.

29 call sites push `"All"` undo points (align, distribute, flip, resize, delete model/group, paste, replace). Worse, `CreateUndoPoint` promotes a *single-model move* to `"All"` when that model is in a Set — so **moving a Set member and pressing Ctrl+Z**, the core Sets workflow, was the likeliest trigger.

Found by walking a user's `Backup/` folder, which is `_OnStart`-only, so one sample per launch gives a clean timeline:

| Backup | Sets |
|---|---|
| 07-27 17:37, 19:30 | `Busts` (4 props) |
| **07-28 06:51** | **0** |
| 07-30 10:52 | `Octopus` (7 props) |
| **07-30 11:30** | **0** |

## The fix

Serialize `<modelSets>` into the `"All"` snapshot and reload it in `DoUndo` **after** `LoadModels`, which is what clears it. Undo now also correctly reverses Set create/delete, which it could never do before.

Restoring is guarded on the node being present, so a snapshot taken before this change leaves Sets alone rather than wiping them — the fix can't reintroduce the bug through a stale undo-buffer entry.

Cleanups that fell out of it:

- The write is factored into `XmlSerializer::SerializeAllModelSets`, shared with `xLightsFrame::SerializeModelSets`.
- The `>= 2 members` / non-empty-name rule moves to `ModelSet::IsPersistable()`. This fixes a real round-trip asymmetry: `Save` allowed an unnamed Set that `Load` would then silently drop.
- Node and attribute names move to `XmlNodeKeys` instead of being repeated as literals across five sites.

## iPad

Not exposed. The iPad's layout undo is a value snapshot of a single object (`PushLayoutUndoSnapshotForModel` -> `UndoLastLayoutChange`), not serialized XML — it never calls `LoadModels`, so `ModelManager::clear()` isn't on its undo path. Parity scorecard row 47a added to `plans/platform-parity/06-layout-models.md`.

## Testing

- macOS Debug: pass
- macOS Debug with `NO_PCH`: pass (confirms the new `XmlNodeKeys` includes aren't PCH-masked)
- Manual: drag a Set member -> Ctrl+Z -> Set survives; align / flip / delete -> Ctrl+Z -> Sets survive; undo a *Link as Set* -> correctly unlinks

No new source files, so no `.cbp` / `.vcxproj` / CMake changes needed.


dkulp:
Review notes — the serialization side checks out. `XmlNodeKeys::NameAttribute == "name"` and `ModelSetModelsAttribute == "models"`, so the literal→constant swaps in `ModelSet::Load`/`Save` and `ModelSetManager` are behavior-preserving. `SerializeAllModels` closes both `<models>` and `<modelGroups>`, so `<modelSets>` lands as a sibling under `<root>` and `mroot.child("modelSets")` finds it. The restore is correctly ordered after `ModelManager::LoadModels` (which does `clear()` → `_setManager.Clear()`) and after `LoadGroups` (which doesn't clear). The empty-set case still emits a real `<modelSets>` element, so undoing back to a no-sets state genuinely clears sets instead of silently keeping them. `IsPersistable()` matches the old load-side rule exactly, and the only writer whose behavior changes (unnamed sets now dropped) was already producing sets that `Load` discarded. No include cycle from `XmlSerializer.h` → `models/ModelSetManager.h`, and the other two `AllModels.LoadModels` callers already reload sets afterwards.

Two things worth addressing:

### 1. `src-ui-wx/layout/LayoutPanel.cpp:2936` — the Set mutation handlers don't create an undo point

None of `DoLinkAsSet` (:2936), `DoAddSelectedToSet` (:2971), `DoRemoveSelectedFromSet` (:3000), `DoDeleteSet` (:3016), `DoRenameSet` (:3038) or `DoManageSet` (:3168) calls `CreateUndoPoint`.

Now that `"All"` snapshots carry sets, an unrelated undo rolls Set edits back silently. Move a model (that snapshot has no sets) → *Link as Set* on four props → Ctrl+Z. What the user expects is the set link to be undone, and the PR description says undo now reverses Set create/delete — but what actually happens is the earlier *move* is reverted **and** the brand-new Set is destroyed, with no way to redo it. Symmetrically, deleting a Set and pressing Ctrl+Z only resurrects it if some unrelated `"All"` op happened to be the previous stack entry.

Having the Set handlers push `CreateUndoPoint("All", ...)` before mutating closes both halves and would make the PR's claim actually true.

### 2. `src-ui-wx/layout/LayoutPanel.cpp:10711` — the `if (setsNode)` comment describes protection it doesn't provide

The stated rationale is "a snapshot taken before this change has no node — leave sets alone rather than wiping them", but that can't hold. `undoBuffer` is an in-memory `LayoutPanel` member, so within any single process every `"All"` snapshot comes from the new code and always has the node. And in the hypothetical where it didn't, `LoadModels` three lines above has *already* wiped the set manager, so skipping the restore leaves sets destroyed, not "alone."

The guard itself is harmless defensive code — it's just the comment that's misleading, and worth rewording so a future reader doesn't rely on it.

heffneil:
Both addressed in 52c28d736.

**1. Set handlers now push their own undo point.** Verified all six had zero `CreateUndoPoint` calls. Each now snapshots before mutating.

The handlers that prompt part-way through a mutation loop (`DoAddSelectedToSet`, `DoRemoveSelectedFromSet`, `DoManageSet`) go through a small `CreateSetUndoPointOnce(bool& taken, ...)` helper, so the snapshot is taken on the first *actual* change rather than on entry — otherwise declining every "already in Set X, move it?" prompt would strand an undo step that reverts nothing. Both rename paths check `GetSetByName(newName) == nullptr` before snapshotting, since `RenameSet` leaves the Set untouched on a collision and the re-prompt loop can still be cancelled.

One refinement to the scenario: a plain single-model move stays a `"SingleModel"` snapshot, which carries no sets, so that specific sequence is safe. The hazard needs the previous entry to be `"All"` — which, with 29 such call sites (align, distribute, flip, resize, delete, paste, replace) plus the Set-member move promotion at :10579, is the common case. The conclusion holds; only the example needed narrowing.

**2. Comment reworded.** You're right that the stated rationale can't hold — `undoBuffer` is a `std::vector<UndoStep>` member cleared on file load (`LayoutPanel.h:659`, `ClearUndo` at `:404`), so every snapshot in a process carries the node, and `LoadModels` has already cleared the set manager three lines up. The guard stays as cheap defensive code; the comment now describes what the node actually does — an empty `<modelSets>` is what makes undoing back past a Set's creation remove it.

macOS Debug and `NO_PCH` both pass.

PR #6897, opened 2026-08-10
iPad: undoing a Model Set edit restores every member

heffneil:
## The bug (iPad only)

Dragging a Model Set member on iPad translates every peer — `moveModel:` resolves the Set (`XLMetalBridge.mm:2301`) and calls `TranslateModelSetPeers` (`:2354`). But the drag pushed an undo snapshot only for the *grabbed* model (`PreviewPaneView.swift:930`, `:955`, `:1121`), and `UndoLastLayoutChange` popped a single entry.

So: drag a Set, every member moves. Undo, and only the grabbed model comes back — the rest stay where the drag left them, and the Set is broken apart visually.

Desktop already handles this by promoting a Set member's move to a full `"All"` undo snapshot (`LayoutPanel.cpp:10579`). The iPad's undo is a value snapshot of a single object, so it had no equivalent.

## The fix

`LayoutUndoEntry` gains a `groupId`. `PushLayoutUndoSnapshotForModel` resolves the Set containing the model and snapshots every member under one shared id; `UndoLastLayoutChange` pops the whole run when the top entry carries a non-zero id. One undo restores the entire Set.

Doing the expansion inside the push function means all three drag sites plus every property-edit call site inherit it — no Swift changes and nothing to keep in sync as new call sites appear.

Supporting cleanups:

- Capture and apply are split into `CaptureModelUndoEntry` / `ApplyLayoutUndoEntry` so the multi-entry push and the multi-entry pop reuse them instead of duplicating the field lists.
- The three copies of the depth-trim loop collapse into `TrimLayoutUndoStack`, which drops **whole groups** — otherwise reaching the 100-entry cap could leave half a Set restorable. It also refuses to drop a group spanning the entire stack, so a Set with more members than the cap can't silently discard the snapshot just taken.

`canUndoLayoutChange` (a bool) is the only undo-state accessor the UI reads, so the extra entries aren't visible anywhere.

## Testing

- `xLights-iPadLib` Debug, `-destination 'generic/platform=iOS'`: pass
- iPad-only change, so no desktop build applies and no `.cbp` / `.vcxproj` / CMake entries are needed (no new source files).

Not exercised on device — the change is compile-verified only.

## Parity

Scorecard row 244a added to `plans/platform-parity/06-layout-models.md`. No `README.txt` entry, per the iPad-changes convention in AGENTS.md §10.

### DIFF
--- a/plans/platform-parity/06-layout-models.md
+++ b/plans/platform-parity/06-layout-models.md
@@ -343,6 +343,7 @@ read-only category header (`ModelPropertyAdapter.cpp:307`); the iPad likewise sh
 | 243 | Escape finalizes a poly-line / unselects in 3D; Return finalizes | `LayoutPanel.cpp:9888-9905` | ✅ | `LayoutEditorView.swift:4161-4177` |
 | **Undo / save** |
 | 244 | Undo point per layout operation | `CreateUndoPoint` `LayoutPanel.cpp:10779`, `DoUndo` `:10597` | ✅ | `pushLayoutUndoSnapshotForModel:` / `…ForViewObject:` / `…TerrainHeightmap` (`XLSequenceDocument.h:2002`, `:2007`, `:2013`), undo `:2018`, can-undo `:2021` |
+| 244a | Undoing a Model Set edit restores every member | Snapshot promoted to `"All"` when the model is in a Set, `CreateUndoPoint` `LayoutPanel.cpp:10579` | ✅ | `PushLayoutUndoSnapshotForModel` snapshots all Set members under one `LayoutUndoEntry::groupId`; `UndoLastLayoutChange` pops the whole run (`iPadRenderContext.cpp:1416`, `:1497`). Previously restored only the grabbed model and left the peers displaced, since a Set drag translates every peer (`XLMetalBridge.mm:2354`) |
 | 245 | Save layout | `ButtonSavePreview` `LayoutPanel.cpp:1092`, handler `:4435`; disabled read-only `xLightsMain.cpp:9307` | ✅ | Canvas Save with unsaved-dot `LayoutEditorView.swift:3100`, bridge `XLSequenceDocument.h:1421`, dirty flag `:1987` |
 | 246 | Confirm-before-save with explicit discard-and-rollback | Desktop saves directly | 🔵 | `LayoutEditorView.swift:1052-1065`, `discardChanges` replays the undo stack `:4623` |
 | **Added by the 2026-08-01 cross-check** (appended, not renumbered — see the numbering rule) |
--- a/src-iPad/Bridge/iPadRenderContext.cpp
+++ b/src-iPad/Bridge/iPadRenderContext.cpp
@@ -28,6 +28,7 @@
 #include "effects/ShaderEffect.h"
 #include "models/Model.h"
 #include "models/ModelGroup.h"
+#include "models/ModelSet.h"
 #include "models/MeshObject.h"
 #include "models/ImageObject.h"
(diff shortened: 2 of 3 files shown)


dkulp:
Review notes — the core mechanism looks sound. Groups are only ever appended at the back and removed as complete runs from either end, so a group can never be split and the trim loop always terminates, and `applied = ApplyLayoutUndoEntry(e) || applied` correctly avoids short-circuiting. Splitting capture from apply is a nice improvement too — a missing peer now skips one entry instead of aborting the whole undo.

The items below all come from doing the Set expansion unconditionally inside the push function.

### 1. `src-iPad/Bridge/iPadRenderContext.cpp:1420` — undo of a single-model edit rewrites every Set peer's controller/preview/lock/size

The expansion applies to *all* callers, not just body drags: `deleteVertexAtIndex`, `insertVertexInSegment`, `setCurve` (`XLSequenceDocument.mm:5566/5582/7128`), arrow-key nudge (`LayoutEditorView.swift:4193`), and every property commit.

`ApplyLayoutUndoEntry` restores `layoutGroup` and `controllerName` on each entry, and several mutators push no snapshot at all — `assignModelToController` (`XLSequenceDocument.mm:18666`), `removeModelFromController` (`ControllerVisualizeView.swift:598`), and the `cc.*` commits in `ModelControllerPropertiesSheet.swift:216`.

Scenario: Set {A,B}; drag A (the group snapshot captures B's empty controller); go to the Controller Visualiser and assign B to a port; back to Layout, press Undo to revert the drag — B is silently unassigned again. Before this change undo could never touch B.

Narrowing the expansion to the operations that actually move peers, or capturing only the translated fields for peers, would avoid this.

### 2. `src-iPad/Bridge/iPadRenderContext.cpp:1383` — the 100-entry cap is per *entry*, so a Set collapses usable undo depth

`kLayoutUndoMaxDepth` counts entries, but a gesture on an M-member Set now costs M entries. With a 20-member Set, five arrow-key nudges (one push per tap, `LayoutEditorView.swift:4193`) fill the stack, and `TrimLayoutUndoStack` drops whole groups from the front — so those nudges wipe the entire earlier undo history rather than the oldest step or two. For M > 100 the protective `break` also leaves the stack permanently above the cap. Counting gestures rather than entries would restore the depth the constant advertises.

### 3. `src-iPad/App/LayoutEditorView.swift:4358` — align/distribute/match-size push one group per selected name

`performAlign` (`:4358`), `performDistribute` (`:4381`) and `performMatchSize` (`:4404`) push a snapshot for every selected model. The bridge side is deliberately Set-aware and de-duplicates (`doneSets` / `seenSets`, `XLMetalBridge.mm:1350`, `:1404`), but the push side is not — so selecting K members of the same Set pushes K identical M-entry groups. The user then has to press Undo K times, K−1 of which are visible no-ops, and K×M entries are burned against the cap. Mirroring the apply side's de-duplication on the push side (`GetSetContaining`, skip already-seen sets) fixes both.

### 4. `src-iPad/Bridge/iPadRenderContext.cpp:1424` — the "model does not exist" early-out was dropped

The old code did `Model* m = AllModels.GetModel(modelName); if (!m) return;` before pushing anything. The new path resolves the Set first, so a push for a name that is a stale Set member but no longer in `AllModels` still pushes snapshots for the surviving peers and leaves a phantom undo entry for an operation that never happened; pressing Undo then rewinds those peers to a stale position. `ModelManager::OnModelDeleted` (`ModelManager.cpp:2228`) means this needs a stale index to trigger, but the guard is cheap to re-add.

Issue #6801, opened 2026-07-30
Additional Tool Bar Items


deplanche:
With some recent changes in the Toolbars, including the ability to customize or hide them entirely, it would be nice to have a "Tools" toolbar.  The icons for that that I think would be most useful would be:

- FPP connect
- Test
- Package Sequence

These get used fairly often, or at least should be used fairly often.  A Package sequence button would be most used by vendors, but I also use that when someone is sending me a sequence that have a problem with, I make the changes, and send it back to them.

cybercop23:
FPP Connect has already been submitted as a PR, yesterday.

cybercop23:
@computergeek1507 are you okay with adding this as it "own" toolbar, that won't be auto selected, since we can do that now, and users can turn it on themselves?

cybercop23:
PR ready.

PR #6920, opened 2026-08-13
Add additional tools toolbar (#6801)

cybercop23:
Add "utils" Tools Toolbar #6801


derwin12:
Post a screenshot for folks.

cybercop23:
image

Should have mentioned:
Test, Check Sequence, Package Sequence, Batch Render, FPP Connect, Bulk Controller Upload, View Logs

computergeek1507:
Today most of the tool bars are related to "Sequencer" actions, other tab(s) actions are confined to the specific tab. This could muddy the water and make it more confusing to the end user what toolbar is related to what. I understand you and other power users will understand, but the normal user doesn't know what the light bulb or color pallet represents

cybercop23:
It is off by default, and only if you want to have it, you can turn it on, so that should not confuse regular uses that never touch them. Many we see on zoom also have the AC toolbar on, don't ask us why.
I think it should be safe and not confusing folks - unless they turn it on, and then they have tooltips to tell them what they are.

computergeek1507:
AC toolbar is a sequence related action, makes sense to me.

computergeek1507:
Are you going to add toolbars for the layout actions and other utils? Is the tool only things the zoom room deems worthy?

cybercop23:
> Are you going to add toolbars for the layout actions and other utils? Is the tool only things the zoom room deems worthy?

It was added because of  #6801. Seems to make sense. You were tagged in it too.

derwin12:
Do tooltips pop on each icon?

computergeek1507:
this goes against the whole UI design of xLights, Currently all the toolbars are "Sequence Actions" now you are mixing not "Sequence Actions" to the Sequence Tab. xLights UI is split into "Configuation Tab" and "Sequence Tab" (I understand the toolbars are in the layout tab but inactive), this change with throw the UI design out the window and spam buttons everywhere. Then someday in the future **if** we separate the windows this toolbar will be in the wrong windows as it mixes different utilities.

dkulp:
IMO...    " in the future if we separate the windows this toolbar will be in the wrong windows"  is wrong.  This toolbar would be then be on all the windows that need it.   That said, worrying about the future window layouts and stuff is not something I'd worry about now.   If we move from wxWidgets or re-design the UI or whatever, EVERYTHING will need a complete re-think, not just the toolbars.   For example, on the Mac, the entire Menu  should change to reflect the active window.  

If this can put the users tools they frequently use into a place that is more easily accessible instead of digging through menus, it's a good thing.   I hate our menus.   They are a mess.   (even with the recent cleanups).

dkulp:
That said, I don't like the FPP icon in the above screen shots.  It's smaller than the others.

cybercop23:
@derwin12 yes, the tooltips show on each icon.
@dkulp I think that's the best Icon, if you agree I can make it larger.

I do have additional changes I've been thinking about, specifically about only showing what's relevant, like only show the effects toolbar when on the sequencer screen - since that's the largest and give back "height" on the layout tab.

cybercop23:
image

Issue #6841, opened 2026-08-04
confirm canvas 'model blending' layer is automatically selected for sequences prior to introduciton


sidelights:
NOTE: IF YOU DO NOT ATTACH A SEQUENCE THAT DEMONSTRATES YOUR PROBLEM THEN THERE IS A HIGH PROBABILITY YOUR ISSUE WILL JUST BE CLOSED AND NOT INVESTIGATED.

**Describe the bug**
sequences created prior to the introduction of the 'model blending' checkbox in canvas mode defaulted to the result of having model blending turned on in canvas model layer 'model blending'.  However I suspect that there is not a version check on the sequence last created prior to 2024.09 does not default model blending to be set to on.  As seqeunces being reviewed in 2026.14 show up as if it is not checked, and in fact is not checked and effects render different. 

**To Reproduce**
(and/or link to a short video showing the problem with audio track describing what you are doing)
in a version prior to model blending in canvas. 

Create the following
1.  on a matrix or other large group
2. 3 layers
3. transparent off effect on the first layer
4. 2nd layer shockwave with radius large enough to cross the entire matrix or group starting from the bottom center. (normal blending) 
5. 3rd layer shockwave with radius large enough to cross the entire matrix or group starting from the top center. (normal blending) 
6.  on the off effect check canvas and all layers (2,3)

Result should be a shockwave effects crossing each other. 2 of them one from the top and bottom.

Save sequence and open in 2026.14. 

Result should be only one shockvalue appears.   As canvas (model blending) is not checked.  Fix is to check model blending in canvas layers.   

8.  **Expected behavior**A clear and concise description of what you expected to happen and why.**Screenshots**

model blending should automatically be checked as it was the default behavior prior to the introduction. 

**Versions (please complete the following information):**
 - OS: [e.g. Windows 10, Mac 10.14.5, etc...]
 - xLights version [e.g. 2019.37]

**Additional context**

**Attachments**
- If associated with a crash please attach/link to a dropbox containing the crash log that was produced
- If it crashed but no crash log was produced please use Tools Menu/Package Logs and attach/link to a dropbox containing the resulting zip file
- If the error relates to rendering an effect or a crash while sequencing please use Tools Menu/Package Sequence and attach/link to a dropbox containing the resulting zip file

If you would prefer not to post your sequences publically please say so and be ready to PM them to one of the developers when asked.

derwin12:
@dkulp  Dan -- this is your territory :-)

[...]

PR #6921, opened 2026-08-13
Handle the legacy model blending (#6841)

derwin12:
In 2024 the last option in the canvas layers was Model Blending but was stored as an extra layer. Now we expect it to say "Blend".
So 0|1|2, the 2 meant layer blend. 
The code now just throws away the 2 - sort of - it shows Model Blend as checked but it is ignored unless you go in and go out.
It needs to be stored as 0|1|Blend.
This change handles both cases ..

Going to need some second/third opinions on this one..

### DIFF
--- a/src-core/render/RenderEngine.cpp
+++ b/src-core/render/RenderEngine.cpp
@@ -1387,6 +1387,19 @@ public:
                         if (!ls.empty() && ls.back() == "Blend") {
                             doBlendLayer = true;
                             ls.pop_back();
+                        } else if (!ls.empty() && _seqElements != nullptr && _seqElements->SupportsModelBlending() &&
+                                   std::atoi(ls.back().c_str()) >= numLayers - layer - 1) {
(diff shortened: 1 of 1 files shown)


cybercop23:
Should we have the usual check, if sequence saved/created before 2024.9 turn that option on at blending? If only at render time, what happens when you import from an older sequence? Wouldn't we want that option turned on automagically?

derwin12:
This would catch the bad settings and correct it even after they saved in a newer version.

cybercop23:
But should we? If sequenced after that option was introduced and that wasn't checked, should we still force it?
Won't this change the look if the creator didn't check that option?

derwin12:
In the prior version if layers count was greater than the actual layer count, it deemed Model Blending was enabled. We still do that . it show the Model Blending checked - BUT - it is not really checked internally, unless you uncheck/check it again. This honors that checked status.

Issue #6831, opened 2026-08-02
Add Effect Icon Size and Effect Grid Spacing to what is saved to a Perspective


mailgod:
I typically switch between a monitor and my laptop screen while working in Xlights. I have a Full Monitor and Laptop perspective that I have created to better utilize the screen when switch between the two. But I typically have to manual resize the Effect Grid and the effect icon size when switching perspectives. Would really be helpful if these setting could be included when saving a perspective.

cybercop23:
Not that this is the fix, but... you can now choose what toolbars you want to see, View -> Toolbars, you can also move them (some) to the left side and in 14 you can choose what effects you want to see in that list.

Saving sizes with perspectives, I think, makes sense.

AGFazio:
PR 6923 in review

PR #6923, opened 2026-08-13
#6831-Keep-Effect-Grid-and-Icon-Size-with-Perspectives

AGFazio:
Perspectives now remember the Effects Grid Spacing (row height) and the Effect Icon Size, and restore them when the perspective is loaded. Switching between a perspective built for a large monitor and one built for a laptop screen no longer means resetting both sizes by hand each time.

What changed

- Perspective gains gridSpacing and iconSize fields, persisted as optional gridSpacing/iconSize attributes on <perspective> in xlights_rgbeffects.xml.
- Both values are captured wherever a perspective is saved: default-perspective creation, Save Perspective, Save As (which delegates to Save Perspective), and the 1.0→2.0 migration re-save.
- On load, ApplyPerspectiveViewSettings applies them after the AUI layout is restored. Only the sizes the preference panels actually offer are accepted, so a hand-edited or foreign rgbeffects file can't set an unusable row height. Each is skipped when it already matches the current value, avoiding a needless toolbar relayout.

Backward compatibility

Perspectives saved by earlier versions have no attributes and load as 0, which means "leave the current sizes alone." They pick the sizes up the next time the perspective is saved or migrated. Older xLights versions ignore the two new attributes, so the file stays readable both ways.

iPad parity

Captured as State B — new row 79a in plans/platform-parity/10-presets-views-perspectives.md, marked ❌ and folded into the existing rows 74–79 gap. The iPad has no perspective concept to attach the settings to, and no equivalent of either preference: row height comes from its own grid metrics and there is no effects toolbar to size.


computergeek1507:
What happens to the preferences?

computergeek1507:
These settings are currently in the preferences, which takes priority, why have it in two locations?

AGFazio:
They stay with what was loaded from the perspective. This way if you switch show folders your perspective size setting stays even though your show may not have a perspective set. You won't need to manually change them.

computergeek1507:
If I load a perspective or have auto save perspective enabled, then go into the preferences and change the size, what happens? Does the new size get saved in the perspective too? Then does loading another perspective override the preference?

AGFazio:
Not in 2 locations just saved with the perspective. So changing a perspective "laptop screen" which has small saved, then switching to "Large Monitors" which has large icons saved, doesn't require going to the preferences and changing them.

computergeek1507:
It's already in preferences

computergeek1507:
image

computergeek1507:
image

AGFazio:
Loading a perspective that has different sizes selected than what is currently populated in preferences would change the preferences fields to match the current perspectives. So if another show folder was opened the effects grid and icon sizes would stay what they currently are from the current perspective. even though that perpestive may not exist. If auto-save is enabled and you have say small saved in a perspective and then switch preferences to large, the new preference would be saved in the perspective, as I think it would be expected using auto-save. If not wanted, Perspective could be changed to explicit save only, where the change only save when you tell it to.

derwin12:
Seems to work .. it just flips the preference as needed..

PR #6927, opened 2026-08-14
Filterable pickers everywhere, with ticked items pinned to the top

heffneil:
Supersedes #6858, whose commit is included here.

## Filters where there were none

Five model/element pickers were still a plain `wxMultiChoiceDialog` with no filter, which is painful once a show has a lot of submodels or states:

| Picker | Where |
|---|---|
| Select SubModel | `ModelFacesPanel`, `ModelStatesPanel` |
| Select States | `ModelStatesPanel` |
| Select Sub-state(s) | `SubModelsPanel` |
| Select face elements | `SubModelsPanel` |
| Add / Remove to Existing Groups | `LayoutPanel` (from #6858) |

They now use the existing filterable `CheckboxSelectDialog`, the same one the Export-to-Other-Models pickers already use.

Selection is read by name via `GetSelectedItems()` rather than by index into the choices array. That returns items in `_allItems` order, so callers that depend on ordering — `Select States` assigns sequential state indices as it iterates — behave exactly as before.

## Ticked items go to the top, and stay visible

`CheckboxSelectDialog::PopulateList` now lists ticked items first, then everything matching the filter, each block name-ordered using the same number-aware comparator the Import Previews tree uses (so `Tree-2` precedes `Tree-10`).

Ticked items are also listed when the filter *stops* matching them. Their check state was already preserved — `_checked` has always been the source of truth — but the row disappearing made a selection assembled over several searches impossible to review, which reads as though the ticks were thrown away.

`PopulateList` runs only on construction and on a filter change, **never on a tick**, so an item can't jump out from under the cursor mid-selection.

**This re-sorts every caller's list.** Most were already alphabetical (the group pickers and all four Export pickers sort in the caller). Two were not: `Select SubModel` was in model definition order, and `Select Sub-state(s)` in map order over *values*, so effectively arbitrary. Both read better sorted, but it is a visible change and worth a look.

Thirteen call sites share this dialog, so they all gain the behaviour together.

## Import Previews/Models

Same treatment for its tree: ticked models, groups, viewpoints and preview rows all survive a filter that would otherwise exclude them.

Three separate spots needed it — the item kind is threaded through `collect()` because `_checkedModels` is keyed on `{name, kind}`, so a ticked group and a ticked model of the same name don't alias; and the empty-row prune now spares a preview row that is itself ticked, which would otherwise vanish even after its children were kept.

## Testing

- macOS Debug: pass
- Manual, all eight dialogs: tick several, filter them out, confirm they stay listed, ticked, and sorted to the top; confirm OK returns the full set.

`wxSingleChoiceDialog` pickers (Select Model, Select State, Select Face, Aspect Ratio, Order Type) are deliberately untouched — pick-exactly-one has no tick state to preserve. No new source files, so no `.cbp` / `.vcxproj` / CMake changes.

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -77,6 +77,7 @@
 #include "model/WiringDialog.h"
 #include "model/ModelDimmingCurveDialog.h"
 #include "UtilFunctions.h"
+#include "shared/dialogs/CheckboxSelectDialog.h"
 #include "shared/utils/ExternalHooksUI.h"
 #include "color/ColorManager.h"
 #include "utils/VectorMath.h"
@@ -3658,7 +3659,10 @@ void LayoutPanel::AddSelectedToExistingGroups() {
         return;
     }
 
-    wxMultiChoiceDialog dlg(this, "Select existing groups to add selections to", "Existing Group", choices);
+    // Filterable picker - group lists get long. Groups you have already ticked
+    // stay visible when you change the filter, so you can search, tick, search
+    // again and tick more without losing your earlier choices.
+    CheckboxSelectDialog dlg(this, _("Select existing groups to add selections to"), choices);
     OptimiseDialogPosition(&dlg);
 
     std::string selectgroupName;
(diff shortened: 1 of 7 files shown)


heffneil:
Screenshots:

Screenshot 2026-08-14 at 12 39 31 PM
Screenshot 2026-08-14 at 12 39 27 PM
Screenshot 2026-08-14 at 12 39 17 PM
Screenshot 2026-08-14 at 12 39 09 PM

heffneil:
Can we please review this?  I think it will be helpful to folks as they are working on their layouts and want to add models to groups.

derwin12:
I looked at this and tested a bit .. I see there is a sort in there .. we have to watch that we dont sort subs and states automagically (I was not able to make it do this - good). EFL and others have subs and states in a certain render order and we offer that ability to sort them via drag n drop.

PR #6929, opened 2026-08-14
Display Elements: find bar for the Added models list

heffneil:
The Available list got a filter in #6241. The **Added** list — the models actually in the sequence view — had nothing, and that's the one that gets long.

## Why a find bar and not a filter

The Added list's row order **is** the view's display order. Hiding rows breaks:

- drag-reorder and the Move Up / Down / Top / Bottom buttons — a drop position maps to the wrong index in the real order
- Remove All, Hide All, Show All, Hide Unused, Remove Unused — each walks the list control, so they'd silently act on the visible subset while their names promise the whole view

An earlier revision of this branch added guards disabling all of the above while filtering. That works but leaves the feature useless for its main purpose: you search to *find* a model, usually because you want to move it. A find bar removes the hazard rather than defending against it — the list is never incomplete, so every operation stays correct and enabled.

## Behaviour

A search box with ▲ / ▼ above the list.

- Typing searches from the current selection, so extra characters narrow onto the same match rather than jumping away
- ▼ / Enter step forward, ▲ steps back, both wrapping at either end
- A hit is selected and scrolled into view; a miss tints the box rather than silently doing nothing
- Reordering is untouched and works with a search active

Matching is a **whitespace-tokenised AND**, matching `CheckboxSelectDialog` and the Import Previews tree: `mini tree` finds `Mini-Tree-2` and `MiniTree-Star`, where a contiguous `Contains()` required the separator to be typed exactly. Word order doesn't matter.

No debounce needed — nothing rebuilds, it's a linear scan over the visible rows.

## Also fixed

Two things in the neighbouring Available filter:

- It acted on `wxEVT_TEXT` events that didn't change the value — holding backspace on an already-empty field rebuilt both lists and flickered for nothing.
- `PopulateModels` now freezes the whole panel rather than the two lists separately, so filtering one list no longer flashes the other.

## Testing

- macOS Debug: pass
- Manual: `mini tree` matches hyphenated and run-together names; ▲▼ cycle and wrap; miss tints the box; drag and the move buttons behave normally with a search active.

No new source files, so no `.cbp` / `.vcxproj` / CMake changes.

## Note for reviewers

This wraps `ListCtrlModels` in a box sizer by detaching it from the GridBagSizer and re-adding — the same pattern the shipped Available-list filter already uses on `ListCtrlNonModels` two blocks above. I saw a wxWidgets Debug assert (`SetContainingSizer(): Adding a window already in a sizer`) when reopening the Display Elements pane while testing. Instrumentation showed the panel constructor is clean — the detach returns true and leaves no containing sizer — so it appears to come from the wxAUI pane show/hide path rather than this code, and the pattern it would implicate is already in master. Flagging it rather than leaving it unsaid; if a reviewer can reproduce it on plain master, that confirms it's pre-existing.

### DIFF
--- a/src-ui-wx/layout/ViewsModelsPanel.cpp
+++ b/src-ui-wx/layout/ViewsModelsPanel.cpp
@@ -19,6 +19,7 @@
 #include <wx/artprov.h>
 #include <wx/dnd.h>
 #include <wx/srchctrl.h>
+#include <wx/settings.h>
 #include <wx/timer.h>
 
 #include "model-16.xpm"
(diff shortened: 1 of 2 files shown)


heffneil:
Find Next and Find Previous.  This preserves order on the already assigned elements in the view.  This adds the ability to search for models contextually and not just the first X letters.

Screenshot 2026-08-15 at 7 11 32 AM

heffneil:
Screenshot 2026-08-15 at 7 11 32 AM

PR #6945, opened 2026-08-18
Layout filters: match multiple search terms (Models, Groups, Controllers)

heffneil:
## What
The Layout-panel **Models**, **Groups**, and **Controllers** search boxes treated the whole entry as one literal string, so a multi-word search failed: typing `midwest pumpkin` found nothing even though `Midwest-Coro-Pumpkin-1` contains both words (just not that contiguous string). Searching `pumpkin` alone worked.

This makes those three filters tokenize on whitespace and require **every** term to appear in the name, in any order — so `midwest pumpkin` now finds `Midwest-Coro-Pumpkin`. This matches the behaviour of the newer filters already in the app (vendor model catalog #6733, the Choose-Models / Export-to-Other-Models pickers #6670 / #6751).

## How
- `LayoutPanel::MatchesFilter` (drives both the Models and Groups tabs) and `ControllerListPanel::ControllerMatchesFilter`: split the filter text on whitespace; if there's more than one term, every term must be a case-insensitive substring of the model/controller name (AND).
- **A single term keeps the existing behaviour**, including the regex power-search (`filterRegex.Matches`) — so no existing single-word or regex search changes.

```cpp
wxArrayString terms = wxStringTokenize(filterString.Lower(), " \t");
if (terms.size() <= 1) {
    if (filterRegexValid) return filterRegex.Matches(name);   // unchanged
    return name.Lower().Contains(filterString.Lower());        // unchanged
}
for (const auto& term : terms)
    if (!name.Lower().Contains(term)) return false;
return true;
```

Added `#include <wx/tokenzr.h>` to both files.

## Testing
Built (macOS Debug). Verified `midwest pumpkin` matches `Midwest-Coro-Pumpkin`, single-word and regex searches behave as before, and the Groups/Controllers tabs get the same behaviour through the shared code paths.

## iPad parity
Desktop-only (`src-ui-wx/layout/`); the iPad model list has its own filtering path. No iPad counterpart to change here.

### DIFF
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -13175,10 +13175,6 @@ void LayoutPanel::OnGroupFilterTextChanged(wxCommandEvent& event) {
 bool LayoutPanel::MatchesFilter(Model* model, const wxString& filterString, const wxRegEx& filterRegex, bool filterRegexValid) {
     if (filterString.IsEmpty()) return true;
 
-    // A single term keeps the existing behaviour (regex if it compiles, else a
-    // substring match). Multiple whitespace-separated terms narrow with AND -
-    // every term must appear somewhere in the name, in any order - so
-    // "midwest pumpkin" finds "Midwest-Coro-Pumpkin". Matches the other filters.
     wxArrayString terms = wxStringTokenize(filterString.Lower(), " \t");
     if (terms.size() <= 1) {
         if (filterRegexValid)
--- a/src-ui-wx/layout/LayoutPanel.cpp
+++ b/src-ui-wx/layout/LayoutPanel.cpp
@@ -31,6 +31,7 @@
 #include <wx/propgrid/advprops.h>
 #include <wx/tglbtn.h>
 #include <wx/srchctrl.h>
+#include <wx/tokenzr.h>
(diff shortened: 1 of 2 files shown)


heffneil:
Old:
Screenshot 2026-08-18 at 5 27 52 PM

New:

Screenshot 2026-08-18 at 5 27 41 PM

src-ui-wx/layout/LayoutPanel.cpp
```diff
@@ -13174,10 +13175,23 @@ void LayoutPanel::OnGroupFilterTextChanged(wxCommandEvent& event) {
 bool LayoutPanel::MatchesFilter(Model* model, const wxString& filterString, const wxRegEx& filterRegex, bool filterRegexValid) {
     if (filterString.IsEmpty()) return true;
 
-    if (filterRegexValid)
-        return filterRegex.Matches(model->GetName());
+    // A single term keeps the existing behaviour (regex if it compiles, else a
+    // substring match). Multiple whitespace-separated terms narrow with AND -
+    // every term must appear somewhere in the name, in any order - so
+    // "midwest pumpkin" finds "Midwest-Coro-Pumpkin". Matches the other filters.
```
computergeek1507: probably dont need all these comments in source, makes moresense in the commit details

PR #6972, opened 2026-08-25
Don't let a zero default sequence duration brick new sequences

### LINKED ISSUE #6971: A default sequence duration of 0 makes every new Animation sequence unusable and unrecoverable
### Symptom

Creating a new **Animation** sequence produces a sequence that cannot be used:

- the effects toolbar is greyed out
- Display Elements cannot be opened, and views are empty
- **Sequence Settings, Save, Save As and Close Sequence are all disabled**

Musical sequences are unaffected. Nothing in the UI recovers the sequence — the one control that would fix it (Sequence Settings, to set a duration) is one of the disabled ones. Restarting does not help; every new Animation sequence behaves the same way.

### Cause

The **Set Default** button beside the sequence duration writes whatever is in the box to config with no validation:

```cpp
cfg->Write("DefaultSeqDuration", TextCtrl_Xml_Seq_Duration->GetValue());
```

It reports *"Saved!"* regardless. With `0` in the box, config ends up with:

```json
"DefaultSeqDuration": "0.000"
```

New sequences then read it back:

```cpp
std::string savedDur = cfg->Read("DefaultSeqDuration", std::string("30.0"));
```

The `"30.0"` fallback only applies when the key is **absent**, not when its value is useless — so the sequence gets a duration of 0, and therefore 0 frames:

```
XLDBG new: type=Animation durationMS=0 mediaMS=0 ms=25 frames=0 -> NumFrames=0
```

`EnableSequenceControls` treats 0 frames as "no file is loaded":

```cpp
if (enable && _seqData.NumFrames() == 0) {
    // no file is loaded, disable save/render buttons
    Menu_Settings_Sequence->Enable(false);
    MenuItem_File_Save->Enable(false);
    MenuItem_File_Close_Sequence->Enable(false);
    ...
```

Musical sequences escape this because their length comes from the audio rather than the stored default, which is what makes it look intermittent.

### Reproduction

1. Open Sequence Settings on any sequence, put `0` in Duration, press **Set Default**
2. Create a new **Animation** sequence
3. Toolbar, Display Elements, Sequence Settings and Save are all disabled, permanently, for every new Animation sequence from then on

### Affected

Present in released **2026.15** and in current master, so it would ship in the next release too. macOS observed; the code path is platform-independent.

### Suggested fix

Two independent guards, either of which prevents the dead end, both of which are worth having:

- **On read** — treat a non-positive or unparseable stored value as absent and fall back to 30s, so configs already carrying a bad value heal themselves. This is the important half: a write-side guard alone leaves every already-broken install broken.
- **On write** — refuse a non-positive duration (and timing) instead of storing it and claiming success.

The stored *timing* deserves the same read guard: it becomes the frame interval and is divided by, so a stored `0` is worse than a bad duration.

PR to follow.

heffneil:
Fixes #6971.

A stored default sequence duration of `0` made every new **Animation** sequence unusable, and unrecoverable from the UI.

## What happened

**Set Default** beside the duration wrote whatever was in the box to config with no validation, reporting *"Saved!"* either way:

```cpp
cfg->Write("DefaultSeqDuration", TextCtrl_Xml_Seq_Duration->GetValue());
```

New sequences read it back with a fallback that only fires when the key is **absent**, not when its value is useless:

```cpp
std::string savedDur = cfg->Read("DefaultSeqDuration", std::string("30.0"));
```

So the sequence got 0 frames, and `EnableSequenceControls` treats 0 frames as *"no file is loaded"* — disabling the effects toolbar, Display Elements, Sequence Settings, Save, Save As and Close Sequence. The control needed to set a duration is itself disabled, so there is no way back.

Musical sequences were unaffected because their length comes from the audio, which is why this presents as intermittent.

## The fix

Two independent guards:

**On read** — `GetDefaultSeqDurationSeconds()` treats a non-positive or unparseable stored value as absent and returns 30s. This is the half that matters: a write-side guard alone would leave every already-poisoned config broken. The three read sites now share it rather than each repeating `Read(..., "30.0")`.

**On write** — **Set Default** refuses a non-positive duration or timing and says why, instead of storing a value that guarantees a broken sequence.

The stored **timing** gets the same read guard. It becomes the frame interval and is divided by, so a stored `0` is worse than a bad duration.

## Testing

- macOS Debug: pass
- Verified against a real config still containing `"DefaultSeqDuration": "0.000"` — new Animation sequences come up at 30s with all controls live, without editing settings by hand
- New Musical sequence still takes its length from the audio
- **Set Default** with `0` in the box is refused; with a valid value it saves and is picked up by the next new sequence

No new source files, so no `.cbp` / `.vcxproj` / CMake changes.

### DIFF
--- a/src-ui-wx/import_export/SeqFileUtilities.cpp
+++ b/src-ui-wx/import_export/SeqFileUtilities.cpp
@@ -120,10 +120,11 @@ void xLightsFrame::NewSequence(const std::string& media, uint32_t durationMS, ui
 
     if (wizardactive) {
         auto* cfg = GetXLightsConfig();
-        std::string savedDur = cfg->Read("DefaultSeqDuration", std::string("30.0"));
-        CurrentSeqXmlFile->SetSequenceDuration(savedDur);
+        CurrentSeqXmlFile->SetSequenceDuration(GetDefaultSeqDurationSeconds());
+        // Timing becomes the frame interval and is divided by, so a stored 0
+        // would be worse than a bad duration.
         std::string savedTiming = cfg->Read("DefaultSeqTiming", std::string(""));
-        if (!savedTiming.empty()) {
+        if (!savedTiming.empty() && wxAtoi(savedTiming) > 0) {
             CurrentSeqXmlFile->SetSequenceTiming(savedTiming);
         }
     }
--- a/src-ui-wx/import_export/SeqFileUtilities.cpp
+++ b/src-ui-wx/import_export/SeqFileUtilities.cpp
@@ -12,6 +12,7 @@
 #include "settings/XLightsConfigAdapter.h"
 #include <wx/regex.h>
 #include <wx/tokenzr.h>
+#include <wx/wxcrt.h> // wxAtoi — MSVC needs it explicitly (macOS PCH masks the miss)
 #include <wx/uri.h>
 #include <wx/wfstream.h>
 #include <wx/zipstrm.h>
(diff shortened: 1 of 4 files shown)


derwin12:
It didnt compile on Windows.. can you make these changes..
image
and
image

PR #6980, opened 2026-08-27
Matrix faces: add movement/positioning/scaling controls matching Pictures effect

bulldozer2003:
Matrix faces: add movement/positioning/scaling controls matching Pictures effect

Adds 13 new properties to Faces.json (direction, speed, X/Y offset, start/end scale, wrap, pixel offsets, vector mode, frame rate adj, scaling override) so matrix faces work like a series of picture effects.

The existing cache strategy is preserved when no per-frame variation is active (direction=none, offsets=0, scale=100/100). When movement or animated scaling is enabled, the image renders fresh each frame.

Non-matrix face types (SingleNode, NodeRange, Coro, Rendered) are unaffected.

https://github.com/user-attachments/assets/0e538afb-8429-4df3-8a48-b9d4193a6304


derwin12:
Doesnt this duplicate the render buffer tools?

bulldozer2003:
No more than the Pictures effect duplicates them but the buffer tools do not have X/Y offsets nor the movement presets. The matrix faces already delegated rendering to PicturesEffect::Render() and this exposes the additional parameters instead of setting them to defaults of 0.

derwin12:
ok.. I need to check how it handles loading a sequence with this in it with prior versions.

bulldozer2003:
Thank you for your time. I tested a sequence with 2026.16 and found the extra parameters were ignored and removed when saving the sequence.

derwin12:
I am going to see if I can get the UI a bit more aligned with Pictures.
imageimage

derwin12:
Superseded with my changes to clean up the ui.

PR #6988, opened 2026-08-29
Bars Effect rounding issue

derwin12:
Found the Per Preview would wrap the color bars. @dkulp Can you check this as well.
3 color bars, no cycle, left.

image


computergeek1507:
that was a secret feature

derwin12:
Ah dang .. like the red butterfly pixel.

computergeek1507:
yeah

Issue #6995, opened 2026-08-31
Windows OS fonts render corrupted text on 35x7 matrix in xLights 2026.16


5yrnhsmtkv-sys:
[Rockin' Around the Christmas Tree PARTIAL TEXT ISSUE.zip](https://github.com/user-attachments/files/31624069/Rockin.Around.the.Christmas.Tree.PARTIAL.TEXT.ISSUE.zip)

**Describe the bug**

I have a small 35x7 matrix that I use basically as a Tune-To sign. My old sequence has `TUNE TO 88.7 FM` scrolling across it using a Windows font. It worked correctly in the older version of xLights, but after opening/rendering the sequence in 2026.16 the text gets corrupted.

As the text displays, pieces of individual letters or numbers seem to hang behind. When the text is scrolling, it looks like part of one character gets stuck until the next character comes along, then that character leaves another piece behind. It gives the text a strange grabbing/catching-up appearance.

I originally thought something was wrong with the old sequence, so I tried different fonts, retyping the text, changing movement, etc. The problem seems specifically related to using Windows OS fonts on this small matrix.

The corruption is visible directly in Model Preview, so it is not something happening at the controller or on the physical lights.

**To Reproduce**

1. Create a Horizontal Matrix configured as:

   * 1 string
   * 245 nodes/string
   * 7 strands/string
   * Top Left starting location
   * This produces a 35x7 matrix.
2. Add a Text effect.
3. Enter something like `TUNE TO 88.7 FM`.
4. Set XL Font to `Use OS Fonts`.
5. Select a Windows font. Mine originally used Segoe UI Semibold size 9.
6. Render/view it in Model Preview.
7. Parts of the letters/numbers are corrupted or duplicated.

Things I tried while troubleshooting:

* Different Windows fonts - same problem.
* Retyped the text from scratch - same problem.
* Set Movement to None - same problem, so scrolling itself isn't causing it.
* Changed XL Font to xLights built-in `7-7x9 Bold` - problem completely disappears.
* Made a copy of the matrix for testing and changed only Strands/String from 7 to 14 - Windows OS font then rendered cleanly.
* Changed the test back to 7 strands - problem returned.

So 7 rows seems to be what exposes the problem with OS fonts.

**Expected behavior**

I expect Windows fonts to render normally on the 35x7 matrix like they did before. It is just a small Tune-To sign, and the same sequence/text worked previously.

**Screenshots**

[...]

PR #6998, opened 2026-08-31
Fix Windows OS-font Text effects rendering corrupted

derwin12:
OS Fonts on Windows was producing a corrupted render.. also the color per word was not triggering a re-render.


derwin12:
#6995

PR #7011, opened 2026-09-02
#7002 - Fix/timer pending livelock

### LINKED ISSUE #7002: try to create 40fps animation causes xligths to lock up and stop responding.
I have 2 windows 11 computers running 2026.16 and when trying to create a 40fps anmiation sequence, then dropping an effect on the timeline, xlights just hangs up and goes not responding and never responds again.  I have to terminate and end process to recover.  

Process to repeat, new sequence, animation, 40fps, quick start.  It populates the display list and everything at this point seems fine.  Drop any effect on any model in the model list and it just goes not responding... If I choose 20fps it works without any issue at all.  only when choosing 40fps.   

One of the 2 computers I have 2026.15 installed on also and this behavior does NOT occur in 2026.15 or older only in 2026.16. 
Windows 11 12gb RAM, 4 core ryzen 5.  Other computer is ryzen 7.  Daryl E had me try different preferences for preview graphics setting and same issue with auto, opengl, or vulkan.  Again this worked as expected in 2026.15 without issue. 

[xLightsProblem.zip](https://github.com/user-attachments/files/31665971/xLightsProblem.zip)

AGFazio:
---Fixes #7002 — 40fps animation freezes xLights on Windows.

xLightsTimer::DoSendTimer() cleared _pending before running wxTimer::Notify(). _pending is the only cap on outstanding CallAfters, so while a frame was being delivered the flag read false and the next thread-pool tick queued another one. wxAppConsoleBase::ProcessPendingEvents loops while (!m_handlersWithPendingEvents.IsEmpty()) — once a frame took longer than the timer interval the handler re-added itself faster than the loop drained it, so the loop never exited: no repaint, no input, "Not Responding" until the process was killed. Only a marginal overrun is needed (26ms against a 25ms interval is enough), which is why 40fps hangs and 20fps doesn't.

Clear _pending only after delivery, via a scope guard so an exception escaping Notify() can't leave the flag stuck true and silently kill playback (SafelyProcessEvent catches outside our frame).

Windows-only. macOS drives the timer from a main-thread CADisplayLink and Linux from a GLib timeout — neither hands the tick across a thread boundary, so neither can accumulate a backlog.

Testing: New sequence → Animation → 40fps → Quick Start → drop an effect on a model. Also worth a pass on the Pixel Test dialog, the only other xLightsTimer user.

Two things I'd flag for the reviewer, since I couldn't settle them here:

- Smoothness at 40fps is a separate question from the hang. A tick is now dropped only when the frame itself overran the interval. If frames fit in 25ms you get a true 40fps — better than 2026.15, which per Dan's measurements delivered a median 31.3ms tick and couldn't reach 40fps at all. If frames don't fit, the fix stops the hang but the effective rate drops, and no timer policy can fix that. We never measured actual frame cost on your box, so that's genuinely open.
- The lever if it turns out to be marginal is NeedToRenderFrame in tabSequencer.cpp:2749, which unconditionally returns true on Windows/Linux while macOS throttles per-canvas against screen present time.

### DIFF
--- a/src-ui-wx/shared/utils/xLightsTimer.cpp
+++ b/src-ui-wx/shared/utils/xLightsTimer.cpp
@@ -238,11 +238,17 @@ void xLightsTimer::Notify() {
 }
 
 void xLightsTimer::DoSendTimer() {
-    // Cleared before delivering rather than after: a frame that overruns the
-    // interval by any amount would otherwise discard the tick that lands while it
-    // is still running and give up a whole interval, which is enough to drop the
-    // output frame.
-    _pending = false;
+    // Cleared only once the frame has been delivered. Clearing first lets a tick
+    // that lands mid-frame queue another CallAfter, and
+    // wxEvtHandler::ProcessPendingEvents loops until its queue is empty - with a
+    // frame slower than the interval it never is, so the main loop stops
+    // servicing paint and input and the app hangs.
+    struct ClearOnExit {
+        std::atomic<bool>& flag;
+        ~ClearOnExit() {
+            flag = false;
+        }
+    } clearOnExit{ _pending };
     wxTimer::Notify();
 }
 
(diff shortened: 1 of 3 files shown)


computergeek1507:
Please only commit the timer cpp change, not the project file changes

AGFazio:
Yeah. Oops. I was having VS issues. I will resubmit tomorrow.

derwin12:
Confirmed working on Ron's vm.

Issue #7006, opened 2026-09-01
Import Empty Groups Imports Names of Source File Models


MCStoff:
**Describe the bug**
When importing empty groups from an external rgbeffects file, the program brings in the names of the props in that group in the source file into the target layout even though the props are not being imported.  Bringing in the names of the source file models prevents the groups from rendering properly until the source file model names are deleted.

**To Reproduce**
1.  Select Layout tab.
2.  Right click on house visualization.
3.  Import Previews/Models/Groups, etc.
4.  Select groups to import
5.  Check box to import empty groups
6.  Select group name to see the source file model names

**Expected behavior**
Previous behavior was to import the empty groups without the names of the source file models.  The groups could then be propagated with the target layout models without the necessity of deleting the source file model names.  If the change was intentional, it should be optional, which could be implemented by another check box at the bottom of the group/model identification dialog/window.

**Screenshots**
NA

**Versions (please complete the following information):**
 - OS:  Apple?
 - xLights version: 2026.16

**Additional context**
Note this issue was encountered while assisting a customer with mapping.  I do not have files illustrating the problem as the issue was encountered while in a Zoom session.  I would consider this a minor bug in that it changes the import behavior and creates additional work to delete the source file models, otherwise rendering doesn't work properly.  In those instances where the intent is to bring in the groups and the models, it is a nice feature, but there should be a way to "opt out" or revert to the previous behavior.

**Attachments**
NA

derwin12:
FYI - been broken since atleast 2026.04 - so perhaps always ;-)

MCStoff:
It could be made into a feature. Lol
MichaelShowstopper Sequences
 

    On Tuesday, September 1, 2026 at 11:42:29 AM CDT, Daryl ***@***.***> wrote:  
 
 derwin12 left a comment (xLightsSequencer/xLights#7006)
FYI - been broken since atleast 2026.04 - so perhaps always ;-)

—
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications, keep track of coding agent tasks and review pull requests on the go with GitHub Mobile for iOS and Android. Download it today!

AGFazio:
PR 7012

derwin12:
Andy has this in the next update

PR #7012, opened 2026-09-02
#7006 Fix for Importing Empty Groups

AGFazio:
Fix Import Previews/Models/Groups not importing selected groups (#7006)

Importing model groups from another show's xlights_rgbeffects.xml had two defects:

- A selected group wasn't imported when its member models weren't also selected. The "Include empty model groups" gate tested whether any member resolved in the destination at that instant, so a group you ticked without its models was skipped entirely. It now tests whether the group is empty in the source layout, so a selected group always imports — empty if none of its members came across. (The resulting empty group shows in Check Sequence, which is expected.)
- An imported group carried the source show's full member list. The membership list was filtered for the models added after creation, but the group was created straight from the source XML node, whose models attribute still named every member of the source group. The group arrived referencing models that were never imported and wouldn't render until they were deleted by hand. The filtered list is now written onto the node before the group is created.

Also fixed while in here: the filter dropped members that simply didn't exist yet — a member group ordered after its parent, or a member in a preview handled by a later pass — which permanently lost them from the group. Members belonging to the same import are now retained, and all of them exist by the time the import finishes.

Same three fixes applied to the iPad group pass in XLSequenceDocument.mm, which shared the identical logic. Parity row 218a updated in plans/platform-parity/06-layout-models.md.

The import dialog is unchanged — no .wxs edit, and the checkbox keeps its label and unticked default.


derwin12:
Andy .. this commit has your new stuff plus your old/bad stuff. You didnt have a clean master when you started this branch.
Perhaps create a new branch and have it pull in the good stuff and do a new PR.

AGFazio:
Rebased against the clean master.

computergeek1507:
git config --global pull.rebase true

derwin12:
Seems to work from my testing..

PR #7068, opened 2026-09-14
Import From Model pickers: add a filter to the model list

heffneil:
## What
The **Import From Model** pickers in the model editor (Faces, States, SubModels) used a plain `wxSingleChoiceDialog` with no filter — a long model list could only be scrolled. This adds a **"Filter models…"** search box to that picker, consistent with the filters elsewhere in the app.

Covers all four "Select Model" sites:
- **States** → Import → Import From Model
- **Faces** → Import → Import From Model
- **SubModels** → Import From Model (from a show model, and from an imported `.xmodel` file)

## How
Adds a small reusable free function `ChooseModelWithFilter(parent, choices, title)` to the existing `shared/dialogs/CheckboxSelectDialog.{h,cpp}` — a single-select model list with a search box above it, whitespace-tokenized AND matching (`midwest pumpkin` → `Midwest-Coro-Pumpkin`), double-click to accept, returns the chosen name or empty on cancel. All four sites drop `wxSingleChoiceDialog` for this helper.

Put in the existing shared dialog file (not a new file) on purpose: `.cbp`/`.vcxproj` list sources explicitly, so a new file risks a missed Windows/Linux entry; CMake globs `shared/dialogs/*.cpp` but the other two don't. Reusing the compiled-everywhere file avoids that.

Also adds a missing `m == nullptr` guard at the SubModels import site, matching the Faces/States siblings.

## Testing
Built (macOS Debug). Verified the filter box appears and narrows the list (single and multi-term) in States/Faces/SubModels Import From Model, and that a pick still imports correctly.

## iPad parity
Desktop-only model-editor panels (`src-ui-wx/model/`); the iPad model editor is a separate surface. No iPad counterpart to change here.


heffneil:
Screenshot 2026-09-14 at 7 12 48 AM
Screenshot 2026-09-14 at 7 13 04 AM
Screenshot 2026-09-14 at 7 12 53 AM

PR #7069, opened 2026-09-14
Record the sequence(s) an import came from, with Open Original File re-import

heffneil:
## What
Imports now **remember which sequence(s) the effects came from** (the "donor"), so you can see the source later and re-run the same import without hunting for the file.

Covers every effect-import path: **Tools → Import Effects** and the LMS/LAS, LPE, S5 `.loredit`, Vixen 2 `.vix`, Vixen 3 `.tim`, and xLights `.xsq/.xsqz` importers.

Three touch points:
- **Map Channels dialog** — a new opt-in **"Record donor sequence in this sequence's metadata"** checkbox (default on) with the donor filename shown beside it (full path on the tooltip).
- **Import → Open Original File** — a submenu listing the recorded donors; picking one re-runs that import. Disabled until the current sequence has recorded a donor; if a donor file has since moved/been deleted it says so rather than failing silently.
- **Sequence Settings → Meta Data** — a read-only **"Imported From:"** field listing the donors (most recent first).

## How
Persistence lives in the wx-free core: `SequenceFile` gains an `IMPORTED_FROM` header and a small MRU list (`RecordImportedFrom` / `GetImportedFrom`), serialized as repeating `<importedFrom>` header elements (deduped, most-recent-first, capped at 10). Because it's in core, the record round-trips on both desktop and iPad.

The desktop UI records the donor only when the Map Channels dialog was actually shown and the opt-in is ticked (`xLightsFrame::RecordImportDonor`), rebuilds the submenu from the list (`UpdateImportFromOriginalMenu`, per-item `NewControlId()` with matching unbind/unreserve cleanup), and re-imports on click (`OnMenuItemImportFromOriginal` → shared `ImportEffectsFromFile`).

Package imports record the **archive the user opened** — new `SequencePackage::GetPkgFile()` — instead of the temp-extracted `.xsq`, so "Open Original File" can find the donor again.

wxSmith: the new controls in `SeqSettingsDialog` and `xLightsImportChannelMapDialog` are mirrored in their `.wxs` files; the Import dialog's dynamic donor label/tooltip is set outside the generated guard.

## Testing
Built macOS desktop (Debug) and the iPad library (Debug) — both succeed. Verified the checkbox + donor name appear in Map Channels, the donor is recorded on OK, **Open Original File** lists and re-runs it, the Meta Data field shows the list, and the record survives save → reopen.

## iPad parity
Desktop-only UI. The `<importedFrom>` metadata is read/written by the shared core `SequenceFile`, so it survives an iPad open/save, but the iPad has no UI to opt in, view, or re-open donors. Captured as **State B**, row 90 in `plans/platform-parity/08-import-export.md`.


heffneil:
Screenshot 2026-09-14 at 10 27 55 AM
Screenshot 2026-09-14 at 10 28 23 AM

Screenshot 2026-09-14 at 10 28 53 AM

derwin12:
Need to rebase and move that new checkbox up to the top along with the other options.
Why are there 2 very different songs in the import effects dropdown - was that just testing?

heffneil:
Yes we were testing with two different sequences.  Like you I have a template sequence I use for "standard stuff" and then the actual song based sequence.  Its just to demonstrate if you have more than one sequence you import from they are logged if you choose.

heffneil:
Screenshot 2026-09-19 at 11 04 06 PM

PR #7082, opened 2026-09-16
Update for ILT controllers

PeteMatthews:
Add new Scene controller
Allow serial outputs to stand in for relays on Scene


PeteMatthews:
Ugh - raised in error. Back soon :)

PR #7093, opened 2026-09-18
Update for ILT controllers

PeteMatthews:
Add new Scene controller
Allow serial outputs to stand in for relays on Scene


PeteMatthews:
As discussed briefly with @dkulp - thanks for your assistance.

derwin12 (commented):
Did an AI review and it doesnt like line 168 as it might be empty.

  1. Line 168 — silent fallback failure: The new loop copies port_type from jsonVal["ports"], but jsonVal is only populated if the earlier /settings GET succeeds
     (lines 63-80). If that fetch fails or returns unparseable JSON, jsonVal stays null, so the loop silently iterates zero times — leaving port_type unset for
     any gap slots (e.g., unused relay ports) instead of falling back to a sane default like the old code did. No error is surfaced.
  2. Line 168 — stale/mismatched port count: The loop bounds itself on jsonVal["ports"].size(), i.e. how many ports the device currently reports, not how many
     the controller capability rules (GetControllerCaps()) say should exist. If the device reports fewer ports than the new config actually builds (e.g.,
     first-time setup, firmware default, or switching to a variant with more ports), trailing ports never get port_type set even though num_pixels/models were
     already written for them. Also, jsonVal["ports"][i]["port_type"] is read without a .contains() check, so a missing key writes JSON null instead of a safe
     default.

PR #7095, opened 2026-09-18
Add Match Depth to the layout Resize menu in 3D mode

heffneil:
## What

The layout right-click **Resize** submenu offers Match Width, Match Height and Match Size, but there is no way to match depth on its own — even though the neighbouring **Align** and **Distribute** submenus both already expose a Depth entry in 3D.

`Match Depth` is now appended to the Resize submenu, gated on `is_3d` exactly the way `AddAlignOptionsToMenu` gates "Depth Center" and `AddDistributeOptionsToMenu` gates "Depth", so it does not appear in 2D.

The iPad layout editor has had a **Same Depth** action for a while (`LayoutEditorView.swift` -> `matchSizeOfModels:dimension:`), so this is the desktop catching up to the iPad rather than the other way around.

## Behaviour preserved

`PreviewModelResize` / `PreviewObjectResize` take a `sameDepth` flag, and the existing menu items keep doing exactly what they did before:

- Width has always carried depth with it on models that z-scale (`SetMDepth(depth)` inside the `sameWidth` branch).
- A partial resize of a boxed model has always copied the source's `matrixScale.z`.

So the only new effect is being able to match depth *without* also matching width. This leaves a deliberate difference from the iPad, which treats all three dimensions independently — changing the desktop to match would alter what Match Width does today, so it is not done here.

## Drive-by fix

`ViewObjectPanel::PreviewObjectResize` set a view object's depth to the **width** value on Match Width:

```cpp
view_object->GetBaseObjectScreenLocation().SetMDepth(width);   // before
```

It now uses the source object's depth, read via `GetRestorableMDepth()` — the true inverse of `SetMDepth()`. `GetMDepth()` is not, and using it as a get/set pair makes the value drift on repeated matches.

## Parity

Desktop-only UI change in `src-ui-wx/layout/` — **State B**, with row 41 of `plans/platform-parity/06-layout-models.md` updated. The iPad side already has the feature, so no iPad work is needed; the row now also records the width/depth coupling difference between the two.

## Testing

- macOS desktop Debug build succeeds.
- Match Depth appears in the Resize submenu in 3D and is absent in 2D, from both the preview right-click and the model-tree right-click.
- Match Width / Match Height / Match Size unchanged.


heffneil:
Screenshot 2026-09-18 at 11 40 19 AM

heffneil:
2d no depth:

Screenshot 2026-09-18 at 1 31 55 PM

PR #7104, opened 2026-09-19
Open the controller visualiser in its own window

heffneil:
## What

The controller visualiser ran `ShowModal()`, so the entire main window was blocked for as long as it was open — you could not look at the Layout tab, check a model's properties, or compare against another controller without closing it first. It now opens as its own window.

## Why this was more than swapping ShowModal for Show

The dialog itself is well suited to being modeless: it has no OK/Cancel and no `EndModal`, and applies every change as you make it. But three things had to change before it could safely outlive the call that opened it.

**It borrowed a `UDController` the caller built on the stack.** Both call sites did:

```cpp
UDController cud(controller, _frame->GetOutputManager(), &_frame->AllModels, true);
ControllerModelDialog dlg(_frame, &cud, &_frame->AllModels, controller);
dlg.ShowModal();
```

The dialog now owns it — the constructor takes a `std::unique_ptr<UDController>`, and `ControllerModelDialog::ShowFor(frame, controller)` builds and hands it over.

**A modeless `wxDialog` hides rather than destroys.** Left alone it would keep the stale `UDController` alive behind an invisible window, so the dialog destroys itself on `wxEVT_CLOSE_WINDOW`.

**With the main window usable, the show can change underneath it.** That could not happen while it was modal:

| Event | Handling |
|---|---|
| Models added/deleted/renumbered | `RefreshAll()` from `xLightsFrame::DoWork` rebuilds every open visualiser on the model/controller work bits |
| Its controller deleted | `CloseFor()` before the delete in `DeleteSelectedControllers`; `ControllerStillExists()` also closes it on the next refresh |
| Show folder changed | `CloseAll()` once `SetDir` is past its point of no return |
| Visualise clicked again for the same controller | `ShowFor` raises the open window instead of opening a second |

`RefreshAll` carries a re-entrancy guard: `ReloadModels` → `FixDMXChannels` → `DoAllWork` can land back in `DoWork`, and the outer call rebuilds anyway.

`s_activeCount` becomes `s_openDialogs`, so the existing `IsAnyActive()` deferral of model-tree selection in `DoWork` still works. Its comment said the visualiser "runs modal over the Layout tab" — updated, though the deferral itself is still wanted so the dialog and the tree do not fight over the selection.

## Drive-by fix

`UDController::Rescan` clears the ports and rebuilds from the current models, but `_noConnectionModels` was only ever appended to. That did not matter when the `UDController` died with a modal dialog; a long-lived one would hold pointers to models deleted since the last scan. It is now cleared on rescan.

## Note for reviewers

It is still a `wxDialog` parented to the frame, so on macOS it floats above the main window rather than interleaving with it. The main window is fully usable, which is the point — but if the preference is for a window that can go behind xLights, that means converting wxSmith-generated code to a `wxFrame` and is better as its own change.

## Parity

`src-core/` + `src-ui-wx/` — the core change is shared and needs no iPad work (the iPad visualiser is a SwiftUI view that is already non-blocking). **State B**, row 29 of `plans/platform-parity/07-controllers-setup-upload.md` updated.

## Testing

- macOS desktop Debug and `xLights-iPadLib` both build.
- Visualiser opens non-blocking; the main window stays usable with it up.
- Re-clicking Visualise for the same controller raises the existing window.
- Deleting a model on the Layout tab with the visualiser open rebuilds it.
- Deleting the visualised controller closes it; changing show folder closes it.


dkulp:
Just want to say that I'm against this for the same reasons the Faces/Submodel/etc... dialogs should not be non-modal.

heffneil:
the problem this hopes to solve it to identify props as you are visualizing them. Since the visualizer likely covers the layout its a challenge even with my other PR to truly identify them when selecting them. If there is a better solution, I am all for it.

PR #7107, opened 2026-09-20
Add a Default Max Brightness preference for newly added controllers

heffneil:
## What
Adds a **Default Max Brightness** application preference so newly-added controllers start at a chosen brightness instead of always 100%.

- **Preferences → Output** gains a **"Default Max Brightness"** field (1–100). When set, every controller you add from then on is stamped with that value for its *Default Port Brightness*.
- **Blank = unset** — new controllers keep the existing default (100), so behavior is unchanged until you set a value.
- A companion **"Apply to all existing controllers"** button pushes the current value onto every existing controller that supports it (with a confirmation), for retrofitting a show.

## How
- Preference stored on `xLightsFrame` (`_defaultMaxBrightness`, getter/setter), persisted via config key `xLightsDefaultMaxBrightness` (read at startup, written on shutdown) — mirrors the existing `xLightsControllerPingInterval` pattern.
- UI is a digits-only `wxTextCtrl` (empty == unset) clamped live to **1–100**, plus the bulk-apply button, added to `OutputSettingsPanel`; the new controls are mirrored in `OutputSettingsPanel.wxs`.
- The value is stamped **only at the interactive add sites** — the Controllers tab right-click (ADDSERIAL/ADDETHERNET/ADDNULL), the legacy Setup "Add …" buttons, and the `addEthernetController` automation command — so discovery / import / clone paths that mirror real hardware are deliberately left untouched.
- Reuses the existing per-controller `Controller::SetDefaultBrightnessUnderFullControl`, gated on `SupportsDefaultBrightness()`, so no new serialization or core behavior. Bulk apply is `xLightsFrame::ApplyDefaultMaxBrightnessToAllControllers` (marks the network dirty + refreshes the list).

## Testing
Built macOS desktop (Debug). Verified: the field appears on Preferences → Output; typing a value stamps newly-added Ethernet/Serial/Null controllers with that Default Port Brightness while existing ones are untouched; entering >100 snaps to 100 and non-digits are rejected; blank leaves new controllers at 100; the Apply button (enabled only when a value is set) updates all existing controllers after confirmation; and the preference persists across a restart.

## iPad parity
Desktop-only preference (no Preferences screen or controller-add UI on iPad). The underlying `DefaultBrightnessUnderFullControl` is core and already round-trips on both platforms. Captured as **State B**, row 48 in `plans/platform-parity/11-preferences-settings.md`.


heffneil:
Screenshot 2026-09-19 at 10 45 36 PM
Screenshot 2026-09-19 at 10 45 12 PM

PR #7108, opened 2026-09-20
Highlight the model picked in the visualiser in the layout preview

heffneil:
## What

Clicking a model tile in the controller visualiser now colours that prop in the layout preview, so you can see which physical prop a tile refers to without hunting for it by name.

## What it replaced

A mouse-down in the models pane already did something, but only as a warning and only for a minority of tiles:

```cpp
const bool unassigned = !clickedModel->IsShadowModel() &&
    (controllerName.empty() || controllerName == NO_CONTROLLER);
for (auto pm : preview->GetModels()) {
    pm->NotOnController = unassigned && (pm == clickedModel);
}
```

So an *unassigned* model flashed orange while the button was held, and an assigned one gave no feedback at all. Picking any tile now highlights it, in a colour of its own.

## Behaviour

| Action | Result |
|---|---|
| Click a tile in either pane | That prop draws in the highlight colour |
| Click the picked tile again | Clears |
| Click a different tile | Highlight moves |
| Drag a model | Stays lit through the drag, clears once it lands |
| Click empty space in a pane | Clears |
| Clear the model filter | Clears |
| Click in the layout preview, or close the visualiser | Clears |

Colour is a new `COLOR_MODEL_IN_VISUALISER`, default `(0,128,255)`, shown in Preferences → Colors → Layout Tab as "Model Picked In Visualiser".

## Three things worth reviewing

**Assigned models needed the other pane.** They are not tiles in the models pane at all — they sit on the ports panel, handled by `OnPanelControllerLeftDown`, which had no highlight code. Both panes now call one `SetVisualiserHighlight`.

**The port highlight was swallowing the colour.** A model on the selected controller carries the per-node yellow/dim port highlight, which ends with:

```cpp
color = nullptr; // DisplayModelOnWindow reads per-node colors when c == nullptr
```

That discarded the base colour outright, which is exactly why unassigned models appeared to work and assigned ones did not. The picked model now skips that override; its siblings keep the yellow/dim treatment.

**Page gating.** The orange not-on-controller highlight is gated to the Controllers page. This one is not — the visualiser can be open over either page — but it is still limited to the layout editor's own preview, never the house or sequencer previews.

## Lifetime and input details

- A drop calls `ReloadModels()`, which deletes every `BaseCMObject` in the pane. The `Model*` is therefore captured **before** `DoDragDrop` rather than read back off the tile afterwards; reading it after was a use-after-free.
- Whether a press was a click or a drag is decided by whether the pointer actually moved, not by `DoDragDrop`'s result — a release over the same pane can come back as `wxDragMove` even when nothing moved.
- The current pick is read back from the model flags rather than cached in the dialog, because the layout preview clears those flags in its own click handling and a cached copy would go stale.

## Parity

`src-core/` (one transient flag) + `src-ui-wx/`. **State B** — new row 30a in `plans/platform-parity/07-controllers-setup-upload.md`. The iPad visualizer is a full-screen SwiftUI view with no layout preview beside it, so there is nothing to highlight into.

## Testing

- macOS desktop Debug and `xLights-iPadLib` both build.
- Verified for assigned and unassigned models, in both panes, for click / re-click / different tile / drag to another port / drag out to the unassigned column / click empty space / clear filter.

Best seen together with #7104, which stops the visualiser blocking the main window — though this stands on its own.


heffneil:
Screenshot 2026-09-19 at 10 55 14 PM
Screenshot 2026-09-19 at 10 55 01 PM

derwin12:
Going to make a small update to this as it removes  the highlight if you click on string 2/3 etc of a multi string prop

PR #7111, opened 2026-09-20
Share one filter matcher across the filter and search boxes

heffneil:
## The problem

Typing `all house` into a filter box finds nothing when the group is called `grp_all_house_display`. Almost every filter in the app does a single substring test:

```cpp
wxString(model->GetName()).Lower().Contains(filterString.Lower())
```

so a query containing a space has to occur literally, spaces and all. On top of that, each screen had its own copy of the matching logic — **thirteen filters across eleven files**, several subtly different from each other, and one (`ImportPreviewsModelsDialog::MatchesFilter`) already smarter than the rest but private to that dialog.

## What they share now

`xl::FilterQuery` in `src-core/utils/FilterMatch.h` — header-only and wx-free, so there are no project-file edits and the iPad can adopt the same rules later — behind a `wxFilterQuery` wrapper in `wxUtilities.h` that adds the wx-only part (regex, Unicode case folding).

| Query | Behaviour |
|---|---|
| `all house` | every word must appear, in any order |
| `allhouse` | separators optional — matches `all_house` |
| `grp_*_house`, `h?use` | `*` and `?` wildcards, unanchored |
| `/^grp_.*house/` | a real regex |
| `ouse`, `grp_all` | plain substrings, exactly as before |

**Every rule is additive**: the matcher tries the plain substring first, so nothing that matched before stops matching, and it only builds segments when a token actually needs them.

## Converted

Layout model / group / preview filters, the sequencer row filter, the controller list, the visualiser (model pool *and* the dim-out), import mapping, the checkbox pickers, import previews/models, pixel test (flat list and tree prune), and the three download browsers.

Each filter box also gained a tooltip describing the syntax — none of them had one, so the behaviour was undiscoverable either way.

## Two behaviour notes for reviewers

**Bare regex becomes globbing in three places.** The layout model/group trees and the controller list compiled bare input with `wxRE_ICASE`, so `*` there meant regex-star and `grp_*_house` did *not* match `grp_all_house_display`. `*` is now a wildcard; `/pattern/` gets full regex back. This is the only place an existing query can behave differently, and it was never documented in the UI.

**A half-typed regex falls back to plain matching.** `/^grp_[` is uncompilable, and blanking the list on every keystroke while someone types a pattern reads as broken.

## Performance and correctness details

- Queries are built **once per keystroke** and reused for every row. Several sites (including the previous `MatchesFilter`) re-tokenized the query for every row; `PruneTree` now takes the parsed query rather than a string, and the panels that filter from a member cache it next to the filter text.
- Case folding stays on `wxString::Lower()` so non-ASCII names fold as they did before; conversion to `std::string` uses `utf8_string()` rather than `ToStdString()` to avoid locale-dependent narrowing on Windows.
- `wxRegEx` is not copyable, so the wrapper holds it by `shared_ptr` and stays a plain value type.

## Testing

- macOS desktop Debug, `xLights-iPadLib`, and a **NO_PCH** desktop build all succeed (the last matters for a new header).
- The matcher was exercised standalone with 20 assertions before wiring: the reported case and its permutations, wildcard cases, negative cases, names containing spaces, case-insensitivity, and a set asserting that plain substrings still match.
- Driven in the app across the layout, sequencer, controller and visualiser filters.


dkulp:
Is there a particular reason for using wxRegEx instead of std::regex?   If we're updating any of this code, I would think moving to std::regex is the right thing to do (and more core likely live in the stuff in src-core).

heffneil:
Fair question, and having gone back through it I think you're right that this should end up in core with `std::regex`.

The reasoning at the time was mostly about it being *user-typed* input:

- `std::regex` throws `std::regex_error` on an invalid pattern, and a filter box is invalid on most keystrokes while someone types `/^grp_[`. `wxRegEx::Compile()` just returns false, which made the "fall back to plain matching until it compiles" behaviour a one-liner. AGENTS.md's "nearly non-existent exception handling" (and the `std::stoi` ban for the same reason) pushed me the same way.
- `<regex>` is a heavy header to pull into a widely-included utility, and libstdc++'s implementation is recursive enough to blow the stack on pathological patterns/long subjects — a segfault rather than a clean no-match.
- `wxRegEx::Matches()` takes a `wxString`, so it follows the platform's wide/narrow representation. Over UTF-8 bytes `.` matches a byte and character classes misbehave on non-ASCII, which is the same class of bug as the `ToStdString()` → `utf8_string()` fix already in this PR.

But the src-core argument outweighs that, and the exception objection turns out to be already solved in-tree: `AutoMapper.cpp:112-123` builds a `std::regex` from a user-supplied pattern and catches `std::regex_error` to mark it invalid. That's precisely the shape this needs, and there are ~25 files in `src-core/` using `std::regex` already, so wxRegEx is the odd one out rather than the house style.

What I'd change, if you're happy with it:

- Move the `/pattern/` handling into `xl::FilterQuery` in `FilterMatch.h`, using `std::regex` with a try/catch around construction exactly like AutoMapper, so an uncompilable pattern degrades to plain matching instead of throwing.
- That makes the whole matcher wx-free, so the iPad can use the same rules instead of reimplementing them — which was the point of putting the tokens/wildcards in core to begin with.
- `wxFilterQuery` then shrinks to what actually needs wx: `wxString::Lower()` for non-ASCII case folding and the UTF-8 conversion. Call sites don't change.

One thing worth deciding explicitly rather than by accident: `std::regex` over UTF-8 is byte-oriented, so `.` and `[a-z]` operate on bytes and `icase` only folds ASCII. For the token/wildcard path that doesn't matter (case folding happens on `wxString` before the match, and the glob matcher is byte-transparent), but a `/regex/` against a prop name with accented characters would behave differently than it does with wxRegEx today. Given `/pattern/` is the power-user escape hatch and everything else is words and wildcards, I think that's an acceptable trade — but say the word if you'd rather keep regex on the wx side for that reason and have core carry only the tokens and wildcards.

Happy to push the change either way.

heffneil:
Done in f3e5366 — `/pattern/` now compiles with `std::regex` inside `xl::FilterQuery`, using the same `catch (const std::regex_error&)` shape as `AutoMapper.cpp`, so an uncompilable pattern still degrades to plain matching instead of throwing or blanking the list while it is being typed.

The matcher is wx-free as a result, so the iPad can use it directly rather than reimplementing the rules. `wxFilterQuery` is down to the two things that actually need wx — `wxString::Lower()` for non-ASCII case folding, and UTF-8 conversion — and since `std::regex` is copyable the `shared_ptr` went away with it. No call site changed.

One deliberate detail: a regex and its subject are both passed through unfolded, because lowering a pattern would break character classes like `[A-Z]`; `icase` covers the usual intent. The words-and-wildcards path still folds on `wxString` so accented names behave as they did before.

Desktop Debug, `xLights-iPadLib` and a NO_PCH build all pass, and the standalone matcher checks now cover the regex path as well: anchors, `icase`, a half-typed `/^grp_[` falling back to literal, and `//` (empty pattern) staying literal rather than matching everything.

